Issue #5, opened 2018-01-04
Build out unit tests


bwillard:
Right now we have 1 unit test, that is pretty terrible, we need to expand our test coverge

holachuy:
Pushing back due date

Swikar:
Swinging! Anyone here help me out with more detail? I know its long time since last post

Swikar:
/ assign swikar

Gairick52:
@Swikar hello can you please assign it to me, i want to work on this issue

Simpenzwe-Honore-Leandre:
@Gairick52 are you still working on this ?

ofekkr1:
hello,
is the issue still relevant? if it is, may I be assigned to work on it?

igmtz:
Hello! Is someone currently working on this issue? I would like to contribute!

aditya-32:
Hi, Is this issue still relevant,
this is my first contribution, I would like to contribute to this can I be assigned the issue?

bengoh815:
Hey, I noticed this issue from 2018 and there has not been much activity or responses. Is this still something that needs to be addressed, or has it been resolved elsewhere?

lisad:
There is already way more than 1 unit test but nobody remembered to close this until now!  closing now.

PR #11, opened 2018-01-05
Add a test for CrypterImplTest

bwillard:
Adds one test as per #5

### DIFF
--- a/portability-core/src/test/java/org/dataportabilityproject/job/CrypterImplTest.java
+++ b/portability-core/src/test/java/org/dataportabilityproject/job/CrypterImplTest.java
@@ -1,0 +1,18 @@
+package org.dataportabilityproject.job;
+
+import static com.google.common.truth.Truth.assertThat;
+
+import javax.crypto.KeyGenerator;
+import javax.crypto.SecretKey;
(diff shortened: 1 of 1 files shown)


portability-core/src/test/java/org/dataportabilityproject/job/CrypterImplTest.java
```diff
@@ -0,0 +1,18 @@
...
+import javax.crypto.SecretKey;
+import org.junit.Test;
+
+public class CrypterImplTest {
+  @Test
+  public void testRoundTripSecretKey() throws Exception {
+    KeyGenerator generator = KeyGenerator.getInstance("AES");
+    SecretKey key = generator.generateKey();
+    CrypterImpl impl = new CrypterImpl(key);
+    String data = "The lazy dog didn't jump over anything";
```
holachuy: Maybe add a few special characters.

holachuy: Maybe add special chars

bwillard: Done.

Issue #8, opened 2018-01-04
XSRF protection for demo app


bwillard:
Add XSRF protection to our Demo App and API end points.

seehamrun:
I deployed this to our gcp instance and it looks working as intended - closing out the issue.

PR #16, opened 2018-01-05
Switch to HttpClient from angular/common/http

seehamrun:
This provides us with support for client-side xsrf prevention for #8 

See: https://angular.io/guide/http#security-xsrf-protection

### DIFF
--- a/client/src/app/backend.service.ts
+++ b/client/src/app/backend.service.ts
@@ -15,7 +15,7 @@
  */
 import { environment } from '../environments/environment';
 import { Injectable } from '@angular/core';
-import { Http, Headers, Response, RequestOptions, URLSearchParams } from '@angular/http';
+import { HttpClient, HttpResponse, HttpParams, HttpErrorResponse } from '@angular/common/http'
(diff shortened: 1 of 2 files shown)


client/src/app/backend.service.ts
```diff
@@ -60,55 +58,48 @@ export class BackendService {
...
     for (var name in importData) {
       importServices.push(new ServiceDescription(importData[name], importData[name]));
     }
-
     return new ServiceDescriptions(importServices, exportServices);
   }
 
-  private importSetupSuccess(res: Response) {
-    let data = res.json();
+  private importSetupSuccess(res: any) {
```
holachuy: Do we know the object type here or below?

seehamrun: It just returns an Object, we can add type checking to the response like so: https://angular.io/guide/http#typechecking-the-response 

But I didn't think to add it to the listDataTypes call since its just a json array, without a tag. 
Added to listServices, importSetup and copySetup

PR #17, opened 2018-01-05
Switch logging framework to slf4j

### LINKED ISSUE #2: Modularize Logging
Right now we have LogUtils.

We should change that to an interface and put implementations in the cloud providers.

bwillard:
Fixes #2


config/environments/local/log4j.properties
```diff
@@ -0,0 +1,6 @@
+log4j.rootLogger=DEBUG, A1
+log4j.appender.A1=org.apache.log4j.ConsoleAppender
```
rtannenbaum: If the only difference between these files is the log level, might be cleaner to have one log4j.properties file and specify log level in settings/common.yaml.

You can do a string replace before copying the properties file into the binary, for an example we do this a few places in config/gcp/setup_gke_environment.sh with sed

portability-core/pom.xml
```diff
@@ -278,6 +278,19 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <!--Provides compile time bindings for slf4j -->
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>1.8.0-beta0</version>
```
rtannenbaum: FWIW, Google has version 1.7.14 imported (for both libraries)

bin/build_and_run_jar.sh
```diff
@@ -33,7 +33,8 @@ SRC_DIR="portability-$BINARY"
 mvn=$(which mvn)|| { echo "Maven (mvn) not found. Please install it and try again." >&2; exit 1; }
 
 # Copy settings yaml files from ENV/settings/ into $SRC_DIR/src/main/resources/
-SETTINGS_DEST_PATH="$SRC_DIR/src/main/resources/settings/"
+DEST_RESOURCES_PATH="$SRC_DIR/src/main/resources"
```
rtannenbaum: can you name these vars consistently re where _DEST_ occurs in the variable name?

config/environments/local/log4j.properties
bwillard: For now the only diff is the level, but over time I think they will differ more so I think we should just leave them separate.

portability-core/pom.xml
bwillard: Is there any problem with just leaving them on latest here?

bwillard: As per off line discussion, updated to 1.7.14

bin/build_and_run_jar.sh
bwillard: Done

bwillard:
PTAL?

config/environments/local/log4j.properties
rtannenbaum: SGTM

Issue #3, opened 2018-01-04
Finish Encryption flow


bwillard:
Finish the job flow using encryption with separate workers

PR #33, opened 2018-01-09
Fix bugs when encryption flow is enabled in the API

holachuy:
This is a collection of bug fixes to complete the implementation of the encryption flow (using encrypted auth credentials in cookies and storing metadata state in storage) for the API server. #3


config/environments/local/settings/common.yaml
```diff
@@ -1,5 +1,6 @@
 env: LOCAL
 cloud: LOCAL
+encryptedFlow: true
```
seehamrun: Just a note, for the encrypted flow, we need to bring up the worker binary separately when testing locally right? Is there already documentation around that?

portability-api/src/main/java/org/dataportabilityproject/webapp/CryptoHelper.java
```diff
@@ -16,14 +16,15 @@
 package org.dataportabilityproject.webapp;
 
 import static org.apache.axis.transport.http.HTTPConstants.HEADER_SET_COOKIE;
+import static org.apache.axis.transport.http.HTTPConstants.HEADER_SET_COOKIE2;
```
seehamrun: remove

portability-core/src/test/java/org/dataportabilityproject/job/JobDaoTest.java
```diff
@@ -0,0 +1,58 @@
+package org.dataportabilityproject.job;
```
seehamrun: Nit: license header missing

portability-worker/src/main/resources/settings/common.yaml
```diff
@@ -1,2 +1,11 @@
 env: LOCAL
```
seehamrun: Isnt this file copied over by the build script?

portability-core/src/main/java/org/dataportabilityproject/job/JobDao.java
```diff
@@ -182,12 +175,18 @@ public void updateJobStateToAssigneWithAuthData(String id, String encryptedExpor
...
 
-    updateJob(updatedJob, JobState.ASSIGNED_WITH_AUTH_DATA);
-    // Delete the job in the previous state
-    String previousKey = createKey(JobState.ASSIGNED_WITHOUT_AUTH_DATA, id);
-    storage.delete(
-        previousKey); // TODO: Determine whether to Tombstone or Delete from unassigned once it's entered in ready state
+  /**
+   * Deletes an existing {@code job} in storage with the given {@code jobState}.
+   */
+  public void deleteJob(String id, JobState jobState) throws IOException {
```
seehamrun: It looks like deleteJob is only called in testing - is this something that we can expose for TestOnly?

portability-core/src/main/java/org/dataportabilityproject/job/JobDao.java
```diff
@@ -39,7 +39,7 @@
    * The value ASSIGNED_WITHOUT_AUTH_DATA indicates the client has submitted all data required, such as the
    * encrypted auth data, in order to begin processing the portability job.
    */
-  enum JobState {
+  public enum JobState {
```
seehamrun: This also looks like (outside of this class) its only used in Test, is there a way we can limit the visibility?

config/environments/local/settings/common.yaml
holachuy: Reverted. This is not ready yet.

portability-api/src/main/java/org/dataportabilityproject/webapp/CryptoHelper.java
holachuy: Done

portability-core/src/main/java/org/dataportabilityproject/job/JobDao.java
holachuy: Annotated with @VisibleForTesting

portability-core/src/main/java/org/dataportabilityproject/job/JobDao.java
holachuy: I think we'll eventually need it when implementing the restart/recovery parts of the lifecycle

portability-core/src/test/java/org/dataportabilityproject/job/JobDaoTest.java
holachuy: Done

portability-worker/src/main/resources/settings/common.yaml
holachuy: Let me add a .gitignore

holachuy:
Thanks for the review. I think I addressed all the comments. PTAL.

portability-worker/src/main/resources/settings/common.yaml
seehamrun: This file still shows up in the diff of the request

portability-core/src/main/java/org/dataportabilityproject/job/JobDao.java
seehamrun: Oh right, that makes sense! Thanks!

PR #34, opened 2018-01-09
hook up XSRF protection in client app

seehamrun:
This uses JWTTokenManager to create an XSRF token and attach it to a cookie.  #8 

This requires a new secret (JWT_KEY, JWT_SECRET) to be added in secrets.csv (for local dev, can be anything) and uploaded to cloud (for deployment).


portability-api/src/main/java/org/dataportabilityproject/webapp/JsonKeys.java
```diff
@@ -34,4 +34,6 @@ private JsonKeys(){} /** no instantiation */
   public static final String IMPORT_SERVICE_AUTH_EXISTS = "importServiceAuthExists";
   public static final String IMPORT_AUTH_URL = "importAuthUrl";
   public static final String TOKEN = "token";
+  public static final String XSRF_TOKEN = "XSRF-TOKEN";
```
holachuy: Are these our own custom values or standard values? if they are standard, we should link to a rfc or paper where they are explained.

portability-api/src/main/java/org/dataportabilityproject/webapp/StartCopyHandler.java
```diff
@@ -70,6 +76,11 @@ public void handle(HttpExchange exchange) throws IOException {
     // Valid job must be present
     String jobId = JobUtils.decodeId(encodedIdCookie);
 
+    // Validate XSRF token is present in request header. strip out any quotation marks that Angular adds and remove whitespace.
```
holachuy: Is there a doc or bug report about this in angular that we can refer to?

portability-api/src/main/java/org/dataportabilityproject/webapp/PortabilityApiMain.java
```diff
@@ -52,6 +55,8 @@ public static void main(String args[]) throws Exception {
     portabilityJobFactory = new PortabilityJobFactory(new UUIDProvider());
     // TODO: Wire up the correct Crypter.
     cryptoHelper = new CryptoHelper(jobDao);
+    // TODO: Wire up correct secret key
+    tokenManager = new JWTTokenManager("supersecretapistring");
```
rtannenbaum: will this be stored in our key management system?

portability-api/src/main/java/org/dataportabilityproject/webapp/JsonKeys.java
seehamrun: These aren't standard but are the default for the angular app. We can use any names for them as long as we hook them up to the client app component. I'll add them explicitly in the angular app and include a comment here that they should match.

portability-api/src/main/java/org/dataportabilityproject/webapp/PortabilityApiMain.java
seehamrun: Yes, this should be stored in the Cloud KMS and not checked into the repository.

portability-api/src/main/java/org/dataportabilityproject/webapp/StartCopyHandler.java
seehamrun: I updated the comment. This is caused by HttpCookie.toString surrounding the cookie value in doublequotes. This is valid cookie syntax and we parse it correctly, however it causes the angular app to surround the header value in double doublequotes (the first of which gets stripped out with the initialization of the http-exchange).

seehamrun:
Added some more expiration details and test for the startCopyHandler (for #5) PTAL

portability-api/src/main/java/org/dataportabilityproject/webapp/PortabilityApiMain.java
```diff
@@ -15,80 +15,14 @@
...
-import org.dataportabilityproject.job.Crypter;
-import org.dataportabilityproject.job.JWTTokenManager;
-import org.dataportabilityproject.job.JobDao;
-import org.dataportabilityproject.job.PortabilityJobFactory;
-import org.dataportabilityproject.job.TokenManager;
-import org.dataportabilityproject.job.UUIDProvider;
+import com.google.inject.Guice;
+import com.google.inject.Injector;
 
 public class PortabilityApiMain {
```
holachuy: Are these your changes to this file? Maybe you need to checkout or revert to the origin/master

portability-api/src/main/java/org/dataportabilityproject/webapp/StartCopyHandler.java
```diff
@@ -48,51 +47,97 @@
...
-    if (PortabilityFlags.encryptedFlow()) {
-      handleWorkerAssignmentFlow(exchange, jobId);
-    } else {
-      handleStartCopyInApi(exchange, jobId);
-    }
+    // Validate XSRF token is present in request header. The cookie value might be surrounded by
+    // double quotes which causes the angular cli to also surround the header with double quotes.
+    // Since the value itself may not contain quotes or whitespace, trim off the double quotes by
+    // converting them to whitespace.
+    String tokenHeader = requestHeaders.getFirst(JsonKeys.XSRF_HEADER)
```
holachuy: Should this be a utility method in a utils class with the javadoc explaining it?

portability-api/src/test/java/org/dataportabilityproject/webapp/StartCopyHandlerTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+import com.sun.net.httpserver.Headers;
+import org.dataportabilityproject.job.JWTTokenManager;
+import org.dataportabilityproject.job.JobUtils;
+import org.dataportabilityproject.job.PortabilityJob;
+import org.dataportabilityproject.job.TokenManager;
+import org.junit.Before;
+import org.junit.Test;
+
+public class StartCopyHandlerTest {
```
holachuy: Thanks for adding this test!   I'm wondering, if we only ever call validateJobId() for token verification, if that logic could be in a utility class or helper class that we can test directly without bringing in the whole handler.

portability-core/src/main/java/org/dataportabilityproject/job/JWTTokenManager.java
```diff
@@ -74,7 +85,7 @@ public String createNewToken(String uuid) {
     try {
       return JWT.create()
           .withIssuer(JWTTokenManager.ISSUER)
-          .withClaim(JWTTokenManager.ID_CLAIM_KEY, uuid)
+          .withClaim(JWTTokenManager.ID_CLAIM_KEY, uuid).withExpiresAt(new Date(System.currentTimeMillis()+EXPIRATION_TIME_MILLIS))
```
holachuy: space before and after plus

portability-api/src/main/java/org/dataportabilityproject/webapp/PortabilityApiMain.java
seehamrun: Hm I think this might be from looking at  a diff before I had merged willard's change. In the UI there should be a button to view diffs from all commits?

portability-api/src/main/java/org/dataportabilityproject/webapp/StartCopyHandler.java
```diff
@@ -38,44 +40,54 @@
...
     this.jobDao = jobDao;
     this.cloudFactory = cloudFactory;
     this.cryptoHelper = cryptoHelper;
     this.commonSettings = commonSettings;
+    this.tokenManager = tokenManager;
+  }
+
+  /* Test constructor used only for verification of xsrf token flow */
+  @VisibleForTesting
+  StartCopyHandler(TokenManager tokenManager) {
```
holachuy: Looks like this is no longer needed?

seehamrun: Right deleted.

Issue #22, opened 2018-01-07
Set up CI build & test for pull requests


bwillard:
We should ensure that all PRs build and pass tests before submitting.

rtannenbaum:
Brian did an initial investigation, based on that I think we should try the free Travis CI edition to start.

Some notes:
- Free edition does not require code write access, whereas Paid edition might, due to permissions not being granular enough. (Does require webhook access; Github's API for responding to repo events)
- Travis supports Node.js whereas others (like Circle) do not. We might need Node.js support for our angular frontend, which has some dependencies on Node even though we don't use it directly
- Uses a shared worker pool for the rest of the org (google) so we may run into latency problems, but we may move our repo to a different org at some point.

rtannenbaum:
Travis instructions for java projects. Includes both Maven and Gradle examples. https://docs.travis-ci.com/user/languages/java/

rtannenbaum:
Forgot we still need to test frontend, will reopen this

rtannenbaum:
@holachuy once we get the Jasmine/Karma tests passing it looks pretty easy to get Travis to run 'ng test'.

I sent a PR to fix a couple initial issues but I think I'd need to read up a bunch on Angular, Jasmine and Karma to get any further. Would it make more sense for you to fix them?

PR #36, opened 2018-01-10
Create .travis.yml

rtannenbaum:
for #22.

Travis should automatically detect that we use Maven based on the presence of pom.xml.


rtannenbaum:
Woot, it ran mvn test

Issue #25, opened 2018-01-07
Limit OAuth scope requests to specific verticals


bwillard:
Right now we ask for all Oauth scopes that *might* be needed for a particular service.

Instead we should ask for just the scope needed.

e.g. if just exporting photos from Google we shouldn't ask for mail permissions, or any kind of write permision.

seehamrun:
Deployed this to our cloud instance and looks working as intended

PR #46, opened 2018-01-12
Add ServiceMode enum for passing around instead of boolean isExport.

seehamrun:
This is in prep for limiting Oauth scopes #25  - idea is in addition to passing dataType to serviceProviderRegistry.getOnlineAuthDataGenerator, we would pass in the service mode.  Then each service provider can determine what generator to return based on the pair.

### DIFF
--- a/portability-core/src/main/java/org/dataportabilityproject/job/JobUtils.java
+++ b/portability-core/src/main/java/org/dataportabilityproject/job/JobUtils.java
@@ -22,6 +22,7 @@ import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import com.google.common.io.BaseEncoding;
 import org.dataportabilityproject.shared.PortableDataType;
+import org.dataportabilityproject.shared.ServiceMode;
 import org.dataportabilityproject.shared.auth.AuthData;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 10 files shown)


portability-core/src/main/java/org/dataportabilityproject/shared/ServiceMode.java
```diff
@@ -0,0 +1,9 @@
+package org.dataportabilityproject.shared;
+
+/**
+ * The possible service modes for each authorization
```
holachuy: How about the possible service modes for transfering data. 

In case we use for more than auth later?

portability-core/src/main/java/org/dataportabilityproject/job/JobUtils.java
```diff
@@ -42,32 +43,38 @@ public static String encodeId(PortabilityJob job) {
...
-      updatedJob.setExportInitialAuthData(initialAuthData);
-    } else {
-      updatedJob.setImportInitialAuthData(initialAuthData);
+    switch (serviceMode) {
+      case EXPORT:
+        updatedJob.setExportInitialAuthData(initialAuthData);
+        break;
+      case IMPORT:
+        updatedJob.setImportInitialAuthData(initialAuthData);
+        break;
```
holachuy: Add case default that throws exception "Unsupported ServiceMode"

seehamrun: Done.

PR #47, opened 2018-01-12
Add overloaded function for getOnlineAuthData that specifies ServiceMode

seehamrun:
#25

### DIFF
--- a/portability-core/src/main/java/org/dataportabilityproject/shared/ServiceProvider.java
+++ b/portability-core/src/main/java/org/dataportabilityproject/shared/ServiceProvider.java
@@ -48,8 +48,9 @@ public interface ServiceProvider {
     }
 
     /* Same as above, but retrieves AuthGenerator for read/write access (ignoring serviceMode)
-     * TODO: deprecate once all serviceProviders implement the above version of this method.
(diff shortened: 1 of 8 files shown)


portability-core/src/main/java/org/dataportabilityproject/shared/ServiceProvider.java
```diff
@@ -38,18 +39,30 @@
...
+    /* Return an OnlineAuthDataGenerator for the provided dataType. The returned generator will have permissions based
+     * on the serviceMode provided.
+     */
+    default OnlineAuthDataGenerator getOnlineAuthDataGenerator(PortableDataType dataType, ServiceMode serviceMode) {
+        // default implementation is to ignore serviceMode
+        return getOnlineAuthDataGenerator(dataType);
+    }
+
+    /* Same as above, but retrieves AuthGenerator for read/write access (ignoring serviceMode)
+     * TODO: deprecate once all serviceProviders implement the above version of this method.
```
holachuy: Might as well mark as @deprecated now

seehamrun: Done

PR #51, opened 2018-01-16
Limit oauth scope for flickr import/export.

seehamrun:
#25 

Tested with local deployment.


portability-core/src/main/java/org/dataportabilityproject/serviceProviders/flickr/FlickrAuth.java
```diff
@@ -78,7 +81,8 @@ public AuthFlowInitiator generateAuthUrl(String callbackBaseUrl, String id) thro
     AuthInterface authInterface = flickr.getAuthInterface();
     Token token = authInterface.getRequestToken(
         callbackBaseUrl + "/callback1/flickr");
-    String url = authInterface.getAuthorizationUrl(token, Permission.WRITE);
+    String url = authInterface.getAuthorizationUrl(token,
```
holachuy: Nice

PR #52, opened 2018-01-16
Limit oauth scope requested for google APIs based on dataType and serviceMode

seehamrun:
#25


portability-core/src/main/java/org/dataportabilityproject/serviceProviders/google/GoogleServiceProvider.java
```diff
@@ -57,16 +62,41 @@
...
     @Inject
     GoogleServiceProvider(AppCredentialFactory appCredentialFactory) throws Exception {
         AppCredentials appCredentials =
             appCredentialFactory.lookupAndCreate("GOOGLE_KEY", "GOOGLE_SECRET");
-        this.googleAuth = new GoogleAuth(
-                appCredentials,
-                // TODO: only use scopes from the products we are accessing.
-                SCOPES);
+
+        for (Entry<PortableDataType, List<String>> e : dataTypeImportScopes.entrySet()) {
```
holachuy: Instead of creating all of them, could we create the right one on demand, e.g. lazy init or users a Suppliers.memoize, as I don't think we need to support multiple exports and imports at the same time.

seehamrun: How does this look? I'm not sure if the nested Map<> def is preferred over having two separate maps for import and export from a readability perspective.

PR #55, opened 2018-01-16
Fix angular tests 

holachuy:
Fix angular tests so they all pass including testing component creation and fixing form template parsing errors. Issues  #5 #22


portability-core/src/test/java/org/dataportabilityproject/job/JobDaoTest.java
```diff
@@ -34,7 +35,7 @@
 
   @Before
   public void setUp() {
-    jobDao = new JobDao(new LocalCloudFactory().getPersistentKeyValueStore());
```
rtannenbaum: should this be in a separate PR?

holachuy: Done! Good catch.

PR #56, opened 2018-01-16
Fix local environment pointing to GCP & update documentation

rtannenbaum:
Now that we're on Guice, also inject project ID into Google implementation classes instead of using static util method

Store keys & secrets in a LoadingCache since they were being read a ton of times on startup

Tested locally pointing to cloud=GOOGLE, and pointing to LOCAL verified we don't initialize any of the Google impls.

### DIFF
--- a/portability-core/src/main/java/org/dataportabilityproject/PortabilityCoreModule.java
+++ b/portability-core/src/main/java/org/dataportabilityproject/PortabilityCoreModule.java
@@ -28,6 +28,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import org.dataportabilityproject.cloud.SupportedCloud;
 import org.dataportabilityproject.cloud.google.GoogleCloudFactory;
+import org.dataportabilityproject.cloud.google.GoogleCloudModule;
 import org.dataportabilityproject.cloud.interfaces.CloudFactory;
 import org.dataportabilityproject.cloud.local.LocalCloudFactory;
 import org.dataportabilityproject.job.JWTTokenManager;
@@ -44,6 +45,11 @@ import org.dataportabilityproject.shared.local.LocalAppCredentialFactory;
 import org.dataportabilityproject.shared.settings.CommonSettings;
 
 public final class PortabilityCoreModule extends AbstractModule {
+  private final CommonSettings commonSettings;
+
+  public PortabilityCoreModule() {
+    this.commonSettings = getCommonSettings();
+  }
 
   @Override
   protected void configure() {
@@ -54,11 +60,19 @@ public final class PortabilityCoreModule extends AbstractModule {
     install(new MicrosoftModule());
     install(new RememberTheMilkModule());
     install(new SmugmugModule());
+
+    if (commonSettings.getCloud() == SupportedCloud.GOOGLE) {
+      install(new GoogleCloudModule());
+    }
   }
 
   @Singleton
   @Provides
   CommonSettings provideCommonSettings() {
+    return commonSettings;
+  }
+
+  private CommonSettings getCommonSettings() {
     ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
     try {
       InputStream in =
@@ -85,7 +99,6 @@ public final class PortabilityCoreModule extends AbstractModule {
       throw new UnsupportedOperationException(commonSettings.getCloud() + " is not supported yet.");
     }
   }
-
 
   @Provides
   AppCredentialFactory provideAppCredentialFactory(
--- a/portability-core/src/main/java/org/dataportabilityproject/cloud/google/GoogleCloudFactory.java
+++ b/portability-core/src/main/java/org/dataportabilityproject/cloud/google/GoogleCloudFactory.java
@@ -23,8 +23,6 @@ import com.google.cloud.datastore.Key;
 import com.google.cloud.datastore.Query;
 import com.google.cloud.datastore.QueryResults;
 import com.google.cloud.datastore.StructuredQuery.PropertyFilter;
-import com.google.common.base.Preconditions;
-import com.google.common.base.Strings;
 import javax.inject.Inject;
 import org.dataportabilityproject.cloud.interfaces.BucketStore;
 import org.dataportabilityproject.cloud.interfaces.CloudFactory;
@@ -33,29 +31,28 @@ import org.dataportabilityproject.cloud.interfaces.JobDataCache;
 import org.dataportabilityproject.cloud.interfaces.PersistentKeyValueStore;
 
 public final class GoogleCloudFactory implements CloudFactory {
-  // Lazy init this in case we are running a different cloud than Google, in which case this class
-  // won't be used and the environment variable this is set from won't be available.
-  private static String PROJECT_ID;
-
   private final Datastore datastore;
   private final PersistentKeyValueStore persistentKeyValueStore;
   private final CryptoKeyManagementSystem cryptoKeyManagementSystem;
   private final BucketStore bucketStore;
+  private final String projectId;
 
   @Inject
   public GoogleCloudFactory(
       GoogleBucketStore googleBucketStore,
       GoogleCredentials googleCredentials,
-      GoogleCryptoKeyManagementSystem googleCryptoKeyManagementSystem) {
+      GoogleCryptoKeyManagementSystem googleCryptoKeyManagementSystem,
+      ProjectId projectId) {
     this.datastore = DatastoreOptions
         .newBuilder()
-        .setProjectId(getGoogleProjectId())
+        .setProjectId(projectId.getProjectId())
         .setCredentials(googleCredentials)
         .build()
         .getService();
     this.persistentKeyValueStore = new GooglePersistentKeyValueStore(datastore);
     this.cryptoKeyManagementSystem = googleCryptoKeyManagementSystem;
     this.bucketStore = googleBucketStore;
+    this.projectId = projectId.getProjectId();
   }
 
   @Override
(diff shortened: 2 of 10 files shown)


portability-core/src/main/java/org/dataportabilityproject/cloud/google/GoogleCloudModule.java
```diff
@@ -59,4 +60,23 @@ GoogleCredentials getCredentials(CommonSettings commonSettings) throws GoogleCre
     }
   }
 
+  /**
+   * Get project ID from environment variable and validate it is set.
+   *
+   * @throws IllegalArgumentException if project ID is unset
+   */
+  @Provides ProjectId getProjectId() {
```
holachuy: Is this only used for local talking to cloud or also when deployed on the docker image. If the former, we might want to created a separate module and annotation for local only.

portability-core/src/main/java/org/dataportabilityproject/shared/settings/CommonSettings.java
```diff
@@ -60,9 +60,10 @@ public CommonSettings(
       // locally and connecting to GCP
       String googleProjectId = System.getenv("GOOGLE_PROJECT_ID");
```
holachuy: Do we read this twice from the environment variable now?

portability-core/src/main/java/org/dataportabilityproject/shared/CloudAppCredentialFactory.java
```diff
@@ -38,24 +55,59 @@
...
+          }
+        });
   }
 
   @Override
   public AppCredentials lookupAndCreate(String keyName, String secretName) throws IOException {
+    String key;
+    String secret;
+    try {
+      key = keys.get(keyName);
```
holachuy: Could we do each of these in it's own try block so we can get a better exception?

portability-core/src/main/java/org/dataportabilityproject/PortabilityCoreModule.java
```diff
@@ -54,11 +60,19 @@ protected void configure() {
     install(new MicrosoftModule());
     install(new RememberTheMilkModule());
     install(new SmugmugModule());
+
+    if (commonSettings.getCloud() == SupportedCloud.GOOGLE) {
```
holachuy: Can we get away with installing this in a different way. We're bypassing the guicified CommonSettings to find out if we have a specific cloud value.

portability-core/src/main/java/org/dataportabilityproject/cloud/google/GoogleCloudModule.java
rtannenbaum: It's used for the Docker image too. Not a local-only thing

portability-core/src/main/java/org/dataportabilityproject/PortabilityCoreModule.java
rtannenbaum: Done, this isn't great so let me know if you can think of a better way

portability-core/src/main/java/org/dataportabilityproject/shared/CloudAppCredentialFactory.java
rtannenbaum: Done

portability-core/src/main/java/org/dataportabilityproject/shared/settings/CommonSettings.java
rtannenbaum: I just made ProjectId a singleton in the module to avoid that

holachuy: Looking at it closer, maybe we should avoid reading a Google cloud specific value, wether in environment or singleton,  in the CommonSettings. Can we move this logic elsewhere?

portability-core/src/main/java/org/dataportabilityproject/PortabilityCoreModule.java
holachuy: Can GoogleCloudModule declare a RequireBinding for the CommonSettings?

rtannenbaum: good idea, done!

portability-core/src/main/java/org/dataportabilityproject/shared/settings/CommonSettings.java
rtannenbaum: good idea, moved it to GoogleCloudModule and got rid of the extra env variable read

portability-core/src/main/java/org/dataportabilityproject/shared/CloudAppCredentialFactory.java
```diff
@@ -38,24 +55,64 @@
...
     logger.debug("Getting key {} (blob {}) from bucket {}", keyName, keyLocation, bucketName);
     byte[] rawKeyBytes = bucketStore.getBlob(bucketName, keyLocation);
     String key = new String(rawKeyBytes);
     checkState(!Strings.isNullOrEmpty(key), "Couldn't lookup: " + keyName);
+    return key;
+  }
+
+  private String lookupSecret(String secretName) throws IOException {
+    String projectId = cloudFactory.getProjectId();
+    String bucketName = BUCKET_NAME_PREFIX + projectId;
```
holachuy: Could we do this without project id and just have bucketstore internally know the name of the bucket that way each cloud bucketstore doesn't have to follow a specific convention at this layer?

portability-core/src/main/java/org/dataportabilityproject/cloud/google/ProjectId.java
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Wrapper class for a Google Cloud Platform project ID.
+ *
+ * <p>This class exists so that we can inject ProjectId into Google implementation classes without
+ * forcing Guice to eagerly eval Provider<GoogleCloudFactory> when not using cloud=GOOGLE. i.e. if
+ * we just bound the project ID as a String directly ("@ProjectId String projectId") via
+ * bind(String.toClass).annotatedWith(ProjectId.class).toInstance(...), that causes eager init.
+ */
+class ProjectId {
```
holachuy: Could we use Guice bind(String.class).annotatedWith(ProjectId.class).to("the env var") to avoid having a wrapper class?

portability-core/src/main/java/org/dataportabilityproject/cloud/google/GoogleCloudModule.java
```diff
@@ -59,4 +75,36 @@ GoogleCredentials getCredentials(CommonSettings commonSettings) throws GoogleCre
     }
   }
 
+  /**
+   * Get project ID from environment variable and validate it is set.
+   *
+   * @throws IllegalArgumentException if project ID is unset
+   */
+  @Provides @Singleton ProjectId getProjectId(CommonSettings commonSettings) {
```
holachuy: Could we use Guice bind(String.class).annotatedWith(ProjectId.class).to("the env var") to avoid having a wrapper class?

portability-core/src/main/java/org/dataportabilityproject/cloud/google/GoogleCloudFactory.java
```diff
@@ -94,34 +91,6 @@ public void clearJobData(String jobId) {
...
-      throw new IllegalArgumentException("Need to specify a project ID when using Google Cloud. "
-          + "This should be exposed as an environment variable by Kubernetes, see "
-          + "k8s/api-deployment.yaml");
-    }
-    Preconditions.checkArgument(!Strings.isNullOrEmpty(tempProjectId), "Need to specify a project "
-        + "ID when using Google Cloud. This should be exposed as an environment variable by "
-        + "Kubernetes, see k8s/api-deployment.yaml");
-    PROJECT_ID = tempProjectId.toLowerCase();
-    return PROJECT_ID;
+    return projectId;
```
holachuy: Can we just get rid of CloudFactory.getProjectId if it's only used for bucket name?

rtannenbaum: how's this? it's used for crypto key too

portability-core/src/main/java/org/dataportabilityproject/shared/CloudAppCredentialFactory.java
rtannenbaum: I thought about this too but thought it could be good to have a convention for bucket names if it makes sense for other providers. Can we leave as is for now, and if another cloud provider comes along with a different convention, we can refactor to be impl specific?

portability-core/src/main/java/org/dataportabilityproject/cloud/google/GoogleCloudFactory.java
rtannenbaum: it's used for crypto key too. Do you bring that up because it sounds too Google-specific to have in CloudFactory? It's meant to be general since any cloud platform should have some notion of this. Would you prefer a more generic name than project ID? (If so, what?)

portability-core/src/main/java/org/dataportabilityproject/shared/CloudAppCredentialFactory.java
holachuy: It just seems to me not imposing a convention  makes it easier than having another rule for a cloud provider or developer to follow. It seems the easiest onramp is to have the script create or takein a single flag, convention free.

portability-core/src/main/java/org/dataportabilityproject/cloud/google/GoogleCloudFactory.java
holachuy: It seems to be used for forming the cryptoKeyName to pass to the key management system. 

CRYPTO_KEY_FMT_STRING seems Google specific? Can we just construct the path and use projectid or whatever, internally in the cloud specific implementation of CryptoKeyManagementSystem

rtannenbaum: OK, now that I think about this more, the crypto key format is pretty Google-specific and the bucket name is too actually, since there is that weird requirement that bucket names have to be unique across Google projects. That's why we include project ID in the bucket name which people probably wouldn't do normally. So I agree the crypto key and bucket name conventions should be inside the Google impl's, I moved them in the latest commit

portability-core/src/main/java/org/dataportabilityproject/cloud/google/GoogleCloudModule.java
rtannenbaum: I just tried @ProjectId String projectId again with the old binding and it worked. Not sure what I was doing before that caused eager eval. Thanks for bringing it up, it is much better without the wrapper class!

portability-core/src/main/java/org/dataportabilityproject/cloud/google/ProjectId.java
rtannenbaum: Done, see other reply

portability-core/src/main/java/org/dataportabilityproject/shared/CloudAppCredentialFactory.java
rtannenbaum: Done, see my reply to your other comment re crypto key

rtannenbaum:
Thanks for the thorough review!

PR #59, opened 2018-01-17
Update MicrosoftServiceProvider to return AuthGenerator based on serviceMode.

seehamrun:
Also rename combine Offline and Online PasswordAuthDataGenerators into
one class to be consistent with other service provider generators.

Note that this does not actually limit the scope. This is only migrating
them to the new version of the function. ( Setting up for #25 )


portability-core/src/main/java/org/dataportabilityproject/serviceProviders/microsoft/MicrosoftServiceProvider.java
```diff
@@ -44,9 +44,9 @@
         "wl.imap", // outlook export via IMAP
         "wl.offline_access", // provides for refresh tokens
         "wl.calendars", "wl.contacts_calendars"); // calendar export
+
     private final MicrosoftAuth microsoftAuth;
-    private final OfflinePasswordAuthDataGenerator passwordAuth =
-        new OfflinePasswordAuthDataGenerator();
+    private final PasswordAuthDataGenerator  passwordAuth;
```
rtannenbaum: nit: extra space? (here & below)

portability-core/src/main/java/org/dataportabilityproject/serviceProviders/microsoft/MicrosoftServiceProvider.java
```diff
@@ -73,24 +74,30 @@
...
 
     @Override
-    public OfflineAuthDataGenerator getOfflineAuthDataGenerator(PortableDataType dataType) {
+    public OfflineAuthDataGenerator getOfflineAuthDataGenerator(PortableDataType dataType, ServiceMode serviceMode) {
         switch (dataType) {
             case CALENDAR:
+                // IMPORT and EXPORT are supported for CALENDAR
                 return microsoftAuth;
             case MAIL:
+                // Only EXPORT is supported for MAIL
```
rtannenbaum: do we purposely not implement import or just not yet? can you document / add a TODO to implement?

portability-core/src/main/java/org/dataportabilityproject/serviceProviders/microsoft/MicrosoftServiceProvider.java
```diff
@@ -44,9 +44,9 @@
         "wl.imap", // outlook export via IMAP
         "wl.offline_access", // provides for refresh tokens
         "wl.calendars", "wl.contacts_calendars"); // calendar export
+
     private final MicrosoftAuth microsoftAuth;
```
rtannenbaum: I am not very familiar with this code, but it's confusing to me that MicrosoftAuth implements OfflineAuthDataGenerator, OnlineAuthDataGenerator, and PasswordAuthDataGenerator also implements OnlineAuthDataGenerator, OfflineAuthDataGenerator. Why do we need both variables?

seehamrun: PasswordAuth is the plaintext username/password authenticator that is used for reading calendar via IMAP (I think)
Whereas the MicrosoftAuth is what uses the Oauth flow for microsoft and requests the appropriate scopes (TODO). 

I'm trying to figure out why the Password auth was used for mail, given that I think oauth flow should give us access to mail: 
image

but havent had a chance to debug the flow. @holachuy are you familiar with the reason we're using PasswordAuth for Microsoft mail?

portability-core/src/main/java/org/dataportabilityproject/serviceProviders/microsoft/MicrosoftServiceProvider.java
seehamrun: Done.

portability-core/src/main/java/org/dataportabilityproject/serviceProviders/microsoft/MicrosoftServiceProvider.java
seehamrun: I think its purposefully. I'll refactored the way we figure out if a service is supported or and added a TODO for import
PTAL.

portability-core/src/main/java/org/dataportabilityproject/serviceProviders/microsoft/MicrosoftServiceProvider.java
rtannenbaum: ah I remember hearing that MSFT doesn't support OAuth for email. Maybe HTTP API is different than IMAP API. In any case, can you rename the variables consistently (passwordAuth/oauthAuth, mailAuth/calendarAuth etc) or document that one is used for calendar, the other for email?

portability-core/src/main/java/org/dataportabilityproject/serviceProviders/microsoft/MicrosoftServiceProvider.java
```diff
@@ -44,18 +46,27 @@
...
-    private final OfflinePasswordAuthDataGenerator passwordAuth =
-        new OfflinePasswordAuthDataGenerator();
+    private final PasswordAuthDataGenerator passwordAuth;
+    private final AppCredentials appCredentials;
+
+    // The list of supported dataTypes for import and export.
+    // TODO: support IMPORT for MAIL.
+    private final Map<ServiceMode, ImmutableList<PortableDataType>> supportedDataTypes = ImmutableMap.<ServiceMode, ImmutableList<PortableDataType>>builder()
```
holachuy: static

portability-core/src/main/java/org/dataportabilityproject/serviceProviders/microsoft/MicrosoftServiceProvider.java
```diff
@@ -64,16 +75,18 @@
...
     public ImmutableList<PortableDataType> getImportTypes() {
-        return ImmutableList.of(PortableDataType.CALENDAR);
+        return supportedDataTypes.get(ServiceMode.IMPORT);
     }
 
     @Override
-    public OfflineAuthDataGenerator getOfflineAuthDataGenerator(PortableDataType dataType) {
+    public OfflineAuthDataGenerator getOfflineAuthDataGenerator(PortableDataType dataType, ServiceMode serviceMode) {
+        Preconditions.checkArgument(supportedDataTypes.get(serviceMode).contains(dataType),
+            "DataType %s not supported for provided ServiceMode: %s", dataType, serviceMode);
```
holachuy: same here around %s

portability-core/src/main/java/org/dataportabilityproject/serviceProviders/microsoft/MicrosoftServiceProvider.java
```diff
@@ -116,6 +133,8 @@ public OnlineAuthDataGenerator getOnlineAuthDataGenerator(PortableDataType dataT
         PortableDataType type,
         AuthData authData,
         JobDataCache jobDataCache) throws IOException {
+        Preconditions.checkArgument(supportedDataTypes.get(ServiceMode.IMPORT).contains(type),
+            "DataType %s not supported for IMPORT", type);
```
holachuy: same here around %s

portability-core/src/main/java/org/dataportabilityproject/serviceProviders/microsoft/MicrosoftServiceProvider.java
seehamrun: Done, renamed to oauthAuth and passwordAuth since when we add more data types, they may be used for other things.

portability-core/src/main/java/org/dataportabilityproject/serviceProviders/microsoft/MicrosoftServiceProvider.java
seehamrun: Done

portability-core/src/main/java/org/dataportabilityproject/serviceProviders/microsoft/MicrosoftServiceProvider.java
seehamrun: Done

portability-core/src/main/java/org/dataportabilityproject/serviceProviders/microsoft/MicrosoftServiceProvider.java
seehamrun: Done

PR #60, opened 2018-01-17
Limit scopes for AuthDataGenerators for MicrosoftServiceProvider when importing/exporting calendar data.

seehamrun:
prompt when exporting cal data from microsoft: 
image

prompt when importing cal data to microsoft: 
image

I deleted all the extra scopes as well since those aren't supported in the oauth flow (yet).

### DIFF
--- a/portability-core/src/main/java/org/dataportabilityproject/serviceProviders/microsoft/MicrosoftServiceProvider.java
+++ b/portability-core/src/main/java/org/dataportabilityproject/serviceProviders/microsoft/MicrosoftServiceProvider.java
@@ -20,6 +20,9 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.inject.Inject;
 import java.io.IOException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import org.dataportabilityproject.cloud.interfaces.JobDataCache;
 import org.dataportabilityproject.dataModels.DataModel;
@@ -42,15 +45,9 @@ import org.dataportabilityproject.shared.auth.PasswordAuthDataGenerator;
  * The {@link ServiceProvider} for Microsoft (http://www.microsoft.com/).
  */
 final class MicrosoftServiceProvider implements ServiceProvider {
-    private static final ImmutableList<String> SCOPES = ImmutableList.of(
-        "wl.imap", // outlook export via IMAP
-        "wl.offline_access", // provides for refresh tokens
-        "wl.calendars", "wl.contacts_calendars"); // calendar export
-
-    // oauthAuth is used for calendar import and export
-    private final MicrosoftAuth oauthAuth;
     // passwordAuth is used for mail export
     private final PasswordAuthDataGenerator passwordAuth;
+
     // credentials to pass into AuthDataGenerators
     private final AppCredentials appCredentials;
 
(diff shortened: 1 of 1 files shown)


portability-core/src/main/java/org/dataportabilityproject/serviceProviders/microsoft/MicrosoftServiceProvider.java
```diff
@@ -61,14 +58,21 @@
...
+    private final static Map<PortableDataType, Map<ServiceMode, List<String>>> DATA_TYPE_SCOPES =
+        ImmutableMap.<PortableDataType, Map<ServiceMode, List<String>>>builder()
+            .put(PortableDataType.CALENDAR, ImmutableMap.<ServiceMode, List<String>>builder()
+                .put(ServiceMode.EXPORT, Arrays.asList("wl.calendars", "wl.contacts_calendars"))
+                .put(ServiceMode.IMPORT, Arrays.asList("wl.calendars_update"))
+                .build())
+            // TODO: add support for more service scopes here
+            .build();
+
+    private Map<PortableDataType, Map<ServiceMode, MicrosoftAuth>> dataTypeAuths = new HashMap<>();
```
holachuy: final static as well?

portability-core/src/main/java/org/dataportabilityproject/serviceProviders/microsoft/MicrosoftServiceProvider.java
```diff
@@ -61,14 +58,21 @@
         .put(ServiceMode.EXPORT, ImmutableList.of(PortableDataType.CALENDAR, PortableDataType.MAIL))
         .build();
 
+    private final static Map<PortableDataType, Map<ServiceMode, List<String>>> DATA_TYPE_SCOPES =
+        ImmutableMap.<PortableDataType, Map<ServiceMode, List<String>>>builder()
```
holachuy: This could be an Map<PortableDataType, ArrayListMultimap<SeviceMode, String>>

seehamrun: Done. How does this look?

portability-core/src/main/java/org/dataportabilityproject/serviceProviders/microsoft/MicrosoftServiceProvider.java
seehamrun: Done

portability-core/src/main/java/org/dataportabilityproject/serviceProviders/microsoft/MicrosoftServiceProvider.java
```diff
@@ -61,14 +59,21 @@
         .put(ServiceMode.EXPORT, ImmutableList.of(PortableDataType.CALENDAR, PortableDataType.MAIL))
         .build();
 
+    private final static Map<PortableDataType, ListMultimap<ServiceMode, String>> DATA_TYPE_SCOPES =
+        ImmutableMap.<PortableDataType, ListMultimap<ServiceMode, String>>builder()
+            .put(PortableDataType.CALENDAR, ImmutableListMultimap.<ServiceMode, String>builder()
+                .putAll(ServiceMode.EXPORT, Arrays.asList("wl.calendars", "wl.contacts_calendars"))
```
holachuy: Not sure if it's any cleaner but with multipmaps you can continue to add values to the same key. Whatever you think is cleaner to read, even if you don't want to use multimaps, just a suggestion.
.put(ServiceMode.EXPORT, "wl.calendars")
.put(ServiceMode.EXPORT,  "wl.contacts_calendars")

seehamrun: Oh I see - I was thinking it was cleaner and less error prone to put them all on one line. So if the scopes change, the place to modify is clearer.

PR #62, opened 2018-01-17
Request permissions for Smugmug and RememberTheMilk based on serviceMode.

seehamrun:
These are the last two service providers not using serviceMode #25


portability-core/src/main/java/org/dataportabilityproject/serviceProviders/rememberTheMilk/RememberTheMilkServiceProvider.java
```diff
@@ -34,16 +36,18 @@
...
 
     @Inject
     RememberTheMilkServiceProvider(AppCredentialFactory appCredentialFactory) throws IOException {
         this.appCredentials = appCredentialFactory.lookupAndCreate("RTM_KEY", "RTM_SECRET");
-        this.rememberTheMilkAuth = new RememberTheMilkAuth(
-            new RememberTheMilkSignatureGenerator(
-            appCredentials,
-            null
-        ));
+        this.importAuth = new RememberTheMilkAuth(
```
holachuy: I'm not sure what our final instantiation model will be but I rather lazy init these rather than construct in the constructor in case this service is not used as part of a job.

seehamrun: done.

PR #63, opened 2018-01-17
Delete deprecated versions of get*AuthDataGenerator that don't use serviceMode.

seehamrun:
For #25

### DIFF
--- a/portability-core/src/main/java/org/dataportabilityproject/shared/ServiceProvider.java
+++ b/portability-core/src/main/java/org/dataportabilityproject/shared/ServiceProvider.java
@@ -16,13 +16,11 @@
 package org.dataportabilityproject.shared;
 
 import com.google.common.collect.ImmutableList;
+import java.io.IOException;
 import org.dataportabilityproject.cloud.interfaces.JobDataCache;
 import org.dataportabilityproject.dataModels.DataModel;
 import org.dataportabilityproject.dataModels.Exporter;
 import org.dataportabilityproject.dataModels.Importer;
-
-import java.io.IOException;
-
 import org.dataportabilityproject.shared.auth.AuthData;
 import org.dataportabilityproject.shared.auth.OfflineAuthDataGenerator;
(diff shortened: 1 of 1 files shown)


portability-core/src/main/java/org/dataportabilityproject/shared/ServiceProvider.java
```diff
@@ -46,37 +46,19 @@
      * have permissions based on the serviceMode provided.
      */
     default OfflineAuthDataGenerator getOfflineAuthDataGenerator(PortableDataType dataType, ServiceMode serviceMode) {
-      // default implementation to ignore serviceMode
-      return getOfflineAuthDataGenerator(dataType);
+      logger.debug("WARNING: getOnlineAuthDataGenerator not implemented for type: {}, serviceMode: {}, service: {}",
```
holachuy: how about logger.warn( ?

seehamrun: Done.

portability-core/src/main/java/org/dataportabilityproject/shared/ServiceProvider.java
```diff
@@ -46,37 +46,19 @@
      * have permissions based on the serviceMode provided.
      */
     default OfflineAuthDataGenerator getOfflineAuthDataGenerator(PortableDataType dataType, ServiceMode serviceMode) {
-      // default implementation to ignore serviceMode
-      return getOfflineAuthDataGenerator(dataType);
+      logger.warn("WARNING: getOnlineAuthDataGenerator not implemented for type: {}, serviceMode: {}, service: {}",
```
holachuy: One last thing, you can remove the WARNING: from the log statement as the warn() method will prepend that to the output. Here and below.

seehamrun: Good catch! Done.

PR #66, opened 2018-01-18
Update qa's index.html to reflect new static content updated.

seehamrun:


seehamrun:
Agreed, we can look to see if theres a more automated way to do so, but in the meantime I filed https://github.com/google/data-portability/issues/67

PR #68, opened 2018-01-18
Reformat ServiceProvider directories based on IntelliJ autoformatter

seehamrun:
This touches a lot of files, but nothing was touched manually and each subdirectory is its own commit. Let me know if you want me to split it out into separate pull requests.

A lot of this is Intellij moving around functions (due to I think a combination of ABC order and order that they're declared in the interface?) and spaces (moving to consistent indentation with 2 spaces).


seehamrun:
Tested locally with various PHOTOS combinations and things look as expected.

PR #74, opened 2018-01-22
Remove debug condition. Forgot to add to previous commit

holachuy:
Forgot to push this per your comment on the previous commit.


portability-worker/src/main/java/org/dataportabilityproject/worker/WorkerImpl.java
```diff
@@ -88,11 +88,6 @@ private void processJob(PortabilityJob job) {
     PortableDataType dataType = PortableDataType.valueOf(job.dataType());
     try {
       try {
```
seehamrun: Do we need two try statements here?

holachuy: Nope, you're right. Fixed.

Issue #71, opened 2018-01-22
Modularize data types with new gradle structure


seehamrun:
blocks #21 (switch from maven to gradle)

Issue #67, opened 2018-01-18
Tracker for frontend client changes


seehamrun:
General catch-all tracker for any front end client changes. Any front end changes require a new static content deployment to cloud (build script under /config/gcp/build_and_deploy_static_content.sh) as well as a new binary build and rollout.

lisad:
Tracker issue no longer needed, closing

PR #80, opened 2018-01-24
Use new DataTypes module in angular front end 

seehamrun:
This switches over to using angulars http client for the submission of the configure form (instead of htmls form submit). 

Switching over the angulars http submit means that the redirects are no longer blindly followed by the browser so as a workaround, I modified the type returned by the configure handler to be a DataTransfer and let the app redirect explicitly to it. This is similar to the way we redirect for the import auth flow and for copy setup. 

Note: I manually generated the interface files. I'll look to see if there are tools that will create these for us. 

#71 

Note: This requires a static content deployment and a new image deploy (#67)


client/proxy.conf.json
```diff
@@ -1,4 +1,9 @@
 {
+  "/configure/*":{
```
holachuy: Can we just change the endpoint name to /_/configure where it is used on the client and where it is mapped in the server then we dont' need this block?

client/src/app/backend.service.ts
```diff
@@ -55,6 +58,14 @@ export class BackendService {
       .catch(err => this.handleError(err));
   }
 
+  configure(formData: DataTransferRequest) {
+    let url = '/configure';
```
holachuy: Can we use  /_/configure ?

client/src/app/backend.service.ts
```diff
@@ -126,6 +136,13 @@ export class BackendService {
     return body;
   }
 
+  private configureSuccess(res:DataTransfer){
+    // Redirect to the export authorization flow after configure.
+    // this should be returned from the configure request and is checked
+    // upon creation of the DataTransfer object.
+    window.location.href=res.nextURL;
```
holachuy: Actually, I just checked and Angular has a location object you can import and use that is a bit more angular-ly
location.go("the url");

https://angular.io/api/common/Location
https://alligator.io/angular/location-service/

portability-types-client/src/main/java/org/dataportabilityproject/types/client/transfer/DataTransfer.java
```diff
@@ -32,19 +32,25 @@
...
+    private String nextURL;
 
     @JsonCreator
     public DataTransfer(
             @JsonProperty(value = "source", required = true) String source,
             @JsonProperty(value = "destination", required = true) String destination,
             @JsonProperty(value = "transferDataType", required = true) String transferDataType,
-            @JsonProperty(value = "status", required = true) Status status) {
+            @JsonProperty(value = "status", required = true) Status status,
+            @JsonProperty(value = "nextURL", required = true) String nextURL) {
```
holachuy: Will this be a property only on certain types of requests or all? My guess was that this might not be required.

Can we add documentation as well.

portability-types-client/src/test/java/org/dataportabilityproject/types/client/transfer/DataTransferTest.java
```diff
@@ -13,11 +13,12 @@
     public void verifySerializeDeserialize() throws Exception {
         ObjectMapper objectMapper = new ObjectMapper();
 
-        DataTransfer transfer = new DataTransfer("testSource", "testDestination", "application/json", DataTransfer.Status.INPROCESS);
+        DataTransfer transfer = new DataTransfer("testSource", "testDestination", "application/json", DataTransfer.Status.INPROCESS,"nexturltofollow");
```
holachuy: space after INPROCESS,

client/src/app/backend.service.ts
seehamrun: So looking at this more, I think the location object is meant to control within the app. When I try location.go("https://external/oauth/url"); it prefixes it with the app domain (localhost/https://external/oauth/url). It also doesnt trigger a page reload to the new path (so having something valid like location.go("/next") only changes the address bar) 

The JS guide says to use the window.location instead for this: https://docs.angularjs.org/guide/$location

portability-types-client/src/main/java/org/dataportabilityproject/types/client/transfer/DataTransfer.java
seehamrun: Done - I added the comment on the class member since adding it at the constructor looked awkward.

portability-types-client/src/test/java/org/dataportabilityproject/types/client/transfer/DataTransferTest.java
seehamrun: Done - reformatted the file with intellij autoformatter.

portability-types-client/src/main/java/org/dataportabilityproject/types/client/transfer/DataTransfer.java
jimmarino: I was thinking the DataTransfer entity would be (eventually) POSTed to a URL when the user wanted to transfer data, but before they have supplied credentials. For my own understanding, what does nextURL do?

seehamrun: My intention was that nextURL would tell the front end client which urls to redirect to after handling the response. 

There are three main points where the api server needs to inform the front end where to redirect to: 
- in ConfigureHandler (which is wrongly named, should be something like CreateJobHandler), after the front end client creates the job, it should redirect to oauth flow for the source service. 
- in ImportSetupHandler, the client should be redirected to the oauth flow for the destination service 
- in CopySetup, the client should redirect to the CopyHandler which will start the Copy.

I imagine if this is meant to be a sort of status page, it could be used to redirect the user to the correct page to go to. Example: if the user hasn't supplied creds yet, we could use it to redirect to the oauth flows. Or if theyre checking on the status of the copy, we could redirect them back to the in-status page (or a successful page if the copy is complete)? 

Does that make sense?

PR #84, opened 2018-01-24
Update Overview.md

olsona:
Trying to add bits and pieces to documentation as I learn about them.


olsona:
I agree.  Perhaps when we're done moving over to the new structure we could use that as an opportunity to refresh the documentation.

olsona:
Squashing and merging, since the Travis CI testing seems to be flaky.

Issue #85, opened 2018-01-24
Travis Angular test is flaky


rtannenbaum:
ng test --single-run --no-progress --browser=ChromeNoSandbox --environment local
24 01 2018 22:52:08.810:INFO [karma]: Karma v1.4.1 server started at http://0.0.0.0:9876/
24 01 2018 22:52:08.813:INFO [launcher]: Launching browser ChromeNoSandbox with unlimited concurrency
24 01 2018 22:52:08.822:INFO [launcher]: Starting browser Chrome
24 01 2018 22:52:51.057:INFO [Chromium 62.0.3202 (Ubuntu 0.0.0)]: Connected on socket onhuHgqvEOP4J9oqAAAA with id 56551463
24 01 2018 22:53:01.059:WARN [Chromium 62.0.3202 (Ubuntu 0.0.0)]: Disconnected (1 times), because no message in 10000 ms.
Chromium 62.0.3202 (Ubuntu 0.0.0) ERROR
  Disconnected, because no message in 10000 ms.

rtannenbaum:
PR above fixes this, closing

PR #86, opened 2018-01-25
Increase timeout for Karma to receive activity from Chrome

rtannenbaum:
Do not pull yet; testing on Travis with a lower timeout first to make sure it is the correct flag


rtannenbaum:
That test failed as expected so we have the correct flag. Increasing timeout now.

rtannenbaum:
Fixes #85

PR #89, opened 2018-01-25
Use new DataTypes from portability-client-types in SetupHandler. 

seehamrun:
For using the new data types module (#71)
This requires a new binary and static content build (#67)

### DIFF
--- a/portability-api/src/main/java/org/dataportabilityproject/webapp/SetupHandler.java
+++ b/portability-api/src/main/java/org/dataportabilityproject/webapp/SetupHandler.java
@@ -29,9 +29,6 @@ import com.sun.net.httpserver.HttpHandler;
 import java.io.IOException;
 import java.net.HttpCookie;
 import java.nio.charset.StandardCharsets;
-import javax.json.Json;
-import javax.json.JsonObject;
-import javax.json.JsonWriter;
 import org.dataportabilityproject.ServiceProviderRegistry;
 import org.dataportabilityproject.job.JobDao;
 import org.dataportabilityproject.job.JobUtils;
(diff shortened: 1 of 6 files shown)


portability-api/src/main/java/org/dataportabilityproject/webapp/SetupHandler.java
```diff
@@ -180,9 +183,7 @@ JsonObject handleCopySetup(Headers requestHeaders, PortabilityJob job) {
 
     }
 
-    return Json.createObjectBuilder().add(JsonKeys.DATA_TYPE, job.dataType())
-        .add(JsonKeys.EXPORT_SERVICE, job.exportService())
-        .add(JsonKeys.IMPORT_SERVICE, job.importService()).build();
+    return new DataTransferResponse(job.exportService(), job.importService(), job.dataType(), Status.INPROCESS, "/startCopy");
```
holachuy: Either now or as a TODO, let's make /startCopy a constant in a shared class so we can keep all our urls in a central place

portability-api/src/main/java/org/dataportabilityproject/webapp/SetupHandler.java
```diff
@@ -56,6 +59,7 @@
   private final Mode mode;
   private final String handlerUrlPath;
   private final TokenManager tokenManager;
+  private final ObjectMapper objectMapper;
```
holachuy: Can this be static final?

seehamrun: Done (also made logger final static)

portability-api/src/main/java/org/dataportabilityproject/webapp/SetupHandler.java
seehamrun: Ack - I'll do that in my next PR

PR #90, opened 2018-01-25
Update LocalCopierModule.java

olsona:
This way, we don't bind GoogleCloudModule twice.


olsona:
@holachuy Looks like it works fine anyway.

Issue #88, opened 2018-01-25
Deploy worker on Kubernetes


rtannenbaum:

rtannenbaum:
Initial setup for worker pool should be pretty similar to API pool

Workers are configured to shut themselves down after processing a job, and our Kubernetes pod restart policy is configured to Always restart, so we should be good for refilling the initial pool of workers as they process jobs

For scaling, I think we should try Kubernetes autoscaler, which has a beta feature to scale based on a custom metric. We could make that use whether the worker is processing a job (i.e. it's at capacity) or not. https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough/

rtannenbaum:
This is working now with a fixed pool size. More advanced autoscaling work can be tracked in a separate issue.

PR #91, opened 2018-01-26
Kubernetes worker pool config

rtannenbaum:
Workers are up on our QA instance and are polling, but encryption flow in API is not enabled.

Also switch logging level for QA to DEBUG so we can see the worker polling statements in QA.

for #88


config/k8s/worker-deployment.yaml
```diff
@@ -0,0 +1,46 @@
...
+        env:
+        - name: GOOGLE_APPLICATION_CREDENTIALS
+          value: /var/secrets/portability-service-account-key/key.json
+        - name: GOOGLE_PROJECT_ID
+          value: world-takeout-qa
+        resources: {}
+        terminationMessagePath: /dev/termination-log
+        terminationMessagePolicy: File
+      dnsPolicy: ClusterFirst
+      restartPolicy: Always
```
holachuy: Is this what makes it restart when it fails?

config/k8s/worker-deployment.yaml
```diff
@@ -0,0 +1,46 @@
+apiVersion: extensions/v1beta1
```
holachuy: Let's add the license header and also an description including that we count on these to restart themeselves.

config/gcp/setup_gke_environment.sh
```diff
@@ -54,6 +55,111 @@ print_step() {
...
+
+  print_step "Importing the service account credentials, created earlier, as a Kubernetes Secret"
+  kubectl create secret generic portability-service-account-creds --from-file=key.json=/tmp/service_account_creds.json
+
+  if [[ ${BACKEND} != "api" ]]; then
+    echo -e "Done creating ${BACKEND} pool!"
+    return
+  fi
+
+  # -----------------------------------------------------------------------
```
holachuy: Should this be a separate method?

rtannenbaum:
I tried rebasing but I think it's too late since I had done some merges. I'll stick with rebasing from now on.

holachuy:
Yeah, I still don't know what works best as far as a git workflow but it usually leaves the history a bit cleaner. Maybe you can squash and merge in as one commit to avoid to long commit history

PR #94, opened 2018-01-26
update qa index.html to reflect new content from #80 and #89, explicitly turn off encryptedFlow in qa

seehamrun:
This pushes frontend and binary changes from #80 and #89 to qa  
#67


seehamrun:
Turned off encryptedFlow in QA until we debug issues. Frontend deployment w/o encyrpted flow looks good.

PR #97, opened 2018-01-26
update angular frontend and api server to use types in portability-client-types

seehamrun:
This is part of #71 
This requires a new static content deploy and binary push (#67)

Let me know if this is on the right track. I created a couple of new "Response" type classes but am not sure if this is the model we want to use.


portability-api/src/main/java/org/dataportabilityproject/webapp/ListDataTypesHandler.java
```diff
@@ -56,22 +57,22 @@ public void handle(HttpExchange exchange) throws IOException {
     Headers headers = exchange.getResponseHeaders();
     headers.set(HEADER_CONTENT_TYPE, "application/json; charset=" + StandardCharsets.UTF_8.name());
 
-    JsonArrayBuilder builder = Json.createArrayBuilder();
+    Vector<String> data_types = new Vector<>();
```
holachuy: ArrayList

seehamrun: Done. Forgot that this wasnt C++ haha

Issue #72, opened 2018-01-22
Modularize ServiceProviders with new gradle structure


seehamrun:
blocks #21 (switch from maven to gradle)

bwillard:
I am going to take a crack at moving the Google one over to Gradle.

PR #102, opened 2018-01-29
Add a copy of the transfer data types to the new project structure

bwillard:
Makes progress on #72


portability-types-transfer/src/main/java/org/dataportabilityproject/types/transfer/models/calendar/CalendarModelWrapper.java
```diff
@@ -0,0 +1,47 @@
...
+import com.fasterxml.jackson.annotation.JsonTypeName;
+import com.google.common.collect.ImmutableList;
+import java.util.Collection;
+import org.dataportabilityproject.types.transfer.models.DataModel;
+
+/**
+ * A Wrapper for all the possible objects that can be returned by a calendar exporter.
+ */
+@JsonTypeName("CalendarModelWrapper")
+public class CalendarModelWrapper extends DataModel {
```
holachuy: The name wrapper throws me off occasionally. Maybe we can take advantage of this refactor to rename the top level entities that encapsulate all the stuff for a type of data.

CalendarDataToTransfer
CalendarDataModel (since it extends DataModel)
CalendarTransferData 
CalendarTransferModel 
CalendarData

PR #107, opened 2018-01-30
Change SimpleSubmitHandler and angular page to use new DataTypes from portability-types-client

seehamrun:
This requires both a binary push and a static content deploy (#67)

This is the last piece to switching the portability-api to use the new data types (#71)


client/src/app/simplelogin/simplelogin.component.ts
```diff
@@ -15,37 +15,43 @@
...
 })
+
 export class SimpleLoginComponent implements OnInit {
   url = environment.apiPostUrl;
   username: string = "";
   password: string = "";
   enableSubmit: boolean = false;
   error_text: string = "";
-  constructor() { }
+  constructor(private service : BackendService) { }
```
holachuy: Run ng test and e2e tests to make sure we don't have to update anything to make them pass. Travis should pick up the failure, if there is one, too.

portability-types-client/src/main/java/org/dataportabilityproject/types/client/transfer/SimpleLoginRequest.java
```diff
@@ -0,0 +1,28 @@
...
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+
+@ApiModel
+public class SimpleLoginRequest {
+  private final String username;
+  private final String password;
```
holachuy: TODO: Determine how to avoid storing the password in the clear.

client/src/app/simplelogin/simplelogin.component.ts
seehamrun: Done.

portability-types-client/src/main/java/org/dataportabilityproject/types/client/transfer/SimpleLoginRequest.java
seehamrun: Done. Note that this is also only used in the microsoft mail auth flow. We can delete it once we move that over to oauth or Live Sync.

PR #108, opened 2018-01-30
Re add in ExportInformation

bwillard:
Using ContinuationData as the input to the export call doesn't work because you then don't know which resource to apply the continuation data to.

So a call to export need to take 0 or 1 resources and 0 or 1 continuation tokens, which is now represented in ExportInformation


jimmarino:
+1 Looks good to me

jimmarino:
OopsI think I accidentally closed it...

PR #109, opened 2018-02-01
Key jobs in key-value store just by job ID, instead of by 'jobId::jobState'

rtannenbaum:
Key jobs in key-value store just by job ID, instead of by jobId::jobState.

Now that db updates are atomic, the job should only ever exist in one state at a given time, so this should be safe.

This will speed up our indexing for the future.

Also:
- Remove JobDao and have callers use storage directly, to make the transition to JobStore easier, as JobStore uses jobId and PortabilityJob in its interface (in Jim's latest PR - https://github.com/google/data-portability/pull/104/files).
- Simplify how we check that the job was present in the expected state (do this in PersistentKeyValueStore.atomicUpdate)
- Simplify how we check that a new entry doesn’t already exist (do this in the PersistentKeyValueStore.put)

Tested locally with the encryption flow both on & off.


portability-core/src/main/java/org/dataportabilityproject/cloud/google/GooglePersistentKeyValueStore.java
```diff
@@ -55,67 +56,89 @@ public GooglePersistentKeyValueStore(Datastore datastore) {
     this.datastore = datastore;
   }
 
+  /**
+   * Puts {@code jobId} in the database and verifies it doesn't already exist. To update the value
```
holachuy: nit, maybe something like:
 Inserts a new {@link PortabilityJob} keyed by {@code jobId}. Throws SOMEEXCPETION if job already exists. To update an existing {@link PortabilitJob}, use {@link #atomicUpdate(...)} method

portability-core/src/main/java/org/dataportabilityproject/cloud/google/GooglePersistentKeyValueStore.java
```diff
@@ -55,67 +56,89 @@ public GooglePersistentKeyValueStore(Datastore datastore) {
...
-  public String getFirst(String prefix) {
-    Query<Key> query = Query.newKeyQueryBuilder().setKind(KIND).build();
-    QueryResults<Key> results = datastore.run(query);
-    while(results.hasNext()) {
-      String key = results.next().getName();
-      if(key.startsWith(prefix)) {
-        return key;
-      }
+  public String getFirst(JobState jobState) {
+    Query<Entity> query = Query.newEntityQueryBuilder()
```
holachuy: Could we make this a static helper method for easier testing?

portability-core/src/main/java/org/dataportabilityproject/cloud/google/GooglePersistentKeyValueStore.java
```diff
@@ -185,7 +208,11 @@ private Entity createEntity(String key, Map<String, Object> data) throws IOExcep
...
   }
 
-  private Key getKey(String key) {
-    return datastore.newKeyFactory().setKind(KIND).newKey(key);
+  private Key getKey(String jobId) {
+    return datastore.newKeyFactory().setKind(KIND).newKey(jobId);
+  }
+
+  private JobState getJobState(Entity entity) {
+    return JobState.valueOf(entity.getString(PortabilityJobConverter.JOB_STATE));
```
holachuy: Should we check that 'entity.getString(PortabilityJobConverter.JOB_STATE))' is not null?

portability-core/src/main/java/org/dataportabilityproject/cloud/local/InMemoryPersistentKeyValueStore.java
```diff
@@ -37,55 +40,65 @@ public InMemoryPersistentKeyValueStore() {
...
+          "Couldn't update job " + jobId + " from previous state " + previousState, e);
     }
-
-    put(newKeyStr, data);
-    delete(previousKeyStr);
-    return true;
   }
 
+  private JobState getJobState(Map<String, Object> data) {
+    return JobState.valueOf(data.get(PortabilityJobConverter.JOB_STATE).toString());
```
holachuy: Check for null?

portability-core/src/main/java/org/dataportabilityproject/job/JobDao.java
```diff
@@ -101,13 +93,13 @@ public PortabilityJob lookupAssignedWithAuthDataJob(String id) {
...
+   */
   private PortabilityJob lookupJob(String id, JobState jobState) {
     Preconditions.checkNotNull(id);
-    Map<String, Object> data = storage.get(createKey(jobState, id));
-    if (data == null || data.isEmpty()) {
-      return null;
-    }
-    PortabilityJob job = PortabilityJob.mapToJob(data);
+    PortabilityJob job = storage.get(id);
+    Preconditions.checkArgument(job == null || job.jobState() == jobState);
```
holachuy: &&

portability-core/src/main/java/org/dataportabilityproject/job/JobDao.java
```diff
@@ -193,96 +177,44 @@ public void updateJobStateToAssigneWithAuthData(String id, String encryptedExpor
...
-    if (updated) {
-      logger.debug("Successfully updated state for job {}. Previous state: {}. New state: {}",
-          job.id(), previousJobState, newJobState);
-      return true;
-    } else {
-      logger.warn("Failed to update state for job {}. Previous state: {}. New state: {}",
-          job.id(), previousJobState, newJobState);
-      return false;
-    }
+    storage.atomicUpdate(job.id(), previousJobState, job);
```
holachuy: Is data not used? If this becomes a one line, we may just want to inline it above where used

portability-core/src/main/java/org/dataportabilityproject/job/JobDao.java
```diff
@@ -70,11 +66,7 @@ public JobDao(PersistentKeyValueStore persistentKeyValueStore) {
 
   /** Returns the next job in unassigned state. */
   public String findNextJobPendingWorkerAssignment() {
-    String key = storage.getFirst(JobState.PENDING_WORKER_ASSIGNMENT.name());
-    if (key != null) {
-      return getIdFromKey(key);
-    }
-    return null;
+    return storage.getFirst(JobState.PENDING_WORKER_ASSIGNMENT);
```
holachuy: Now that it's one line, can we just inline where it is called and remove it from this object? How would that look?

Or do we want a WorkerJobDao which holds this and API specific methods in another one?

portability-core/src/main/java/org/dataportabilityproject/job/JobDao.java
```diff
@@ -193,96 +177,44 @@ public void updateJobStateToAssigneWithAuthData(String id, String encryptedExpor
...
-   */
-  private void updateJob(PortabilityJob job, JobState jobState) throws IOException {
-    String key = createKey(jobState, job.id());
-    Map<String, Object> existing = storage.get(key);
-    Preconditions.checkArgument(existing != null, "Job not found");
-    // Store the updated job info
-    Map<String, Object> data = job.asMap();
-    logger.debug("Data: {}", data);
-    storage.put(key, data);
+  public void deleteJob(String jobId) throws IOException {
```
holachuy: Now that it's one line, maybe we can move this to WorkerJobDao or just inline it where needed

portability-core/src/main/java/org/dataportabilityproject/cloud/google/GooglePersistentKeyValueStore.java
```diff
@@ -212,7 +249,13 @@ private Key getKey(String jobId) {
...
   }
 
+  /**
+   * Return {@code entity}'s {@link JobState}, or null if missing.
+   *
+   * @param entity a {@link PortabilityJob}'s representation in {@link #datastore}.
+   */
   private JobState getJobState(Entity entity) {
-    return JobState.valueOf(entity.getString(PortabilityJobConverter.JOB_STATE));
+    String jobState = entity.getString(PortabilityJobConverter.JOB_STATE);
```
holachuy: Is it valid for an entity ever to have null? Is this only for the legacy, non encrypted flow case? If so, let's  add a TODO to remove.

portability-core/src/main/java/org/dataportabilityproject/cloud/local/InMemoryPersistentKeyValueStore.java
```diff
@@ -98,7 +133,13 @@ public void atomicUpdate(String jobId, JobState previousState, PortabilityJob jo
...
   }
 
+  /**
+   * Return {@code data}'s {@link JobState}, or null if missing.
+   *
+   * @param data a {@link PortabilityJob}'s representation in {@link #map}.
+   */
   private JobState getJobState(Map<String, Object> data) {
-    return JobState.valueOf(data.get(PortabilityJobConverter.JOB_STATE).toString());
+    Object jobState = data.get(PortabilityJobConverter.JOB_STATE);
```
holachuy: Same here

portability-core/src/main/java/org/dataportabilityproject/cloud/google/GooglePersistentKeyValueStore.java
rtannenbaum: my understanding is it's only for the legacy non encrypted case. I'll add a TODO.

portability-core/src/main/java/org/dataportabilityproject/cloud/local/InMemoryPersistentKeyValueStore.java
rtannenbaum: done

rtannenbaum:
Since you approved pending the requested changes, which I just made, I'm going to try clicking on "Dismiss review" so it lets me merge...

rtannenbaum:
Ha, just kidding, was curious what that would let me do but I still need explicit approval to merge. I guess that's a good thing!

PR #113, opened 2018-02-02
Migrate PortabilityJob, AuthData, and job stores to the new directory structure

rtannenbaum:
Make portability-core call PortabilityJob and AuthData in the new directory structure, and deprecate the old versions.
    
Also replace PersistentKeyValueStore with JobStore, and replace the following impls:
- GooglePersistentKeyValueStore -> GoogleCloudDatastore implements JobStore
- InMemoryPersistentKeyValueStore -> InMemoryKeyValueStore implements JobStore

Tested locally with encryption flow on and off.


rtannenbaum:
I'd prefer to move to the new PortabilityJob in a separate PR, this one already has a lot of changes and I haven't thought too much about how to port over yet.

I just updated the name to LegacyPortabilityJob, good idea

Also, FYI this is branched off of the other branch with the JobStore changes, so after this I'll send another PR to merge the whole thing into master

PR #118, opened 2018-02-06
Modular launching example interfaces

jimmarino:
I don't want to merge at this point but but instead wanted to create a branch to discuss the second half of modularity: how extensions get assembled and launched in processes (gateway, transfer).

The first thing the straw approach does is create a portability-api-launcher module. This module contains interfaces the process uses to launch and boot an extension. There may be a better name than "launcher" but I used "api" since it consists of interfaces used by client as opposed to a pure extension point (SPI). 

The portability-api-launcher module defines the base extension lifecycle in AbstractExtension. Each extension type (Gateway, Cloud, Transfer) has a subtype defined in its SPI module, e.g. CloudExtension. These subtypes are used to load the extensions as well as provide a point of reference for implementors to know exactly what they need to provide in an extension.

My thought is the AbstractExtensions would be loaded by the main process method. I created an example in WorkerMain. This example uses Java's ServiceLoader:

 ServiceLoader.load(TransferExtension.class)

In pre-Java 9 ServiceLoader will attempt to load implementations by scanning for META-INF/services files on the classpath. This provides some flexibility around packaging since an uber jar can have a single service file with multiple implementations specified or we can create a packaging where each extension is in a separate jar that includes META-INF/services (in which case Java will effectively concatenate them). If we eventually decide to use Java 9, META-INF/services will not be needed since module-info.java defines the provider implementations.

An important benefit of this approach is the process configuration no longer needs to define which extensions to load in its configuration file.

Another benefit is this approach will allow alternative launch configurations. For example, instead of a process main, a runtime could be launched inside a JUnit test runner, build plugin or from a command line utility.

The AbstractExtension defines this method:

 /**
     * Initializes the extension. Implementations prepare provided services.
     *
     * @param context the extension context.
     */
    void initialize(ExtensionContext context);

ExtensionContext provides primordial services an extension may require. For example, configuration values can be looked up by key:

 <T> T getConfiguration(String key, String defaultValue);

The idea here is to have the launching process parse YAML or provide another configuration source to extensions.

There is also a method for getting a logger that can be passed to extension services:

 Logger getLogger();

The logger is intended to be a wrapper for a delegate. The underlying implementation is provided by the launcher. Again this provides flexibility to enable using logging services such as those from cloud providers, console output for command line tools, or even plugin loggers (e.g. Gradle's logger).

Extensions may also require some service from the runtime, for example the ObjectMapper to register types. This is done using:

 <T> T getService(Class<T> type);

Each AbstractExtension type defines the services the extension must provide the runtime. For example, GatewayExtension has: 

List<AuthServiceProvider> getAuthServiceProviders();

After initializing the extension, the main process would call these methods to obtain the extension instances and register them with core services or the main Guice Injector.

This approach may require a few things:

1. Cloud extensions may need to be loaded before transfer extensions since the latter may rely on the JobStore. This should be easy to achieve using ServiceLoader.

2. Some core services may be "primordial" (e.g. required by extensions, e.g. the ObjectMapper) while others may need to be loaded after extension classes. I don't see any specific cases at the moment but a launcher may need to handle this in the future. This can be done using hierarchical Guice injectors. 

Currently it looks as if the core runtime services can be booted followed by the extensions. After extensions are loaded, their instances can be registered with the appropriate core registries.


portability-worker/src/main/java/org/dataportabilityproject/worker/WorkerMain.java
```diff
@@ -18,17 +18,66 @@
...
-    WorkerImpl worker = injector.getInstance(WorkerImpl.class);
-    worker.processJob();
-    System.exit(0);
-  }
+    public static void main(String[] args) throws Exception {
+        Thread.setDefaultUncaughtExceptionHandler(UncaughtExceptionHandlers.systemExit());
+
+        ExtensionContext extensionContext = getExtensionContext();
+
+        List<TransferExtension> extensions = new ArrayList<>();
```
holachuy: Just a thought, we might want to limit it to two transfer extensions, one for import and one for export, for now and eventually one when we head down to separate importer and export processes, just one.

I think in this case, it helps to be better defined and less general.

portability-worker/src/main/java/org/dataportabilityproject/worker/WorkerMain.java
```diff
@@ -18,17 +18,66 @@
...
+        }
+
+        System.exit(0);
+    }
+
+    private static ExtensionContext getExtensionContext() {
+        return new ExtensionContext() {
+                @Override
+                public Logger getLogger() {
+                    return new Logger() {{
```
holachuy: Does this mean there is a logger per extension?

portability-api-launcher/src/main/java/org/dataportabilityproject/api/launcher/ExtensionContext.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+public interface ExtensionContext {
+
+    /**
+     * Returns the system logger extension service can use to record events.
+     */
+    Logger getLogger();
+
+    /**
+     * Returns a system service such as a type mapper extension services may require.
```
holachuy: Just for clarity, what would be a specific example?

jimmarino: Nothing requires Java 9 so it should be good.

Do you mean an example of the logger or ExtensionContext or both? Let me know and I'll create a Gist.

portability-worker/src/main/java/org/dataportabilityproject/worker/WorkerMain.java
jimmarino: We could do that. One thing I was thinking is the embedded case where multiple extensions are loaded in memory. We could launch multiple workers for that case.

portability-worker/src/main/java/org/dataportabilityproject/worker/WorkerMain.java
jimmarino: Ah that's a good point. Maybe we want to change that to getLogger(Object key)?

Then, the launcher implementation is free to decide what to return.

portability-spi-gateway/src/main/java/org/dataportabilityproject/spi/gateway/extension/GatewayExtension.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Gateway extensions implement this interface to be loaded in a gateway process.
+ */
+public interface GatewayExtension extends AbstractExtension {
+
+    /**
+     * Returns service provider information supported by this extension.
+     */
+    List<ServiceProviderInfo> getServiceProviderInfos();
```
seehamrun: So the result of this would then later be used to initialize the guice injections? 
I think this aligns well with the work I was starting for implementing the Auth and Transfer ServiceProviderRegistry classes. I have already started implementation and was trying to figure out how the registry would obtain the map of providers (similar to what we do [here](https://github.com/google/data-portability/blob/377b38e9d3997ca42356d9615b7cb4147e614644/portability-core/src/main/java/org/dataportabilityproject/ServiceProviderRegistry.java#L52))

portability-worker/src/main/java/org/dataportabilityproject/worker/WorkerMain.java
```diff
@@ -18,17 +18,66 @@
...
-    worker.processJob();
-    System.exit(0);
-  }
+    public static void main(String[] args) throws Exception {
+        Thread.setDefaultUncaughtExceptionHandler(UncaughtExceptionHandlers.systemExit());
+
+        ExtensionContext extensionContext = getExtensionContext();
+
+        List<TransferExtension> extensions = new ArrayList<>();
+        ServiceLoader.load(TransferExtension.class).iterator().forEachRemaining(extensions::add);
```
seehamrun: This is probably due to my lack of java knowledge. You mentioned that this ServiceLoader works pre-java9 by using META-INF/services files, are these automatically generated or something that we would need to check into git? 

I think @holachuy 's question above might also clear this one up: for V1, this would mean that we specify all the valid services in this file and just load them all in our jar. How would this look for a V2, where we selectively load the service classes?

portability-spi-gateway/src/main/java/org/dataportabilityproject/spi/gateway/extension/GatewayExtension.java
jimmarino: Yes I think the injectors could be initialized with those or they could be manually registered with the instances produced by the injectors. We may need to use a hierarchy of injectors if there are timing issues, e.g. extensions require a set of services to initialize. In this case we could use an hierarchy of Injector instances to handle this.

portability-worker/src/main/java/org/dataportabilityproject/worker/WorkerMain.java
jimmarino: Yes, that's correct they work pre-Java 9. Here's an example:

http://www.logicbig.com/tutorials/core-java-tutorial/java-se-api/service-loader/

The idea is each jar/module extension would create a META-INF/services file. When the jar is on the classpath, ServiceLoader will find it's implementation. For uber/shaded jars, the Gradle plugin will concatenate the META-INF/services file so that multiple implementations are included in it.

Java 9 adds the ability to use module-info declarations instead of META-INF services, but we don't need to use this for now. For example:

module com.foo.myprovider {

    provides org.dataportabilityproject.transfer.spi.TransferExtension with com.foo.MyExtensionImpl;

} 

And a module-info that uses the extensions:

module com.foo.myworkerprocess {
    uses org.dataportabilityproject.transfer.spi.TransferExtension;
}

For simplicity, I think each extension can just keep META-INF/services files in its /resource directory. 

For now, the pre-Java9 approach should be OK but in the future, I think we are going to want a way to "assemble" images that's easier to use. For example, someone may want to create a worker process with there own set of extensions rolled up into a Docker image. There are number of different approaches we can probably take here once we have things working with the new APIs/SPIs.

jimmarino:
Here's a good overview of JPMS if anyone is interested:

https://www.youtube.com/watch?v=gtcTftvj0d0

I think we can use pre-Java 9 for now but in the future we may want to consider this as it will solve a lot of problems with library versioning.

portability-api-launcher/src/main/java/org/dataportabilityproject/api/launcher/ExtensionContext.java
holachuy: Oh, I mean what would be an example of a system service?

portability-api-launcher/src/main/java/org/dataportabilityproject/api/launcher/AbstractExtension.java
```diff
@@ -0,0 +1,41 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.dataportabilityproject.api.launcher;
+
+/**
+ * Implementations provide an extension to the system.
+ */
+public interface AbstractExtension {
```
holachuy: I'm wondering if the different types of extensions could be confusing. I don't think it's worth changing now but we may want to rename this in the future.

portability-worker/src/main/java/org/dataportabilityproject/worker/WorkerMain.java
holachuy: Yes, that sounds good.

portability-worker/src/main/java/org/dataportabilityproject/worker/WorkerMain.java
rtannenbaum: yeah, this will be interesting come java 9. I can imagine a situation where we have many permutations of different modular jars for the various import/export service combinations, and getting our worker pool (be it Kubernetes or something else) to spin up a worker with the correct image with the correct jar for each job could be complicated. But maybe that is what you have to do in order to get full "modularity" here where we are not pulling in code from unused libraries.

portability-api-launcher/src/main/java/org/dataportabilityproject/api/launcher/Logger.java
```diff
@@ -0,0 +1,52 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.dataportabilityproject.api.launcher;
+
+import java.util.function.Supplier;
+
+/**
+ * The system logger used to record events. Implementations may delegate to a logging implementation.
```
rtannenbaum: Is this so that people don't have to use slf4j if they have a different implementation? If so, I feel like this may add more complexity than it's worth, I think it's OK to pick some impls of things like logging where there is a common, accepted library out there..

Plus different impls may not cleanly fit with this interface if the logging levels don't 100% map 1:1

Let me know if I'm misunderstanding what this is for

portability-worker/src/main/java/org/dataportabilityproject/worker/WorkerMain.java
jimmarino: For this case, I think we could keep things simple to start: all workers run all extensions and hence can support all combinations. 

In the future, if we want to support different scaling characteristics one way to do this would be to use Java 9 and create module layers that loaded modules dynamically based on environment variables. Kubernetes or some other orchestrator could launch an image and set the environment variables based on how it wants to balance processes. 

There's not much written on module layers (outside of the JPMS spec) yet but they are pretty powerful since they allow Java code to dynamically load (and unload) modules at runtime rather than having to specify the module path at startup. They can also be used to support versioning and isolation since a layer can hold one or more modules and be associated with a classloader.

portability-api-launcher/src/main/java/org/dataportabilityproject/api/launcher/AbstractExtension.java
jimmarino: +1 on closeable

portability-api-launcher/src/main/java/org/dataportabilityproject/api/launcher/ExtensionContext.java
jimmarino: A system service could be a registry or other service that extensions must register against. For example, an extension may need to register some type of provider.

portability-api-launcher/src/main/java/org/dataportabilityproject/api/launcher/Logger.java
jimmarino: What I was thinking here is an easy way for people to plug in their own log capture. For example, someone may want to have all logs sent directly to a cloud logging service. If we provide a direct mechanism where they pass in an instance, its easy for them to configure a logging service at startup before passing the instance to the core runtime. 

If we standardize on  SLF4J, someone will either have to pipe the output from an SLF4J implementation to the log service; create and SLF4J binder, or write a handler and set it via some configuration and rely on statics for configuration values.

rtannenbaum: SGTM

jimmarino:
Should we merge this in, i.e. is everyone OK with the pull request? Assuming so, does someone want to do it or should I?

holachuy:
Hi Jim,

So far, I guess as we're all core contributors, we've been merging our own changes once they have approval and requested changes are addressed so feel free to merge in.

Of course, we can also change the flow if you there's a better suggestion. 

Thanks! Great stuff!

rtannenbaum:
I'm pretty confident the Travis failure is either a flake or unrelated to this PR since it's the NodeJS run that's failing. Will try and merge into master if GitHub will let me.

PR #122, opened 2018-02-08
Start of implementation of PortabilityTransferServiceProvider

seehamrun:
This is not yet ready for merging as it depends on modularity changes and launchers to be tested.


portability-transfer/src/main/java/org/dataportabilityproject/transfer/PortabilityTransferServiceProviderRegistry.java
```diff
@@ -0,0 +1,89 @@
...
+   * @param transferDataType the transfer data type
+   */
+  @Override
+  public Importer<AuthData, DataModel> getImporter(String serviceId,
+      String transferDataType) {
+    Preconditions.checkArgument(supportedImportTypes.contains(transferDataType),
+        "TransferDataType [%s] is not valid for import", transferDataType);
+    TransferServiceProvider serviceProvider = serviceProviderMap.get(serviceId);
+    Preconditions.checkArgument(serviceProvider != null);
+    return (Importer<AuthData, DataModel>) serviceProvider.getImporter(transferDataType);
```
seehamrun: I wasnt sure what the best way to cast the Importer<?, ?> object thats returned from the different service providers. Is this something thats commonly done?

holachuy: We lose the subtype info by using the base classes in the cast and method signature, e.g. AuthData and DataModel.  Can we just return wildcards as well?

seehamrun: Done - I had to make some changes to the PortabilityCopier to reflect the new changes and handle the wildcard types. I think I did them correctly after a couple of different casting tries.

portability-transfer/src/test/org/dataportabilityproject/transfer/PortabilityTransferServiceProviderRegistryTest.java
```diff
@@ -0,0 +1,27 @@
...
+
+package org.dataportabilityproject.transfer;
+
+import org.junit.Test;
+
+public class PortabilityTransferServiceProviderRegistryTest {
+
+  @Test
+  public void getImporterTest(){
+
```
holachuy: TODO

seehamrun: Added test

portability-transfer/src/test/java/org/dataportabilityproject/transfer/PortabilityTransferServiceProviderRegistryTest.java
```diff
@@ -0,0 +1,68 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+public class PortabilityTransferServiceProviderRegistryTest {
+  @Rule
+  public ExpectedException thrown = ExpectedException.none();
+
+  @Test
+  public void supportImportAndExportTest(){
```
holachuy: Maybe it would be clearer if it was called: requireImportAndExportTest

portability-transfer/src/test/java/org/dataportabilityproject/transfer/PortabilityTransferServiceProviderRegistryTest.java
```diff
@@ -0,0 +1,68 @@
...
+        ImmutableMap.of("mockServiceProvider", mockTransferProvider));
+  }
+
+  @Test
+  public void serviceProviderNotFoundTest(){
+    List<String> supportedImportTypes = ImmutableList.of("photos", "contacts");
+    List<String> supportedExportTypes = ImmutableList.of("contacts");
+
+    TransferServiceProvider mockTransferProvider = mock(TransferServiceProvider.class);
+    when(mockTransferProvider.getExportTypes()).thenReturn(supportedExportTypes);
```
holachuy: Do you need these two lines for mocking getExportTypes() and getImportTypes()  if they're not called?

portability-transfer/src/test/java/org/dataportabilityproject/transfer/PortabilityTransferServiceProviderRegistryTest.java
seehamrun: Done.

portability-transfer/src/test/java/org/dataportabilityproject/transfer/PortabilityTransferServiceProviderRegistryTest.java
seehamrun: Oh right - those were leftover from before. 
Done.

PR #128, opened 2018-02-12
Google contacts

olsona:
First round of VCard -> Google Contacts conversion, some test cleaning


portability-core/src/main/java/org/dataportabilityproject/serviceProviders/google/contacts/VCardToGoogleContactConverter.java
```diff
@@ -0,0 +1,111 @@
...
+import com.google.common.base.Preconditions;
+import ezvcard.VCard;
+import ezvcard.property.Email;
+import ezvcard.property.StructuredName;
+import ezvcard.property.Telephone;
+import java.util.stream.Collectors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class VCardToGoogleContactConverter {
```
holachuy: Another idea, depending what you think is to expose this converter and the reverse by extending an actual Converter type and delegate to these specific implementations? 

https://google.github.io/guava/releases/17.0/api/docs/com/google/common/base/Converter.html

olsona: Let's talk with Jim about this idea.

PR #130, opened 2018-02-13
Update index.html in QA to reflect the latest changes

rtannenbaum:
This fixes the frontend issues in QA, and is already deployed to our API server (#67)

Been a while since we have deployed, but according to the last PR tagged in #67, we think this deploys changes from #107


rtannenbaum:
Thanks! I'll update the comment to reflect that

Issue #19, opened 2018-01-07
Modularize web server


bwillard:
Right now we are using the built in Java web server, this probably isn't the ideal.

We should make the web server implementation modular so that
a) we can replace it with a more scalable one easily
b) Other implementations can easily be swapped in.

PR #134, opened 2018-02-13
initial implementation of the api in the gateway package

holachuy:
implementation of the api in the gateway package of an initial set of handlers, actions, and supporting classes #19 

This has a barebones main class for creating a binary until the new launcher api comes into play.

Introduces a concept of actions which are independent of the protocol, e.g. http, and web framework.

Most of the handler logic is replicated without changes other than refactoring into actions.


rtannenbaum:
Whoops, forgot about JobAuthorization.State. I'll send you a separate PR.

portability-gateway/src/main/java/org/dataportabilityproject/gateway/action/createjob/CreateJobActionRequest.java
```diff
@@ -0,0 +1,29 @@
+/*
```
seehamrun: I know we're not using gRPC here, but is it possible to use protos for the request and response instead of using POJO's? It probably doesn't matter in the v1, but might get rid of a lot of boiler plate.

portability-gateway/src/main/java/org/dataportabilityproject/gateway/reference/DataTransferHandler.java
```diff
@@ -0,0 +1,153 @@
...
+  public static final String PATH = "/_/DataTransfer";
+  public static final String ERROR_PATH = "/error";
+  private final CreateJobAction createJobAction;
+  private final AuthServiceProviderRegistry registry;
+  private final ApiSettings apiSettings;
+  private final JobStore store;
+  private final ObjectMapper objectMapper;
+
+  @Inject
+  DataTransferHandler(CreateJobAction createJobAction, AuthServiceProviderRegistry registry,
```
seehamrun: the createJobAction is something that will be initialized via dep injection right? In which module would that be initially created?

portability-gateway/src/main/java/org/dataportabilityproject/gateway/reference/ReferenceApiModule.java
```diff
@@ -0,0 +1,93 @@
...
+  protected void configure() {
+    MapBinder<String, HttpHandler> mapbinder
+        = MapBinder.newMapBinder(binder(), String.class, HttpHandler.class);
+
+    // HttpServer does exact longest matching prefix for context matching. This means
+    // /_/listServices, /_/listServicesthisshouldnotwork and /_/listServices/path/to/resource will
+    // all be handled by the ListServicesHandler below. To prevent this, each handler below should
+    // validate the request URI that it is getting passed in.
+    mapbinder.addBinding(DataTransferHandler.PATH).to(DataTransferHandler.class);
+
```
seehamrun: Add TODO to hook up the below as they are implemented

portability-gateway/src/main/java/org/dataportabilityproject/gateway/reference/ReferenceApiModule.java
```diff
@@ -0,0 +1,93 @@
...
+      /*
+      mapbinder.addBinding(CopySetupHandler.PATH).to(CopySetupHandler.class);
+      mapbinder.addBinding(ImportSetupHandler.PATH).to(ImportSetupHandler.class);
+      mapbinder.addBinding(ListDataTypesHandler.PATH).to(ListDataTypesHandler.class);
+      mapbinder.addBinding(ListServicesHandler.PATH).to(ListServicesHandler.class);
```
seehamrun: Do you also want to enable the ListServicesHandler in this PR too?

portability-gateway/src/main/java/org/dataportabilityproject/gateway/ApiSettings.java
```diff
@@ -0,0 +1,62 @@
+/*
```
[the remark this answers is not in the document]
holachuy: Removed ApiFlags. Eventually this will move to the new launcher api based context.

portability-gateway/src/main/java/org/dataportabilityproject/gateway/action/Action.java
```diff
@@ -0,0 +1,25 @@
...
+package org.dataportabilityproject.gateway.action;
+
+/**
+ * Handles an action in the API server.
+ */
+public interface Action<Request, Response> {
+
+  /** Handles the given {@code Request} and returns a {@code Response}. */
+  Response handle(Request request);
+}
```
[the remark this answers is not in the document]
holachuy: Done

portability-gateway/src/main/java/org/dataportabilityproject/gateway/action/Action.java
```diff
@@ -0,0 +1,25 @@
+/*
```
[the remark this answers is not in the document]
holachuy: I was wondering about that. I just moved it to the SPI.

portability-gateway/src/main/java/org/dataportabilityproject/gateway/crypto/AsymmetricKeyGenerator.java
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Creates and encodes keys for asymmetric encryption.
+ */
+public interface AsymmetricKeyGenerator {
+
+  /**
+   * Generates a {@link PublicKey}
+   */
+  KeyPair generate();
+}
```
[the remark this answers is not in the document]
holachuy: Done

portability-gateway/src/main/java/org/dataportabilityproject/gateway/crypto/Encrypter.java
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Encrypts data for the given key type.
+ */
+public interface Encrypter {
+
+  /**
+   * Encrypts the given {@code data}.
+   */
+  String encrypt(String data) ;
+}
```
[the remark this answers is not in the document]
holachuy: Done

portability-gateway/src/main/java/org/dataportabilityproject/gateway/crypto/RsaSymmetricKeyGenerator.java
```diff
@@ -0,0 +1,47 @@
...
+    try {
+      kpg = KeyPairGenerator.getInstance(ALGORITHM);
+    } catch (NoSuchAlgorithmException e) {
+      logger.error("NoSuchAlgorithmException for: {}", ALGORITHM, e);
+      throw new RuntimeException("NoSuchAlgorithmException generating key", e);
+    }
+    kpg.initialize(1024);
+    return kpg.genKeyPair();
+  }
+}
```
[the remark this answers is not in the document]
holachuy: Done

portability-gateway/src/main/java/org/dataportabilityproject/gateway/crypto/SymmetricKeyGenerator.java
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Creates and encodes keys for symmetric encryption.
+ */
+public interface SymmetricKeyGenerator {
+
+  /**
+   * Generates a {@link SecretKey}
+   */
+  SecretKey generate();
+}
```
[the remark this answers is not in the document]
holachuy: Done

portability-transfer/src/main/java/org/dataportabilityproject/transfer/PortabilityTransferServiceProviderRegistry.java
```diff
@@ -84,4 +86,26 @@ public PortabilityTransferServiceProviderRegistry(List<String> enabledServices,
...
+   *
+   * @param transferDataType the transfer data type
+   */
+  @Override
+  public Set<String> getServices(String transferDataType) {
+    Preconditions.checkArgument(supportedExportTypes.contains(transferDataType),
+        "TransferDataType [%s] is not valid for export", transferDataType);
+    Preconditions.checkArgument(supportedImportTypes.contains(transferDataType),
+        "TransferDataType [%s] is not valid for import", transferDataType);
+    return serviceProviderMap.keySet();
```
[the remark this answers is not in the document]
holachuy: I was wondering the same thing. think the goal would be each exportable service also has an importer and vice versa so I wanted to avoid making the code that flexible and allowing for inconsistent services.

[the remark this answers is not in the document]
holachuy: Also, moved these methods to the AuthServiceProviderRegistry

portability-gateway/src/main/java/org/dataportabilityproject/gateway/action/createjob/CreateJobActionRequest.java
holachuy: Yes, very good point. We'll need to investigate a transport protocol and layer if we want to support more than http.

portability-gateway/src/main/java/org/dataportabilityproject/gateway/reference/DataTransferHandler.java
holachuy: Good point, added modules to each package.

portability-gateway/src/main/java/org/dataportabilityproject/gateway/reference/ReferenceApiModule.java
holachuy: Done

portability-gateway/src/main/java/org/dataportabilityproject/gateway/reference/ReferenceApiModule.java
holachuy: Done, good catch.

portability-gateway/src/main/java/org/dataportabilityproject/gateway/action/createjob/CreateJobActionRequest.java
rtannenbaum: another option along those lines, if we don't necessarily want to support RPC but still want to simplify the request/response/action logic, is Jersey.. http://www.mkyong.com/webservices/jax-rs/jersey-hello-world-example/

portability-gateway/src/main/java/org/dataportabilityproject/gateway/reference/ReferenceApiModule.java
```diff
@@ -0,0 +1,96 @@
...
+      public void uncaughtException(Thread thread, Throwable t) {
+        logger.warn("Uncaught exception in thread: {}", thread.getName(), t);
+      }
+    };
+  }
+
+  @Override
+  protected void configure() {
+    // TODO: Bind ApiSettings or migrate to launcher api context
+
```
seehamrun: Are we going to load the action modules in this module as well?  Or does it make sense to have a single ActionModule that binds everything needed for all Actions? (I'm not 100% sure on what the best way to split the modules up is).

holachuy: Added TODO

PR #136, opened 2018-02-13
Migrate JobState to the new modular structure

rtannenbaum:
For #132

### DIFF
--- a/extensions/cloud/portability-cloud-google/src/main/java/org/dataportabilityproject/cloud/google/GoogleCloudDatastore.java
+++ b/extensions/cloud/portability-cloud-google/src/main/java/org/dataportabilityproject/cloud/google/GoogleCloudDatastore.java
@@ -40,9 +40,9 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.UUID;
 import org.dataportabilityproject.spi.cloud.storage.JobStore;
+import org.dataportabilityproject.spi.cloud.types.JobAuthorization;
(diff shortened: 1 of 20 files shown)


rtannenbaum:
Whoops, forgot about JobAuthorization. Will make that change.

Now that I look at that more closely, the fact that State is now within JobAuthorization makes me think each state should be phrased in terms of auth, rather than worker

What do you think of renaming states to the following?
INITIAL -> INITIAL (unchanged)
// The gateway auth flow has completed and raw creds are temporarily available in the gateway server
PENDING_WORKER_ASSIGNMENT -> CREDS_AVAILABLE
// A worker has spun up and generated a key to encrypt the creds above so it (alone) may use the creds
ASSIGNED_WITHOUT_AUTH_DATA -> CREDS_ENCRYPTION_KEY_GENERATED
// The gateway server has encrypted the creds for the worker to use
ASSIGNED_WITH_AUTH_DATA ->        CREDS_ENCRYPTED

I can do that in a separate PR if we do it to keep things clean. Working on porting it over as-is for now.

Also, not really related, but WDYT of corrupting the raw creds in the API/gateway server once we have encrypted them, so they are not accessible in memory on the gateway server?

holachuy:
I like the state renaming, that's much better and we can do it in this PR.

Let's look at creds lifecycle lockdown after this PR to see where limiting access makes sense.

rtannenbaum:
Cool, made the changes in separate commits, PTAL

PR #139, opened 2018-02-14
AuthRegistry Implementation

seehamrun:


portability-spi-gateway/src/main/java/org/dataportabilityproject/spi/gateway/auth/AuthServiceProviderRegistry.java
```diff
@@ -4,11 +4,25 @@
...
-     *
-     * @param serviceId the service id
-     */
-    AuthServiceProvider getServiceProvider(String serviceId);
+  /**
+   * The AuthorizationMode to use for lookups.
+   * IMPORT specifies an authorization that allows you to import a type (implying read-write permissions)
+   * EXPORT specifies an authorization that allows you to export a type (implying read-only permissions)
+   */
+  enum AuthMode {
```
seehamrun: @holachuy  I know you added TransferMode in #134 - Does it make sense to either switch to that, or move that in here before merging?

holachuy: I don't mind going with AuthMode but the bigger question seems that it goes unused except in log statements. Can we remove it alltogether?

seehamrun: I was thinking it will get passed into the getAuthDataGenerator call to the AuthServiceProvider, which will then return a generator for the correct permissions mode.

holachuy: Per offline discussion, SGTM

seehamrun: Done - Added test. Also for posterity, this is meant to mimic the existing ServiceMode enum here: https://github.com/google/data-portability/blob/master/portability-core/src/main/java/org/dataportabilityproject/shared/ServiceMode.java

Used to determine scopes for auth generation (see GoogleServiceProvider for example: https://github.com/google/data-portability/blob/master/portability-core/src/main/java/org/dataportabilityproject/serviceProviders/google/GoogleServiceProvider.java#L64)

extensions/auth/portability-auth-microsoft/src/main/java/org/dataportabilityproject/auth/microsoft/MicrosoftAuthServiceProvider.java
```diff
@@ -17,8 +20,18 @@ public String getServiceId() {
...
 
-    public AuthDataGenerator getAuthDataGenerator(String transferDataType) {
+    public AuthDataGenerator getAuthDataGenerator(String transferDataType, AuthMode mode) {
         // TODO
         throw new UnsupportedOperationException();
     }
+
+    @Override
+    public List<String> getImportTypes() {
+        return ImmutableList.of();
```
holachuy: Do we need to put values here and below?

seehamrun: I left it blank since nothing was implemented yet for it but I didnt want to return null at the same time. I guess I could throw an Unsupported exception too. What makes the most sense?

PR #149, opened 2018-02-20
Google calendar

olsona:
Pulling GoogleCalendarService apart into converters and service
Getting started on tests for Google Calendar
Removing GooglePaginationInfo (StringPaginationToken works)


portability-core/src/main/java/org/dataportabilityproject/serviceProviders/google/GooglePaginationInfo.java
```diff
@@ -1,51 +0,0 @@
-/*
- * Copyright 2018 The Data-Portability Project Authors.
```
rtannenbaum: nice cleanup!

portability-core/src/main/java/org/dataportabilityproject/serviceProviders/google/calendar/GoogleCalendarToModelConverter.java
```diff
@@ -0,0 +1,62 @@
...
+import java.time.LocalDate;
+import java.time.OffsetDateTime;
+import java.time.ZoneId;
+import java.util.List;
+import java.util.stream.Collectors;
+import org.dataportabilityproject.dataModels.calendar.CalendarAttendeeModel;
+import org.dataportabilityproject.dataModels.calendar.CalendarEventModel;
+import org.dataportabilityproject.dataModels.calendar.CalendarModel;
+
+public class GoogleCalendarToModelConverter {
```
rtannenbaum: Not sure if it would make sense for these converters, but you might be able to use Converter<A, B> with doForward and doBackward for a more consistent API. See LegacyPortabilityJobConverter

olsona: I've been following the model Jim outlined earlier, but always happy to revisit!

rtannenbaum: ok SG!

Issue #154, opened 2018-02-23
Fix IDE support for AutoValue or revert to POJOs


holachuy:
If we choose to support developers using IDEs, which I think we should, the project should be buildable in an IDE consistently

This fix is inconsistent:
https://stackoverflow.com/questions/45001621/how-to-fix-cannot-resolve-autovalue-myclass-in-intellij

https://stackoverflow.com/questions/46424952/cannot-resolve-symbol-with-autovalue-and-intellij

PR #155, opened 2018-02-23
Fix Intellij issue so we don't have to manually add generated source roots

rtannenbaum:
With this change, IntelliJ build now works without any manual adding of source roots. Invalidate caches + restart still works as well.

PortabilityJobTest still runs as well, with the test root added in build.gradle.

I think something is still slightly off with a conflict between the declared package for org.dataportabilityproject.spi.cloud.types.AutoValue_PortabilityJob and the directory structure which includes an extra main/. However, this does not affect the IntelliJ build and is only noticeable if you go into an AutoValue generated impl class.

### DIFF
--- a/portability-spi-cloud/build.gradle
+++ b/portability-spi-cloud/build.gradle
@@ -24,11 +24,13 @@ dependencies {
 sourceSets {
     main {
         java {
+            // Includes generated AutoValue_ classes (build/classes/java) as source
             srcDirs = ['build/classes/java','src/main/java']
         }
     }
     test {
         java {
+            // Include test directories so we can build and run tests in IDE.
             srcDirs = ['src/test/java']
         }
     }
--- a/portability-spi-cloud/build.gradle
+++ b/portability-spi-cloud/build.gradle
@@ -20,3 +20,16 @@ dependencies {
     compile ("com.fasterxml.jackson.core:jackson-databind:${jacksonVersion}")
     compile("com.google.auto.value:auto-value:${autoValueVersion}")
 }
+
+sourceSets {
+    main {
+        java {
+            srcDirs = ['build/classes/java','src/main/java']
+        }
+    }
+    test {
+        java {
+            srcDirs = ['src/test/java']
+        }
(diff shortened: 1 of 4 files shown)


portability-spi-cloud/src/main/java/org/dataportabilityproject/spi/cloud/types/PortabilityJob.java
```diff
@@ -59,7 +59,7 @@
 
   public static PortabilityJob.Builder builder() {
     LocalDateTime now = LocalDateTime.now();
-    return new AutoValue_PortabilityJob.Builder()
+    return new org.dataportabilityproject.spi.cloud.types.AutoValue_PortabilityJob.Builder()
```
holachuy: Can we add a note as to why it has to be fully qualified? Does it have to happen for JobAuthorization as well?

portability-spi-cloud/build.gradle
```diff
@@ -20,3 +20,16 @@ dependencies {
     compile ("com.fasterxml.jackson.core:jackson-databind:${jacksonVersion}")
     compile("com.google.auto.value:auto-value:${autoValueVersion}")
 }
+
+sourceSets {
```
holachuy: Let's add comment

portability-spi-cloud/src/main/java/org/dataportabilityproject/spi/cloud/types/PortabilityJob.java
rtannenbaum: Yes, good call. IntelliJ built fine without it, but when you opened the class it would show red underlined AutoValue_ class. Now it's fixed. Added comments as well.

portability-spi-cloud/build.gradle
rtannenbaum: Done

rtannenbaum:
Awesome, thanks for your patience as I broke our dev environment during our migration! 

(Note: fixes #154)

PR #158, opened 2018-02-27
Google calendar

olsona:
Adding tests


portability-core/src/main/java/org/dataportabilityproject/serviceProviders/google/GoogleStaticObjects.java
```diff
@@ -64,4 +63,9 @@ public static HttpTransport getHttpTransport() {
   public static FileDataStoreFactory getDataStoreFactory() {
     return DATA_STORE_FACTORY;
   }
+
+  /**
+   * Static values used in converters
```
holachuy: Maybe add a little more doc on why MAX_ATTENDEES is set at this value

portability-core/src/test/java/org/dataportabilityproject/serviceProviders/google/calendar/GoogleCalendarServiceTest.java
```diff
@@ -21,72 +21,108 @@
...
+   * Sets up a response for a single event
+   */
+  private void setUpSingleEventResponse() throws IOException {
+    eventListResponse = new Events().setItems(Collections.singletonList(EVENT));
+    when(eventListRequest.execute()).thenReturn(eventListResponse);
   }
 
   @Test
-  public void testExportFirstSet() throws IOException {
+  public void testExportCalendarFirstSet() throws IOException {
```
holachuy: Would First or Single be more accurate for this test?

olsona: I think the fact that it's testing the export of the first calendar page is more important.

portability-core/src/main/java/org/dataportabilityproject/serviceProviders/google/GoogleStaticObjects.java
olsona: Adding a TODO to evaluate this.

PR #160, opened 2018-02-27
Create new library directory structure and security package

holachuy:
Introduce a new directory structure for shared 'library' type classes between different binaries. Security library will be for encrypting and decrypting data in the api/gateway and worker.


libraries/security/build.gradle
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+dependencies {
+    // Logging
+    compile("org.slf4j:slf4j-api:${slf4jVersion}")
+    compile("org.slf4j:slf4j-log4j12:${slf4jVersion}")
+}
+
```
seehamrun: nit: extra line

holachuy: Done

PR #164, opened 2018-03-01
Build worker without dependency on legacy portability core objects

holachuy:
Also removed storage of private key

### DIFF
--- a/portability-worker/build.gradle
+++ b/portability-worker/build.gradle
@@ -47,8 +47,11 @@ configurations {
 }
 
 dependencies {
-    compile project(':portability-core')
+    compile project(':portability-spi-cloud')
     compile project(':portability-spi-transfer')
+    compile project(':libraries:security')
+    // TODO: Make this test compile only
+    testCompile project(':extensions:cloud:portability-cloud-local')
 
     compile("com.flickr4java:flickr4java:${flickrVersion}")
     compile("com.google.apis:google-api-services-calendar:${googleCalendarVersion}")
@@ -60,11 +63,17 @@ dependencies {
     }
     compile("com.google.code.gson:gson:${gsonVersion}")
     compile("com.google.guava:guava:${guavaVersion}")
+    compile("com.google.inject.extensions:guice-multibindings:${guiceVersion}")
+    compile("com.google.inject:guice:${guiceVersion}")
     compile("com.google.http-client:google-http-client:${googleApiClient}")
     compile("com.google.http-client:google-http-client-xml:${googleApiClient}")
     compile("com.google.oauth-client:google-oauth-client:${googleApiClient}")
     compile("com.fasterxml.jackson.core:jackson-databind:${jacksonVersion}")
     compile("com.fasterxml.jackson.core:jackson-annotations:${jacksonVersion}")
+    // Logging
+    compile("org.slf4j:slf4j-api:${slf4jVersion}")
+    compile("org.slf4j:slf4j-log4j12:${slf4jVersion}")
+
 
     testCompile("com.google.truth:truth:${truthVersion}")
     testCompile("junit:junit:${junitVersion}")
--- a/portability-worker/build.gradle
+++ b/portability-worker/build.gradle
@@ -50,8 +50,6 @@ dependencies {
     compile project(':portability-spi-cloud')
     compile project(':portability-spi-transfer')
     compile project(':libraries:security')
-    // TODO: Make this test compile only
-    testCompile project(':extensions:cloud:portability-cloud-local')
 
     compile("com.flickr4java:flickr4java:${flickrVersion}")
     compile("com.google.apis:google-api-services-calendar:${googleCalendarVersion}")
@@ -78,4 +76,5 @@ dependencies {
     testCompile("com.google.truth:truth:${truthVersion}")
     testCompile("junit:junit:${junitVersion}")
     testCompile("org.mockito:mockito-all:${mockitoVersion}")
+    testCompile project(':extensions:cloud:portability-cloud-local')
 }
--- a/portability-worker/src/main/java/org/dataportabilityproject/worker/JobPollingService.java
+++ b/portability-worker/src/main/java/org/dataportabilityproject/worker/JobPollingService.java
@@ -45,14 +45,14 @@ class JobPollingService extends AbstractScheduledService {
     private final Logger logger = LoggerFactory.getLogger(JobPollingService.class);
     private final JobStore store;
     private final WorkerJobMetadata jobMetadata;
-    private final AsymmetricKeyGenerator aymmetricKeyGenerator;
+    private final AsymmetricKeyGenerator asymmetricKeyGenerator;
 
     @Inject
     JobPollingService(JobStore store, WorkerJobMetadata jobMetadata,
-                      AsymmetricKeyGenerator aymmetricKeyGenerator) {
+                      AsymmetricKeyGenerator asymmetricKeyGenerator) {
         this.store = store;
         this.jobMetadata = jobMetadata;
-        this.aymmetricKeyGenerator = aymmetricKeyGenerator;
+        this.asymmetricKeyGenerator = asymmetricKeyGenerator;
     }
 
     @Override
@@ -84,7 +84,7 @@ class JobPollingService extends AbstractScheduledService {
         }
         logger.debug("Polled job {}", jobId);
         Preconditions.checkState(!jobMetadata.isInitialized());
-        KeyPair keyPair = aymmetricKeyGenerator.generate();
+        KeyPair keyPair = asymmetricKeyGenerator.generate();
         PublicKey publicKey = keyPair.getPublic();
         // TODO: Move storage of private key to a different location
         PrivateKey privateKey = keyPair.getPrivate();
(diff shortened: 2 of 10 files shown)


portability-worker/build.gradle
```diff
@@ -47,8 +47,11 @@ configurations {
 }
 
 dependencies {
-    compile project(':portability-core')
+    compile project(':portability-spi-cloud')
     compile project(':portability-spi-transfer')
+    compile project(':libraries:security')
+    // TODO: Make this test compile only
```
rtannenbaum: done, right? can you move to the testCompile section?

holachuy:
This should be updated and rebased against previous reformatting changes.

portability-worker/src/main/java/org/dataportabilityproject/worker/JobPollingService.java
```diff
@@ -46,11 +45,14 @@
     private final Logger logger = LoggerFactory.getLogger(JobPollingService.class);
     private final JobStore store;
     private final WorkerJobMetadata jobMetadata;
+    private final AsymmetricKeyGenerator aymmetricKeyGenerator;
 
     @Inject
-    JobPollingService(CloudFactory cloudFactory, WorkerJobMetadata jobMetadata) {
-        this.store = cloudFactory.getJobStore();
+    JobPollingService(JobStore store, WorkerJobMetadata jobMetadata,
+                      AsymmetricKeyGenerator aymmetricKeyGenerator) {
```
rtannenbaum: nit: typo in variable name

portability-worker/src/main/java/org/dataportabilityproject/worker/JobPollingService.java
```diff
@@ -103,19 +105,19 @@ private void pollForUnassignedJob() throws IOException {
...
-    private void keyGenerated(UUID jobId, PublicKey publicKey, PrivateKey privateKey)
+    private void keyGenerated(UUID jobId, PublicKey publicKey)
             throws IOException {
         // Lookup the job so we can append to its existing properties.
         PortabilityJob existingJob = store.findJob(jobId);
         // Verify no worker key
         // Populate job with keys to persist
-        String encodedPublicKey = PublicPrivateKeyPairGenerator.encodeKey(publicKey);
-        String encodedPrivateKey = PublicPrivateKeyPairGenerator.encodeKey(privateKey);
+        Preconditions.checkArgument(existingJob.jobAuthorization().encryptedPublicKey() == null);
```
rtannenbaum: can you add a message, worker key should not be present in the PortabilityJob yet, as we're about to set it to the key we generated just now

portability-worker/src/main/java/org/dataportabilityproject/worker/JobPollingService.java
```diff
@@ -103,19 +105,19 @@ private void pollForUnassignedJob() throws IOException {
...
+        // Populate job with public key to persist
+        String encodedPublicKey = BaseEncoding.base64Url().encode(publicKey.getEncoded());
 
         PortabilityJob updatedJob = existingJob.toBuilder()
                 .setAndValidateJobAuthorization(existingJob.jobAuthorization().toBuilder()
                         .setEncryptedPublicKey(encodedPublicKey)
```
rtannenbaum: not related to this PR but while you're in here, can you change this to 'setEncodedPublicKey'?

actually, does it even still need to be encoded? was that just so we could pass to the client as a backup, and are we still doing that?

portability-worker/src/main/java/org/dataportabilityproject/worker/PortabilityInMemoryTransferCopier.java
```diff
@@ -37,95 +37,99 @@
...
-              new ExportInformation(null, resource));
-        }
-      }
+    private static final AtomicInteger COPY_ITERATION_COUNTER = new AtomicInteger();
+    private static final Logger logger = LoggerFactory
+            .getLogger(PortabilityInMemoryTransferCopier.class);
+
+    private final TransferServiceProviderRegistry registry;
+
+    public PortabilityInMemoryTransferCopier(TransferServiceProviderRegistry registry) {
```
rtannenbaum: this one is mostly indent changes right? is the new ctor the only change?

portability-worker/src/main/java/org/dataportabilityproject/worker/WorkerImpl.java
```diff
@@ -83,29 +80,31 @@ private void pollForJob() {
...
                     decrypter.decrypt(jobAuthorization.encryptedExportAuthData());
             AuthData exportAuthData = deSerialize(serializedExportAuthData);
             String serializedImportAuthData =
                     decrypter.decrypt(jobAuthorization.encryptedImportAuthData());
             AuthData importAuthData = deSerialize(serializedImportAuthData);
-            PortabilityCopier
-                    .copyDataType(registry, dataType, job.exportService(), exportAuthData,
-                            job.importService(), importAuthData, jobId);
+            // TODO: Refactor copy logic
+            InMemoryTransferCopier copier = new PortabilityInMemoryTransferCopier(registry);
```
rtannenbaum: Can copier be injected? (once its ctor is annotated with @Inject)

portability-worker/src/main/java/org/dataportabilityproject/worker/PortabilityInMemoryTransferCopier.java
```diff
@@ -37,95 +37,99 @@
...
+     *
+     * @param dataType       The data type to transfer
+     * @param exportService  The export service to transfer from
+     * @param exportAuthData The auth data for the export service
+     * @param importService  The import service to transfer to
+     * @param importAuthData The auth data for the import service
+     * @param jobId          The job id representing this copy
+     */
+    @Override
+    public void copyDataType(String dataType,
```
rtannenbaum: A future refactor to consider, but I think would be difficult in the current intermediate state with some copying still being done in the API etc.

Each worker instance should be spun up with one (eventual) definition of all these properties (export service, auth data, job ID, etc) although they will not be available until a job is polled. WDYT of injecting all these fields with providers in the copier, so they are only evaluated once we start the copy, at which point the job is polled?

I think we should follow the same model for local worker testing too - spin up/tear down for each job tested locally

portability-worker/src/main/java/org/dataportabilityproject/worker/PortabilityInMemoryTransferCopier.java
rtannenbaum: @Inject?

portability-worker/src/main/java/org/dataportabilityproject/worker/JobPollingService.java
holachuy: Done

portability-worker/src/main/java/org/dataportabilityproject/worker/JobPollingService.java
holachuy: Done

portability-worker/src/main/java/org/dataportabilityproject/worker/JobPollingService.java
holachuy: Done. At some point, we will

portability-worker/src/main/java/org/dataportabilityproject/worker/PortabilityInMemoryTransferCopier.java
holachuy: Yes, sorry, that's the only change.

portability-worker/src/main/java/org/dataportabilityproject/worker/PortabilityInMemoryTransferCopier.java
holachuy: Yes, we can refactor as to how to enforce the singleton nature of the data we want to pass around. Maybe an singleton annotated object with these params or individual data.

portability-worker/src/main/java/org/dataportabilityproject/worker/WorkerImpl.java
holachuy: Yes, the TODO should account for that.

portability-worker/build.gradle
rtannenbaum: Did you see this one?

portability-worker/src/main/java/org/dataportabilityproject/worker/PortabilityInMemoryTransferCopier.java
rtannenbaum: Did you see this one?

rtannenbaum: ok cool, thanks

holachuy:
Still working on why the test dependency doesn't pick up

> Task :portability-worker:compileTestJava FAILED
/home/chuy/Desktop/wt/data-portability/portability-worker/src/test/java/org/dataportabilityproject/worker/JobPollingServiceTest.java:18: error: cannot find symbol
import org.dataportabilityproject.cloud.local.InMemoryKeyValueStore;
                                             ^
  symbol:   class InMemoryKeyValueStore
  location: package org.dataportabilityproject.cloud.local
1 error

Issue #172, opened 2018-03-05
Wire up the launchers for the worker


seehamrun:

rtannenbaum:
Spoke with @holachuy , I'm going to start this one before gateway

rtannenbaum:
These should be run via the build script for the demo

PR #184, opened 2018-03-07
Refactor worker

rtannenbaum:
for #172


portability-worker/src/main/java/org/dataportabilityproject/worker/WorkerModule.java
```diff
@@ -1,9 +1,95 @@
...
+      }
+
+      @Override
+      public <T> T getService(Class<T> type) {
+        return null;
+      }
+
+      @Override
+      public <T> T getConfiguration(String key, String defaultValue) {
+        return null;
```
holachuy: TODO pass in flags?

portability-worker/src/main/java/org/dataportabilityproject/worker/JobMetadata.java
```diff
@@ -16,66 +16,67 @@
...
 import java.util.UUID;
 
 /**
- * A class that contains the value of the job and key for a worker instance. This classes uses the
- * initialization-on-demand holder idiom to ensure it is a singleton.
+ * A class that contains metadata for a worker's job.
+ *
+ * <p>This class is completely static to ensure it is a singleton within each worker instance.
  */
 @Singleton
```
holachuy: We can remove @Singleton because iff we're using the static singleton approach, this won't be guice injected

portability-worker/src/main/java/org/dataportabilityproject/worker/WorkerModule.java
rtannenbaum: Done, thanks

portability-worker/src/main/java/org/dataportabilityproject/worker/JobMetadata.java
rtannenbaum: Done

PR #186, opened 2018-03-07
Bind and configure AuthServiceProviderRegistry to new gateway

seehamrun:
WIP: This does not dynamically populate the registry yet. Need to wire up a few more things before we have something runnable to check against. Also only one AuthProvider is currently set up.


seehamrun:
I made the changes we discussed offline: 
-> only binding services that are enabled (as determined by flags in main or something else)
-> moving installation of auth registry module to the main class and not as part of the api module. 

PTAL

portability-gateway/src/main/java/org/dataportabilityproject/gateway/PortabilityAuthServiceProviderModule.java
```diff
@@ -0,0 +1,45 @@
...
+import org.slf4j.LoggerFactory;
+
+public class PortabilityAuthServiceProviderModule extends AbstractModule {
+
+  private static final Logger logger =
+      LoggerFactory.getLogger(PortabilityAuthServiceProviderModule.class);
+
+  private final ImmutableList<String> enabledServices;
+
+  PortabilityAuthServiceProviderModule(List<String> enabledServices) {
```
holachuy: Maybe this should be a Set unless we care about order.

portability-gateway/src/main/java/org/dataportabilityproject/gateway/PortabilityAuthServiceProviderModule.java
```diff
@@ -0,0 +1,45 @@
...
+import org.dataportabilityproject.spi.gateway.auth.AuthServiceProviderRegistry;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class PortabilityAuthServiceProviderModule extends AbstractModule {
+
+  private static final Logger logger =
+      LoggerFactory.getLogger(PortabilityAuthServiceProviderModule.class);
+
+  private final ImmutableList<String> enabledServices;
```
holachuy: ImmutableSet ?

portability-gateway/src/main/java/org/dataportabilityproject/gateway/PortabilityAuthServiceProviderRegistry.java
```diff
@@ -33,34 +34,29 @@
...
-            type,
-            service);
-        supportedImportTypesBuilder.add(type);
-      }
-
-      supportedExportTypesBuilder.addAll(exportTypes);
-      serviceProviderBuilder.put(service, authServiceProvider);
-    }
+    serviceProviderMap.forEach(
+        (service, provider) -> {
```
holachuy: fanceee!!!!

portability-gateway/src/main/java/org/dataportabilityproject/gateway/PortabilityAuthServiceProviderModule.java
seehamrun: Done

portability-gateway/src/main/java/org/dataportabilityproject/gateway/PortabilityAuthServiceProviderModule.java
seehamrun: Done

Issue #98, opened 2018-01-26
Race condition: 2 workers poll the same job


rtannenbaum:
API:
2018-01-26 20:49:36,196 [http-server-16] INFO  org.dataportabilityproject.job.PortabilityJobFactory - Creating new PortabilityJob, id: ed03f0ba-0b8c-4d55-bce0-70c9c747f35e

Worker 1:

2018-01-26 20:50:12,244 [JobPollingService RUNNING] DEBUG org.dataportabilityproject.worker.JobPollingService - Polled pollForUnassignedJob, found id: ed03f0ba-0b8c-4d55-bce0-70c9c747f35e
2018-01-26 20:50:12,943 [JobPollingService RUNNING] DEBUG org.dataportabilityproject.worker.JobPollingService - Completed updateJobStateToAssignedWithoutAuthData, publicKey: 162

2018-01-26 20:50:33,051 [JobPollingService RUNNING] DEBUG org.dataportabilityproject.worker.JobPollingService - Polled lookupAssignedWithAuthDataJob, found auth data, id: ed03f0ba-0b8c-4d55-bce0-70c9c747f35e
2018-01-26 20:50:33,056 [main] DEBUG org.dataportabilityproject.worker.WorkerImpl - Begin processing jobId: ed03f0ba-0b8c-4d55-bce0-70c9c747f35e
Jan 26, 2018 8:50:33 PM com.google.common.util.concurrent.UncaughtExceptionHandlers$Exiter uncaughtException
SEVERE: Caught an exception in Thread[main,5,main].  Shutting down.
java.lang.RuntimeException: Failed to invoke public org.dataportabilityproject.shared.auth.AuthData() with no args
	at com.google.gson.internal.ConstructorConstructor$3.construct(ConstructorConstructor.java:107)
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:162)
	at com.google.gson.Gson.fromJson(Gson.java:803)
	at com.google.gson.Gson.fromJson(Gson.java:768)
	at com.google.gson.Gson.fromJson(Gson.java:717)
	at com.google.gson.Gson.fromJson(Gson.java:689)
	at org.dataportabilityproject.worker.WorkerImpl.deSerialize(WorkerImpl.java:108)
	at org.dataportabilityproject.worker.WorkerImpl.processJob(WorkerImpl.java:92)
[...]

Issue #176, opened 2018-03-06
Checked or Unchecked exceptions


jimmarino:
The issue of checked vs. unchecked exceptions has been raised on Slack. Based on the outcome of that discussion, changes will need to be made to harmonize the codebase.

rtannenbaum:
Jim's comment from PR #167 about this:

"I would argue for unchecked exceptions except in cases where it is absolutely necessary to signal a condition is recoverable. Runtime exceptions can bubble up to the top of the stack where they can be trapped and logged. "

PR #187, opened 2018-03-07
Reintroduce worker race condition fix from #98 that was lost during the migration

rtannenbaum:
If we go with this approach, I put a placeholder in the Cosmos Store to implement this

### DIFF
--- a/extensions/cloud/portability-cloud-google/src/main/java/org/dataportabilityproject/cloud/google/GoogleJobStore.java
+++ b/extensions/cloud/portability-cloud-google/src/main/java/org/dataportabilityproject/cloud/google/GoogleJobStore.java
@@ -54,49 +54,12 @@ public final class GoogleJobStore implements JobStore {
     this.datastore = datastore;
   }
 
-  private static Map<String, Object> getProperties(Entity entity) {
-    if (entity == null) return null;
-    ImmutableMap.Builder<String, Object> builder = new ImmutableMap.Builder<>();
-    for (String property : entity.getNames()) {
-      // builder.put(property, entity.getValue(property));
-      if (entity.getValue(property) instanceof StringValue) {
-        builder.put(property, (String) entity.getString(property));
-      } else if (entity.getValue(property) instanceof LongValue) {
-        // This conversion is safe because of integer to long conversion above
-        builder.put(property, new Long(entity.getLong(property)).intValue());
-      } else if (entity.getValue(property) instanceof DoubleValue) {
-        builder.put(property, (Double) entity.getDouble(property));
-      } else if (entity.getValue(property) instanceof BooleanValue) {
-        builder.put(property, (Boolean) entity.getBoolean(property));
-      } else if (entity.getValue(property) instanceof TimestampValue) {
-        builder.put(property, (Timestamp) entity.getTimestamp(property));
-      } else {
-        Blob blob = entity.getBlob(property);
-        Object obj = null;
-        try {
-          try (ObjectInputStream in = new ObjectInputStream(blob.asInputStream())) {
-            try {
-              obj = in.readObject();
-            } catch (ClassNotFoundException e) {
-              e.printStackTrace();
-            }
-          }
-        } catch (IOException e) {
-          e.printStackTrace();
-        }
-        builder.put(property, obj); // BlobValue
-      }
-    }
-
-    return builder.build();
-  }
-
   /**
-   * Inserts a new {@link PortabilityJob} keyed by its job ID in Datastore.
+   * Inserts a new {@link PortabilityJob} keyed by {@code jobId} in Datastore.
    *
    * <p>To update an existing {@link PortabilityJob} instead, use {@link #update}.
    *
-   * @throws IOException if a job already exists for {@code job}'s ID, or if there was a different
+   * @throws IOException if a job already exists for {@code jobId}, or if there was a different
    *     problem inserting the job.
    */
   @Override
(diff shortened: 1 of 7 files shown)


extensions/cloud/portability-cloud-google/src/main/java/org/dataportabilityproject/cloud/google/GoogleJobStore.java
```diff
@@ -142,6 +118,11 @@ public void updateJob(UUID jobId, PortabilityJob job) throws IOException {
         throw new IOException("Could not find record for jobId: " + jobId);
       }
 
+      if (previousState != null) {
+        PortabilityJob previousJob = PortabilityJob.fromMap(getProperties(previousEntity));
```
holachuy: For a future PR: Maybe if we json serialize the job, we won't have to go from entity -> map -> job object.

portability-worker/src/main/java/org/dataportabilityproject/worker/JobPollingService.java
```diff
@@ -119,7 +115,16 @@ private void keyGenerated(UUID jobId, KeyPair keyPair) throws IOException {
                     .setEncodedPublicKey(encodedPublicKey)
                     .setState(JobAuthorization.State.CREDS_ENCRYPTION_KEY_GENERATED)
                     .build()).build();
-    store.updateJob(jobId, updatedJob);
+    // Attempt to 'claim' this job by validating it is still in state CREDS_AVAILABLE as we
+    // update it to state CREDS_ENCRYPTION_KEY_GENERATED, along with our key. If another worker
+    // instance polled the same job, and already claimed it, it will have updated the job's state
+    // to CREDS_ENCRYPTION_KEY_GENERATED.
+    try {
+      store.updateJob(jobId, updatedJob, JobAuthorization.State.CREDS_AVAILABLE);
```
holachuy: I think we can ensure here that the existingJob state is CREDS_AVAILABLE rather than pass it into the JobStore

portability-api/src/main/java/org/dataportabilityproject/webapp/StartCopyHandler.java
```diff
@@ -124,13 +124,13 @@ private void handleWorkerAssignmentFlow(HttpExchange exchange, UUID jobId) throw
     Preconditions.checkArgument(
         !Strings.isNullOrEmpty(importAuthCookieValue), "Import auth cookie required");
 
-    // We have the data, now update to 'pending worker assignment' so a worker may be assigned
+    // We have the data, now update to 'CREDS_AVAILABLE' so a worker may poll the job.
```
holachuy: We also have the existing job here and can validate it was INITIAL before setting the new state.

portability-spi-cloud/src/main/java/org/dataportabilityproject/spi/cloud/storage/JobStore.java
```diff
@@ -55,12 +55,29 @@ default void update(UUID jobId, LegacyPortabilityJob job, JobAuthorization.State
...
+   *
+   * <p>For auth-state-changing operations, the {@code previousState} may be validated as part of
+   * the transaction. Set {@code previousState} to null if no validation is desired.
+   *
+   * @throws IOException if a job didn't already exist for {@code jobId} or there was a problem
+   * updating it
+   * @throws IllegalStateException if state validation was requested ({@code previousState} was
+   * non-null) and the job's state was not the expected {@code previousState}
+   */
+  void updateJob(UUID jobId, PortabilityJob job, JobAuthorization.State previousState)
```
holachuy: I'm weary of adding the third param simply for validation since this class mostly just cares about storage. Since we usually have the existing job object before updateJob is called, can we validate before the updateJob method is called rather than having it leak into the JobStore?

portability-worker/src/main/java/org/dataportabilityproject/worker/JobPollingService.java
```diff
@@ -119,7 +115,16 @@ private void keyGenerated(UUID jobId, KeyPair keyPair) throws IOException {
...
+    } catch (IllegalStateException e) {
+      throw new IOException("Could not 'claim' job " + jobId + ". It was probably already "
```
holachuy: Good catch.

portability-spi-cloud/src/main/java/org/dataportabilityproject/spi/cloud/storage/JobStore.java
holachuy: Good description of the issue. Ok, let's move forward with this.

It's up to you if you want to do this now or later, but maybe we can pass in an JobUpdateValidator interface that the client can provide to wrap this logic.

 void updateJob(UUID jobId, PortabilityJob job, JobUpdateValidator validator)

Something like
inteface JobUpdateValidator {
  /** Validates before update. */
  void validate(PortabilityJob previous, PortabilityJob updated);

rtannenbaum: Nice, I like that solution, I went with that

portability-api/src/main/java/org/dataportabilityproject/webapp/StartCopyHandler.java
rtannenbaum: Added TODO since this is the LegacyPortabilityJob impl

jimmarino:
I'll take a look at how to handle this in Cosmos DB. It claims to be transactional so it shold be OK.

portability-spi-cloud/src/main/java/org/dataportabilityproject/spi/cloud/storage/JobStore.java
```diff
@@ -15,9 +15,16 @@
  * back-end services.
  */
 public interface JobStore {
+  interface JobUpdateValidator {
+    /**
+     * Validation to do as part of an atomic update. Implementers should throw an
+     * {@code IllegalStateException} if the validation fails.
+     */
+    void validate(PortabilityJob previous, PortabilityJob updated);
```
holachuy: So clients can see it in the contract of the method signature, let's explicitly declare throws IllegalStateException or a custom ValidationException

rtannenbaum: IllegalStateException is unchecked, and this is the same convention as Preconditions.checkState (i.e. that also documents @throws IllegalStateException but does not explicitly declare in method signature)

I think it's probably unrecoverable if validate fails, so should stick with an unchecked exception, depending on what we decide for #176. Can we revisit this with #176?

holachuy: Sure, I'm thinking more about defining explicitly how a client handles exceptions to the validation to avoid a mishmash of inconsistent exception handling.

Issue #192, opened 2018-03-08
Implement CloudExtension for all Cloud Implementations


seehamrun:
This is so we can use the ServiceLoader to grab the right jobstore based on configuration

jimmarino:
OK sounds good

rtannenbaum:
Forgot to tag this in pr #201 which makes some changes to Google's cloud extension

PR #193, opened 2018-03-08
Implementation of GoogleCloudExtensions

seehamrun:
I added getCryptoKeyStore and getBucketStore to the CloudExtension since I suspect they might be needed (the CloudFactory had these). 

I'm not sure of where the initialization of the constructor params should go. My initial thought was to pull them in from the ExtensionContext, but its not clear to me what that looks like for things like creds or the json factories. 

#192

### DIFF
--- a/extensions/cloud/portability-cloud-google/src/main/java/org/dataportabilityproject/cloud/google/GoogleCloudExtension.java
+++ b/extensions/cloud/portability-cloud-google/src/main/java/org/dataportabilityproject/cloud/google/GoogleCloudExtension.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 The Data-Portability Project Authors.
+ * Copyright 2018 The Data Transfer Project Authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/extensions/cloud/portability-cloud-google/src/main/java/org/dataportabilityproject/cloud/google/GoogleCloudExtension.java
+++ b/extensions/cloud/portability-cloud-google/src/main/java/org/dataportabilityproject/cloud/google/GoogleCloudExtension.java
@@ -1,13 +1,11 @@
 package org.dataportabilityproject.cloud.google;
 
-import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport;
 import com.google.api.client.http.HttpTransport;
 import com.google.api.client.json.JsonFactory;
-import com.google.api.client.json.jackson2.JacksonFactory;
 import com.google.auth.oauth2.GoogleCredentials;
 import com.google.cloud.datastore.Datastore;
-import com.google.cloud.datastore.DatastoreOptions;
 import com.google.common.base.Preconditions;
+import com.google.common.base.Strings;
 import org.dataportabilityproject.api.launcher.ExtensionContext;
 import org.dataportabilityproject.spi.cloud.extension.CloudExtension;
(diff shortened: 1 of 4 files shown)


extensions/cloud/portability-cloud-google/src/main/java/org/dataportabilityproject/cloud/google/GoogleCloudExtension.java
```diff
@@ -0,0 +1,89 @@
...
+import org.dataportabilityproject.api.launcher.ExtensionContext;
+import org.dataportabilityproject.spi.cloud.extension.CloudExtension;
+import org.dataportabilityproject.spi.cloud.storage.BucketStore;
+import org.dataportabilityproject.spi.cloud.storage.CryptoKeyStore;
+import org.dataportabilityproject.spi.cloud.storage.JobStore;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class GoogleCloudExtension implements CloudExtension {
+  // TODO: Should these be initialized from the ExtensionContext and not injected?
```
holachuy: Yep, the context sounds right

extensions/cloud/portability-cloud-google/src/main/java/org/dataportabilityproject/cloud/google/GoogleCloudExtension.java
```diff
@@ -0,0 +1,89 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(GoogleCloudExtension.class);
+
+  private boolean initialized = false;
+  private Datastore datastore;
+  private JobStore jobstore;
+  private BucketStore bucketStore;
+  private GoogleCryptoKeyStore cryptoKeyManagementSystem;
+
+  @Inject
+  GoogleCloudExtension(
```
holachuy: I think extensions will have no constructor if I understand correctly?

rtannenbaum: I think just a public no-arg one right? I think that's required by the ServiceLoader

seehamrun: Changed to have no constructor. I'm not convinced I'm creating the transport and the json factory correctly, but thats how it was done throughout the project - I don't have anything runnable to test it with though

extensions/cloud/portability-cloud-google/src/main/java/org/dataportabilityproject/cloud/google/GoogleCloudExtension.java
```diff
@@ -0,0 +1,85 @@
...
+
+      // TODO: Hook these up with the global instances
+      transport =  GoogleNetHttpTransport.newTrustedTransport();
+      jsonFactory = new JacksonFactory();
+      cryptoKeyManagementSystem = new GoogleCryptoKeyStore(transport, jsonFactory, projectId);
+
+      initialized = true;
+    } catch (IOException | GoogleCredentialException | GeneralSecurityException e) {
+      // TODO: the method doesn't throw an exception, how do we pass this onto the user?
+      logger.debug("Error initializing extension: " + this.getClass().getName(), e);
```
holachuy: logger.warn

seehamrun:
Oh good point Jim! The extension context is something that would get created in our Main class, correct? I think that works because at that point we should know what each extension requires. 

I've switched to pulling them from the ExtensionContext with .getService(); but havent written up the context yet.

PR #204, opened 2018-03-11
Create the initial outline for an integration guide

holachuy:
This starts to capture the areas that have to be documented for integration providers.


holachuy:
Issue #207

PR #205, opened 2018-03-12
Start on support for MS photo transfer

jimmarino:


extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/dataportabilityproject/transfer/microsoft/photos/MicrosoftPhotosExporter.java
```diff
@@ -0,0 +1,253 @@
...
+  private List<Map<String, Object>> requestAllPhotoItems(TokenAuthData authData)
+      throws IOException {
+    List<Map<String, Object>> photoItems = new ArrayList<>();
+
+    Deque<Map<String, Object>> folderItems = new ArrayDeque<>();
+    requestItems(photosRootUrl, authData, folderItems, photoItems);
+
+    // request items for all folders recursively until the collection is exhausted, adding contained
+    // folders to the end of the deque for processing
+    while (!folderItems.isEmpty()) {
```
holachuy: Rather than fetch everything in memory at once and store, the design we've evolved is to fetch the top level node with it's photos and any sub (nested) albums, then return that export result for import along with continuation data which then re-enters the copier, and so on recursively through the tree.   This is primarily for scalability but also for reliability when we add retry support.

marleyjustin08 (commented):
#924 `

extensions/data-transfer/portability-data-transfer-flickr/src/main/java/org/dataportabilityproject/datatransfer/flickr/photos/FlickrPhotosExporter.java
```diff
@@ -54,6 +48,13 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.ArrayList;
```
marleyjustin08, line 51: Metaform:calendar-updates

marleyjustin08 (commented):
🤘

marleyjustin08:
Thank u

PR #215, opened 2018-03-12
Add Shadow Jar-based distro for the default worker process

jimmarino:


distributions/worker-default/build.gradle
```diff
@@ -0,0 +1,40 @@
...
+}
+
+
+repositories {
+    jcenter()
+}
+
+dependencies {
+    compile project(':portability-worker')
+    compile project(':extensions:data-transfer:portability-data-transfer-google')
```
holachuy: Do we need to include a cloud extension?

build.gradle
```diff
@@ -61,3 +69,12 @@ configure(sourceProjects()) {
     }
 
 }
+
+def addCloudExtensionDependency(proj) {
+
+    if (proj.rootProject.ext.cloudType == "google") {
+        proj.dependencies { compile project(':extensions:cloud:portability-cloud-google') }
```
seehamrun: do we need this case statement? Can we require that a cloud type be set and just use "compile project(':extensions:cloud:portability-cloud-$cloudType')"

jimmarino: Good point. I'll add that in

build.gradle
```diff
@@ -36,6 +41,9 @@ plugins {
     id 'java'
 }
 
+ext.cloudType = hasProperty('cloudType') ? cloudType : 'default'
```
seehamrun: should we have the default be "local" - I think we talked about moving the cloud / datastore over to its own extension. We already have a local worker too: https://github.com/google/data-transfer-project/tree/master/extensions/cloud/portability-cloud-local/src/main/java/org/dataportabilityproject/cloud/local

seehamrun:
Thanks Jim!

PR #216, opened 2018-03-12
Start on service extension types

jimmarino:


portability-worker/src/main/java/org/dataportabilityproject/worker/WorkerExtensionContext.java
```diff
@@ -38,11 +40,22 @@
           .put(JsonFactory.class, new JacksonFactory())
           .build();
 
+  private final TypeManager typeManager;
+
+  public WorkerExtensionContext() {
+    this.typeManager = new TypeManagerImpl();
```
holachuy: Should we add a todo to use create an initialize a TypeManager with all the registered types?

portability-worker/src/main/java/org/dataportabilityproject/worker/WorkerExtensionContext.java
```diff
@@ -38,11 +40,22 @@
           .put(JsonFactory.class, new JacksonFactory())
```
holachuy: Do we remove the JacksonFactory now that we have the type manager?

portability-api-launcher/src/main/java/org/dataportabilityproject/api/launcher/TypeManager.java
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.dataportabilityproject.spi.cloud.types;
+package org.dataportabilityproject.api.launcher;
```
seehamrun: We briefly talked about this in the call, should we rename portability-api-launcher? Since nothing gets "launched" from here and this isnt the entry point of the api. Also the context and extensions used are not unique to the gateway.

portability-worker/src/main/java/org/dataportabilityproject/worker/WorkerExtensionContext.java
jimmarino: OK let me add that

portability-worker/src/main/java/org/dataportabilityproject/worker/WorkerExtensionContext.java
jimmarino: Sure, I'll do that in this PR

portability-api-launcher/src/main/java/org/dataportabilityproject/api/launcher/TypeManager.java
jimmarino: Sounds good to me. Should we do this in another PR? Also, does anyone have a preference/suggestion for name?

seehamrun: another PR sgtm - how about "portability-bootstrap" or maybe "portability-base" 🤔 naming things is hard.

jimmarino: bootstrap sounds good

PR #220, opened 2018-03-12
Moving GoogleContacts over, starting with Exporter

olsona:


extensions/data-transfer/portability-data-transfer-google/src/main/java/org/dataportabilityproject/datatransfer/google/common/GoogleStaticObjects.java
```diff
@@ -73,10 +73,23 @@ public static FileDataStoreFactory getDataStoreFactory() {
...
-   * Static values used in converters
+   * Static values used in calendar converters
    */
   // TODO(olsona): find optimum value
   public static final int MAX_ATTENDEES = 100;
   public static final String CALENDAR_TOKEN_PREFIX = "calendar:";
   public static final String EVENT_TOKEN_PREFIX = "event:";
+
+  /**
+   * Static values used in contacts converters
```
holachuy: As we add more, we should consider breaking out these constants into a service specific folder as opposed to this constant package since they are not actually common to more than one service.

olsona: Agreed.

Issue #202, opened 2018-03-10
Move GCP-specific config into Google distribution


rtannenbaum:
e.g. config/gcp/* should be in extensions/ somewhere for Google

rtannenbaum:
This is blocked by #323. demo-google-deployment was (knowingly) broken by #321. Once we update the google deployment to use the new gradle-based docker build, we can then get the rest of the environment configs (qa, test, prod) over to the Google distro.

Issue #171, opened 2018-03-05
Wire up the launchers for the api


seehamrun:

rtannenbaum:
This should be run via the build script for the demo

PR #224, opened 2018-03-13
Add distribution for portability-gateway

seehamrun:
Update build script to build distribution

this can be run now with 
`
$ ./bin/build_and_run_jar.sh gateway local google 
$ ./bin/build_and_run_jar.sh worker local google 
`

PR #226, opened 2018-03-13
Provide ApiSettings in ReferenceApiModule, using parsed yaml settings

rtannenbaum:
Unlike ExtensionContext flags, which may be arbitrary, validate the expected API flags baseUrl and baseApiUrl with Jackson.

ApiSettings is injected throughout the Reference Api server

#171  #202


rtannenbaum:
Let me know if this makes any sense... I don't think we should be using ExtensionContext for this

extensions/cloud/portability-cloud-google/config/common.yaml
```diff
@@ -0,0 +1,2 @@
+# Google cloud settings. See environments/ for environment-specific settings.
+cloud: GOOGLE
```
holachuy: Can we add a list of example options to the comment

libraries/security/src/main/java/org/dataportabilityproject/security/ConfigUtils.java
```diff
@@ -0,0 +1,54 @@
...
+import com.google.common.collect.ImmutableList;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.SequenceInputStream;
+import java.util.Map;
+
+// TODO(rtannenbaum): Move this out of security/ into a new module config/. Couldn't figure it
+// out in Intellij.
+public class ConfigUtils {
+  public static Map<String, Object> parse(InputStream in) throws IOException {
```
holachuy: Document this public method

libraries/security/src/main/java/org/dataportabilityproject/security/ConfigUtils.java
```diff
@@ -0,0 +1,54 @@
...
+    ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
+    return mapper.readValue(in, Map.class);
+  }
+
+  public static InputStream getSettingsCombinedInputStream(ImmutableList<String> settingsFiles) {
```
holachuy: same

portability-worker/src/main/java/org/dataportabilityproject/worker/WorkerExtensionContext.java
```diff
@@ -28,13 +29,16 @@
...
 import org.dataportabilityproject.launcher.impl.TypeManagerImpl;
+import org.dataportabilityproject.security.ConfigUtils;
 
 /**
  * {@link ExtensionContext} used by the worker.
  */
 final class WorkerExtensionContext implements ExtensionContext {
-  private static final String CLOUD_SETTINGS_PATH = "config/settings.yaml";
-  private static final String CLOUD_ENV_SETTINGS_PATH = "config/env/settings.yaml";
+  private static final String WORKER_SETTINGS_PATH = "config/worker.yaml";
```
holachuy: Are all these distinct? Can they be externally configured?

holachuy:
I patched this in and tried to run the following command but it didn't find a cloud type. Do we have to update a config file?

./bin/build_and_run_jar.sh worker local
ERROR: Must provide a cloud type, e.g. 'local', 'google', or 'microsoft'

rtannenbaum:
@holachuy that's broken in master as well. I can fix that in my next PR when I make the script compile the yaml files into the jar

extensions/cloud/portability-cloud-google/config/common.yaml
rtannenbaum: I think this is the only common setting left actually now that encryptionFlow and serviceProviderClasses are deprecated

libraries/security/src/main/java/org/dataportabilityproject/security/ConfigUtils.java
rtannenbaum: done

libraries/security/src/main/java/org/dataportabilityproject/security/ConfigUtils.java
rtannenbaum: done

seehamrun:
Sorry that might have been me that broke it - I added a 3rd option in #224 so we can configure which cloud package to compile into the jar

bin/build_and_run_jar.sh
```diff
@@ -56,9 +56,9 @@ gradle=$(which gradle)|| { echo "Gradle (gradle) not found. Please install it an
 
 # Copy settings yaml files from ENV/settings/ into $SRC_DIR/src/main/resources/
 # TODO: this script should take cloud extension name as a param, in addition to ENV.
-# It should copy settings from extension-cloud-dir/config/settings.yaml into
-# config/settings.yaml, and extension-cloud-dir/config/environments/$ENV/settings.yaml into
-# config/env/settings.yaml.
+# It should copy common and api/worker settings from extension-cloud-dir/config/ into config/, and
```
seehamrun: nit: gateway/worker

Issue #208, opened 2018-03-12
ServiceLoad appropriate extensions in gateway and worker


rtannenbaum:

rtannenbaum:
Blocks #171 and #172

PR #229, opened 2018-03-13
User serviceLoader for loading the CloudExtension in worker-main

seehamrun:
#208

### DIFF
--- a/portability-worker/build.gradle
+++ b/portability-worker/build.gradle
@@ -47,10 +47,6 @@ configurations {
 }
 
 dependencies {
-    // TODO(seehamrun): this is here for testing hardcoded Google cloud extension, is it needed
-    // with ServiceLoading?
-    compile project(':extensions:cloud:portability-cloud-google')
-
     compile project(':portability-spi-service')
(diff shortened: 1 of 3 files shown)


rtannenbaum:
woot! does it run locally?

seehamrun:
I'm having some issues running locally using the build script since I couldnt get the docker env working. 

But running locally through gradle manually, I verified that the right ExtensionContext was picked in the logger after removing the explicit dependencies on GoogleCloud

rtannenbaum:
If you want to test it w/ Google Cloud you'll need to use the docker script since it sets GCP-specific environment vars within the container. You'll need to download service account creds for the project too if you don't have them.

./config/gcp/build_and_upload_docker_image.sh worker local qa

I have the files set up so can try it in the office, feel free to merge it in the meantime

seehamrun:
Yea I ran into issues because I had made some changes to the old build script and forgot that there was the docker image script as well (sorry!) - I'm going to submit for now until we determine how to fix the build script (dont want to step on your toes).

PR #230, opened 2018-03-13
Start of demo server

jimmarino:


extensions/cloud/portability-cloud-local/src/main/java/org/dataportabilityproject/cloud/local/LocalJobStore.java
```diff
@@ -16,21 +16,30 @@
...
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
-import org.dataportabilityproject.spi.cloud.storage.JobStore;
-import org.dataportabilityproject.spi.cloud.types.JobAuthorization;
-import org.dataportabilityproject.spi.cloud.types.PortabilityJob;
 
 /** An in-memory {@link JobStore} implementation that uses a concurrent map as its store. */
 public final class LocalJobStore implements JobStore {
-  private final ConcurrentHashMap<UUID, Map<String, Object>> map;
+  private static ConcurrentHashMap<UUID, Map<String, Object>> SINGLETON_MAP;
```
rtannenbaum: As discussed on the call, I think you can assume that if this store is being used, we want the singleton

The other local impl uses GoogleJobStore backed by DataStore local emulator

I think we can get rid of the singleVM flag, and instead have LocalCloudExtension use the 'cloud' and 'env' flags to determine whether to return LocalJobStore or one of the emulator stores (I can do that along with #234)

portability-gateway/src/main/java/org/dataportabilityproject/gateway/ApiMain.java
```diff
@@ -41,16 +58,119 @@ public void uncaughtException(Thread thread, Throwable t) {
...
     // TODO: Don't hardcode list of services
     Set<String> services = ImmutableSet.of("microsoft");
 
+    // FIXME make configurable
+    SymmetricKeyGenerator keyGenerator = new AesSymmetricKeyGenerator();
+
     Injector injector =
         Guice.createInjector(
-            new PortabilityAuthServiceProviderModule(services), new ReferenceApiModule());
+            new ApiServicesModule(typeManager, cloudExtension.getJobStore(), keyGenerator),
```
rtannenbaum: this guice config looks off to me, but if it gets stuff working for the demo we can look into it more later

if you look at WorkerMain that is all configured to use Guice, you should only need to install the ApiModule and everything else should be automagically configured from there..

extensions/cloud/portability-cloud-local/src/main/java/org/dataportabilityproject/cloud/local/LocalCloudExtension.java
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright 2017 Google Inc.
```
seehamrun: nit: update this

jimmarino: It looks like there are a bunch of these. I've added an issue 241 for this.

extensions/cloud/portability-cloud-local/src/main/java/org/dataportabilityproject/cloud/local/LocalJobStore.java
jimmarino: I removed the singleVM flag. If you can re-work it to use the other flags in 234 that would be great.

portability-gateway/src/main/java/org/dataportabilityproject/gateway/ApiMain.java
jimmarino: I added the ApiServicesModule to make the instances created at bootstrap available for injection in core services. Is there something that doesn't seem right about that?

rtannenbaum: actually never mind, I forgot that JobStore etc isn't provided by Guice all the time, only for CloudExtensions who choose to use Guice. You can disregard this...

portability-gateway/src/main/java/org/dataportabilityproject/gateway/ApiMain.java
```diff
@@ -41,16 +58,119 @@ public void uncaughtException(Thread thread, Throwable t) {
...
+    ApiMain apiMain = new ApiMain();
+    apiMain.initialize();
+    apiMain.start();
+  }
+
+  public void initialize() {
+    TypeManager typeManager = new TypeManagerImpl();
+
+    // TODO implement
+    Map<String, Object> configuration = new HashMap<>();
```
rtannenbaum: FYI feel free to use the config parsing stuff from WorkerExtensionContext I added, if it'd be helpful

String tempSettings = "environment: LOCAL\ncloud: LOCAL";
InputStream in = new ByteArrayInputStream(tempSettings.getBytes(StandardCharsets.UTF_8));
config = ConfigUtils.parse(in);

jimmarino:
Yikes, I messed up the rebase. Let me revert and fix....

jimmarino:
OK fixed it.

Issue #232, opened 2018-03-13
Implement GoogleTransferExtension


rtannenbaum:
blocks #168

seehamrun:
This is completed in #240

PR #240, opened 2018-03-13
Implement GoogleTransferExtension

seehamrun:
Move MicrosoftTransferExtension to root of package
#232


extensions/data-transfer/portability-data-transfer-google/src/main/java/org/dataportabilityproject/datatransfer/google/GoogleTransferExtension.java
```diff
@@ -0,0 +1,69 @@
...
+  public void initialize(ExtensionContext context) {
+    // Note: initialize could be called twice in an account migration scenario where we import and
+    // export to the same service provider. So just return rather than throwing if called multiple
+    // times.
+    if(initialized) return;
+
+    jobStore = context.getService(JobStore.class);
+
+    ImmutableMap.Builder<String, Importer> importerBuilder = ImmutableMap.builder();
+    importerBuilder.put("contacts", new GoogleContactsImporter());
```
rtannenbaum: This is not strictly necessary but would be easy-ish and nice to do. Can be a TODO if you want:

We're using Guice for the Google CloudExtension and would be cool to do for our transfer extension as well. You'd need to annotate GoogleContactsImporter and Exporter's constructors with @Inject, and include a GoogleTransferExtensionModule in this directory, which would @Provide a PeopleService. (Actually I think the no-arg constructors you're calling will initialize the peopleService and calendarInterface's to null and not work, maybe we should get rid of those actually)

You can look in GoogleCloudExtension as an example, I think it'd be pretty easy but it's fine with me to leave as is too.

holachuy: Going to approve and merge for now and we can discuss and change if we want in a future PR

Issue #234, opened 2018-03-13
Implement secrets in new structure


rtannenbaum:
we should port over the remaining secrets loading code from portability-core to the new structure

this will probably entail local BucketStore managing the raw secrets from secrets.csv, and local CryptoKeyStore being a no-op as there's no decryption to do for the raw secrets, unlike the cloud ones which we store encrypted

blocks #192

PR #243, opened 2018-03-13
Create AppCredentialStore to manage secrets

rtannenbaum:
Rename BucketStore -> AppCredentialStore and remove the generic CryptoKeystore

Google's AppCredentialStore uses an internal crypto store to decrypt
its creds which are stored in its internal bucket Storage

Local AppCredentialStore will read creds from secrets.csv.

This ports CloudAppCredentialFactory from the old structure and leaves
secret management as an impl detail to AppCredentialStore rather than
imposing the BucketStore + CryptoKeystore approach we use for the Google
impl.

#234


extensions/cloud/portability-cloud-google/src/main/java/org/dataportabilityproject/cloud/google/GoogleAppCredentialStore.java
```diff
@@ -0,0 +1,148 @@
...
+  private final String bucketName;
+
+  /**
+   * Store keys and secrets in a cache so we can reduce load on cloud storage / KMS when reading
+   * keys and reading/decrypting secrets several times on startup.
+   *
+   * <p>Set the cache to reload keys/secrets periodically so that in the event of a key/secret being
+   * compromised, we can update them without restarting our servers.
+   */
+  private LoadingCache<String, String> keys;
```
holachuy: Both of these can be final.

extensions/cloud/portability-cloud-google/src/main/java/org/dataportabilityproject/cloud/google/GoogleAppCredentialStore.java
```diff
@@ -0,0 +1,148 @@
...
+            .setProjectId(projectId)
+            .setCredentials(googleCredentials)
+            .build()
+            .getService();
+    // Google Cloud Platform requires bucket names be unique across projects, so we include project
+    // ID in the bucket name.
+    this.bucketName = APP_CREDENTIAL_BUCKET_PREFIX + projectId;
+    this.keys =
+        CacheBuilder.newBuilder()
+            .expireAfterWrite(10, TimeUnit.MINUTES)
```
holachuy: Let's make these constants in the future

extensions/cloud/portability-cloud-google/src/main/java/org/dataportabilityproject/cloud/google/GoogleAppCredentialStore.java
```diff
@@ -0,0 +1,148 @@
...
+            .build(
+                new CacheLoader<String, String>() {
+                  @Override
+                  public String load(String key) throws Exception {
+                    return lookupKey(key);
+                  }
+                });
+    this.secrets =
+        CacheBuilder.newBuilder()
+            .expireAfterWrite(10, TimeUnit.MINUTES)
```
holachuy: Same here

holachuy:
Approving for now and we can address comments in future PR

PR #245, opened 2018-03-14
Google contacts move

olsona:


extensions/data-transfer/portability-data-transfer-google/src/main/java/org/dataportabilityproject/datatransfer/google/contacts/GoogleContactsImporter.java
```diff
@@ -29,9 +45,10 @@
 import static org.dataportabilityproject.datatransfer.google.common.GoogleStaticObjects.VCARD_PRIMARY_PREF;
 
 public class GoogleContactsImporter implements Importer<AuthData, ContactsModelWrapper> {
+
   private static final FieldMetadata PRIMARY_FIELD_METADATA = new FieldMetadata().setPrimary(true);
   private static final FieldMetadata SECONDARY_FIELD_METADATA =
-          new FieldMetadata().setPrimary(false);
+      new FieldMetadata().setPrimary(false);
```
holachuy: Let's make sure the formatting in this file is correct. I thought we had updated all files already to the new format.

Issue #237, opened 2018-03-13
Service load TransferExtension's


rtannenbaum:
blocks #208

PR #246, opened 2018-03-14
 Load TransferExtensions from ServiceLoader instead of hardcoding

seehamrun:
#237

### DIFF
--- a/portability-worker/build.gradle
+++ b/portability-worker/build.gradle
@@ -50,7 +50,6 @@ dependencies {
     compile project(':portability-spi-service')
     compile project(':portability-spi-cloud')
     compile project(':portability-spi-transfer')
-    compile project(':extensions:data-transfer:portability-data-transfer-microsoft')
     compile project(':libraries:security')
 
     compile("com.flickr4java:flickr4java:${flickrVersion}")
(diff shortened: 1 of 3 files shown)


portability-worker/build.gradle
```diff
@@ -50,7 +50,6 @@ dependencies {
     compile project(':portability-spi-service')
     compile project(':portability-spi-cloud')
     compile project(':portability-spi-transfer')
-    compile project(':extensions:data-transfer:portability-data-transfer-microsoft')
```
holachuy: Nice!

portability-worker/src/main/java/org/dataportabilityproject/worker/WorkerModule.java
```diff
@@ -45,6 +44,22 @@
     this.context = context;
   }
 
+  private static TransferExtension findTransferExtension(
```
rtannenbaum: nit: can you move this to the bottom of the file? java style guide has private methods at the bottom of classes

seehamrun: Weird, the reformatter keeps moving it up there - set it manually

portability-worker/build.gradle
seehamrun: Done.

PR #250, opened 2018-03-14
Bind auth instances in Guice context

jimmarino:
They were being bound as classes, causing Guice to re-instantiate them during injection


jimmarino:
Yeah. I'm not sure why. Once we do that, I won't need the fork anymore.

PR #257, opened 2018-03-15
Couple fixes for local demo

rtannenbaum:
Update URLs to be https for local demo as well as typo in DataTransferHandler


jimmarino:
That should be OK. It's just a self-signed certificate.

As Chey suggested, we should probably rename it in a separate PR to
something like "demo-selfsigned-keystore" or similar to denote it is for
testing.

Jim

rtannenbaum:
Will add the jks file too

PR #263, opened 2018-03-16
Add back URL encoding in the Encrypter/Decrypter

rtannenbaum:


distributions/demo-server/resources/config/api.yaml
```diff
@@ -1,2 +1,2 @@
-baseUrl: https://localhost:3000
-baseApiUrl: https://localhost:8080
+baseUrl: http://localhost:3000
```
seehamrun: Why are we undoing this?

rtannenbaum: didn't mean to push that, was testing without https. i reverted it.

Issue #258, opened 2018-03-16
Tracker for temporary changes to get local demo working


rtannenbaum:
We are temporarily disabling a number of features for our local demo. Tagging them all with this issue to keep track.

rtannenbaum:
I think just XSRF is left

seehamrun:
re-enabled xsrf with #277

PR #266, opened 2018-03-16
Temporarily disable state token job ID check for local demo - #258

rtannenbaum:


portability-gateway/src/main/java/org/dataportabilityproject/gateway/reference/Oauth2CallbackHandler.java
```diff
@@ -118,14 +118,15 @@ private String handleExchange(HttpExchange exchange) throws IOException {
...
       UUID jobId = ReferenceApiUtils.decodeJobId(encodedIdCookie.getValue());
-      UUID jobIdFromState = ReferenceApiUtils.decodeJobId(authResponse.getState());
-
-      // TODO: Remove sanity check
-      Preconditions.checkState(
-          jobIdFromState.equals(jobId),
-          "Job id in cookie [%s] and request [%s] should match",
-          jobId,
-          jobIdFromState);
+      // TODO(#258): Check job ID in state token, was broken during local demo
```
holachuy: Could we log it so we know what's in there in the mean time?

seehamrun:
Tagging #258  explicitly to this PR

PR #267, opened 2018-03-16
Re-enable encryption using double encryption and encrypting the secon…

holachuy:
Issue #258 

Re-enable encryption using double encryption and encrypting the secod key with the public key of the worker

### DIFF
--- a/libraries/security/src/main/java/org/dataportabilityproject/security/DecrypterImpl.java
+++ b/libraries/security/src/main/java/org/dataportabilityproject/security/DecrypterImpl.java
@@ -44,30 +44,24 @@ final class DecrypterImpl implements Decrypter {
 
   @Override
   public String decrypt(String encrypted) {
-    byte[] data = BaseEncoding.base64Url().decode(encrypted);
-    return new String(data, Charsets.UTF_8);
-    //   System.arraycopy(decrypted, 8, data, 0, data.length);
-    //   return new String(data, Charsets.UTF_8);
-    // TODO(#258): Encryption temporarily disabled to get local demo working. Issue is:
-    // https://stackoverflow.com/questions/10007147/getting-a-illegalblocksizeexception-data-must-not-be-longer-than-256-bytes-when
-    // try {
-    //   byte[] decoded = BaseEncoding.base64Url().decode(encrypted);
-    //   Cipher cipher = Cipher.getInstance(transformation);
-    //   cipher.init(Cipher.DECRYPT_MODE, key);
-    //   byte[] decrypted = cipher.doFinal(decoded);
-    //   if (decrypted == null || decrypted.length <= 8) {
-    //     throw new RuntimeException("incorrect decrypted text.");
-    //   }
-    //   byte[] data = new byte[decrypted.length - 8];
-    //   System.arraycopy(decrypted, 8, data, 0, data.length);
-    //   return new String(data, Charsets.UTF_8);
-    // } catch (BadPaddingException
-    //     | IllegalBlockSizeException
-    //     | InvalidKeyException
-    //     | NoSuchAlgorithmException
-    //     | NoSuchPaddingException e) {
-    //   logger.error("Error decrypting data, length: {}", encrypted.length(), e);
-    //   throw new RuntimeException(e);
-    // }
+    try {
+      byte[] decoded = BaseEncoding.base64Url().decode(encrypted);
+      Cipher cipher = Cipher.getInstance(transformation);
+      cipher.init(Cipher.DECRYPT_MODE, key);
+      byte[] decrypted = cipher.doFinal(decoded);
+      if (decrypted == null || decrypted.length <= 8) {
+        throw new RuntimeException("incorrect decrypted text.");
+      }
+      byte[] data = new byte[decrypted.length - 8];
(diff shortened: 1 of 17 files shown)


holachuy:
PTAL, this is ready for review.

This runs locally and starts the transfer. The log from my local machine is here: https://gist.github.com/holachuy/f2cd10f682104084edfd715bb5cce67f

portability-spi-cloud/src/main/java/org/dataportabilityproject/spi/cloud/types/JobAuthorization.java
```diff
@@ -83,11 +87,14 @@ private static Builder create() {
     @JsonProperty("encryptedImportAuthData")
     public abstract Builder setEncryptedImportAuthData(String authData);
 
+    @JsonProperty("encryptedSymmetricKeyForAuthData")
```
rtannenbaum: The naming of the 3 keys is getting confusing to me.. I think we should either consistently name them according to type (symmetric vs asymmetric) or purpose (session vs worker), or both

what about one of these conventions:
- sessionPrimarySymmetricKey, sessionSecondarySymmetricKey, workerPublicKey
- sessionSymmetricKeyA, sessionSymmetricKeyB, workerPublicKey

and I like the 'forAuthData' but it makes the name pretty long, maybe it could go in the names like this:
- sessionSymmetricAuthDataKeyA, sessionSymmetricAuthDataKeyB, workerPublicAuthDataKey

or don't include it in the field names but have it in a comment instead, "these three keys are all used for encrypting auth data"

portability-gateway/src/main/java/org/dataportabilityproject/gateway/action/startjob/StartJobAction.java
```diff
@@ -179,24 +185,41 @@ private void encryptAndUpdateJobWithCredentials(
...
     // Encrypt the data with the assigned workers PublicKey and persist
-    // NOTE: These auth datas are already encrypted with the session symmetric encryption key
-    Encrypter crypter = EncrypterFactory.create(publicKey);
-    String encryptedExportAuthData = crypter.encrypt(encryptedExportAuthCredential);
-    logger.debug("Encrypted export auth data, has length: {}", encryptedExportAuthData.length());
-    String encryptedImportAuthData = crypter.encrypt(encryptedImportAuthCredential);
-    logger.debug("Encrypted import auth data, has length: {}", encryptedImportAuthData.length());
+
+    /*
+     * NOTE: These auth datas are already encrypted with the session symmetric encryption key
```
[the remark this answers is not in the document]
holachuy: I removed wordy explanation, hopefully this is clearer.

Background:
I made the mistake of first attempting to encrypt all the data with RSA/asymmetric key. This is wrong because RSA is mean to only encrypt a fixed amount of information, not arbitrarily long data.

The general way to do this is to encrypt the arbitrarily long data (in our case, auth data) with a symmetric key then encrypt the symmetric key itself with RSA public key since it's a smaller length.  That way only the private key can unpack the symmetric key to decrypt the data.

portability-spi-cloud/src/main/java/org/dataportabilityproject/spi/cloud/types/JobAuthorization.java
holachuy: Ok, renamed to a simpler approach. I think it's wroth documenting and adding a link.

holachuy:
@rtannenbaum  if you click show outdated here, you should see my responses https://github.com/google/data-transfer-project/pull/267#pullrequestreview-104771747

libraries/security/src/main/java/org/dataportabilityproject/security/DecrypterImpl.java
```diff
@@ -44,30 +44,24 @@
...
-    //   return new String(data, Charsets.UTF_8);
-    // } catch (BadPaddingException
-    //     | IllegalBlockSizeException
-    //     | InvalidKeyException
-    //     | NoSuchAlgorithmException
-    //     | NoSuchPaddingException e) {
-    //   logger.error("Error decrypting data, length: {}", encrypted.length(), e);
-    //   throw new RuntimeException(e);
-    // }
+    try {
```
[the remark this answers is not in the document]
rtannenbaum: thanks, looks like that worked

portability-spi-cloud/src/main/java/org/dataportabilityproject/spi/cloud/types/JobAuthorization.java
```diff
@@ -37,12 +37,22 @@ public static Builder builder() {
   @JsonProperty("encryptedImportAuthData")
   public abstract String encryptedImportAuthData();
 
-  @JsonProperty("encodedSessionKey")
-  public abstract String encodedSessionKey();
+  /** The SecretKey used to encrypt all data associated with this job, encoded for storage. */
```
rtannenbaum: does this mean it's used for other data besides auth data? might be worth clarifying that

also OCD nit, could it be called sessionSecretKey to be consistent with the secret vs public in the other 2 keys?

holachuy: Both done.

PR #277, opened 2018-03-19
 re-enable xsrf protection

seehamrun:
#258 (reverts  #266) 

this requires that a JWT_SECRET and JWT_KEY  be provided from the AppCredentialStore.

For local deployment add the following into your ~/.gradle/gradle.properties
```
data.portability.key.jwt=foo
data.portability.secret.jwt=bar
```
For cloud deployment, make sure a file/blob for JWT_KEY and JWT_SECRET exist where the key can contain foo, and the secret should be generated via ssh_keygen or something similar.


rtannenbaum:
Tagging #234 as well

Issue #276, opened 2018-03-19
Wire up auth in Google contacts transfer extensions


rtannenbaum:

holachuy:
@jimmarino  took care of this in #280

PR #280, opened 2018-03-20
Demo server fixes

jimmarino:
This fixes a bunch of issues to get the end-to-end demo of contacts transfer working.

### DIFF
--- a/portability-gateway/src/main/java/org/dataportabilityproject/gateway/action/startjob/StartJobAction.java
+++ b/portability-gateway/src/main/java/org/dataportabilityproject/gateway/action/startjob/StartJobAction.java
@@ -16,15 +16,10 @@
 package org.dataportabilityproject.gateway.action.startjob;
 
 import com.google.api.client.util.Sleeper;
-import com.google.common.base.Charsets;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import com.google.common.io.BaseEncoding;
 import com.google.inject.Inject;
-import java.io.IOException;
-import java.security.PublicKey;
-import java.util.UUID;
-import javax.crypto.SecretKey;
 import org.dataportabilityproject.gateway.action.Action;
 import org.dataportabilityproject.gateway.action.ActionUtils;
 import org.dataportabilityproject.security.AsymmetricKeyGenerator;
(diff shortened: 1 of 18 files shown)


portability-worker/src/main/java/org/dataportabilityproject/worker/JobMetadata.java
```diff
@@ -56,6 +56,14 @@ static void init(
     importService = initImportService;
   }
 
+  static void reset() {
```
holachuy: If we keep this around, let's synchronize on this

portability-gateway/src/main/java/org/dataportabilityproject/gateway/action/startjob/StartJobAction.java
```diff
@@ -200,8 +200,9 @@ private void encryptAndUpdateJobWithCredentials(
         asymmetricKeyGenerator.parse(
             BaseEncoding.base64Url().decode(job.jobAuthorization().authPublicKey()));
     Encrypter asymmetricEncrypter = EncrypterFactory.create(authPublicKey);
+
     String encryptedAuthSecretKey =
-        asymmetricEncrypter.encrypt(new String(authSecretKey.getEncoded(), Charsets.UTF_8));
+        asymmetricEncrypter.encrypt(BaseEncoding.base64Url().encode(authSecretKey.getEncoded()));
```
holachuy: This was it! I notice it's in the Decrypter, this may make sense in the Encrypter, I'll follow up after this PR

portability-worker/src/main/java/org/dataportabilityproject/worker/JobMetadata.java
jimmarino: Good point, I think we should get rid of this, let me put a TODO in there.

portability-gateway/src/main/java/org/dataportabilityproject/gateway/action/startjob/StartJobAction.java
jimmarino: OK

rtannenbaum: Good find! Do we need the extra encoding? Is authSecretKey already encoded ("getEncoded()")? Can we remove it on both ends?

portability-worker/src/main/java/org/dataportabilityproject/worker/JobMetadata.java
```diff
@@ -56,6 +56,15 @@ static void init(
     importService = initImportService;
   }
 
+  // TODO: remove this
+  static synchronized void reset() {
```
rtannenbaum: Can we assert this is only called when in local-demo environment?

holachuy:
Also took care of issue #276

Issue #197, opened 2018-03-09
Parse flags in ExtensionContext


rtannenbaum:
blocks #171 and #172

PR #287, opened 2018-03-22
Get configuration from Settings extensions, with a YAML implementation

rtannenbaum:
- Create SettingsExtension and ApiSettingsExtension
- Include YAML implementation in extensions/config/portability-config-yaml
- API and worker servers service load settings extension and pass to their ExtensionContext's
- Use YAML extension in demo-server. Tested locally.

#197


libraries/config/src/main/java/org/dataportabilityproject/config/CommonSettings.java
```diff
@@ -0,0 +1,122 @@
...
+      logger.debug("Parsed flags: {}", tempCommonSettings);
+      commonSettings = tempCommonSettings;
+    } catch (IOException e) {
+      throw new IllegalArgumentException("Problem parsing common settings", e);
+    }
+    return commonSettings;
+  }
+
+  public String getCloud() {
+    return cloud;
```
holachuy: do we need an intialized or null check in case this is called before getCommonSettings?

libraries/config/src/main/java/org/dataportabilityproject/config/CommonSettings.java
```diff
@@ -0,0 +1,122 @@
...
+   */
+  private static CommonSettings commonSettings;
+
+  private static final String COMMON_SETTINGS_PATH = "config/common.yaml";
+  private static final String ENV_COMMON_SETTINGS_PATH = "config/env/common.yaml";
+
+  private final String cloud;
+  private final Environment environment;
+
+  public CommonSettings() {
```
holachuy: Should this be private or non existent?

libraries/config/src/main/java/org/dataportabilityproject/config/ConfigUtils.java
```diff
@@ -33,6 +34,9 @@
    * Parses an input stream to a yaml configuration file into a generic Map<String, Object>.
    */
   public static Map<String, Object> parse(InputStream in) throws IOException {
+    if (in == null) {
+      return new HashMap<>();
```
holachuy: warning log?

portability-gateway/src/main/java/org/dataportabilityproject/gateway/reference/ApiSettingsModule.java
```diff
@@ -0,0 +1,76 @@
...
+
+  @Override
+  public void configure() {}
+
+  @Provides
+  @Singleton
+  ApiSettings getApiSettings() {
+    // Any setting in both a base and env config will be overridden by the env definition.
+    // This is enforced by the ordering of the settings files in the list below.
+    try {
```
holachuy: Should we move the initialization logic for CommonSettings into a provider method similar to this? This seems a bit cleaner.

libraries/config/src/main/java/org/dataportabilityproject/config/CommonSettings.java
rtannenbaum: Done

libraries/config/src/main/java/org/dataportabilityproject/config/CommonSettings.java
rtannenbaum: this one isn't static so it would be called on an instance. I just removed the null-init ctor so I think it should be ok now.

libraries/config/src/main/java/org/dataportabilityproject/config/ConfigUtils.java
rtannenbaum: Done

portability-gateway/src/main/java/org/dataportabilityproject/gateway/reference/ApiSettingsModule.java
rtannenbaum: The issue with that is extensions that don't use Guice would have to get it from the module then, which seemed bad. So I have CommonSettings responsible for its own parsing then expose it in the module, so it can either be provided from the module or called statically from CommonSettings for extensions that don't use Guice. ApiSettings is different because none of the extensions are using it (yet), so it can be done entirely in Guice.

jimmarino:
Looks nice, I just had had a few minor questions and suggestions. 

Could ExtensionContext.getConfiguration(String key, T defaultValue) be the way for non-Guice extensions to obtain "normalized"configuration rather than a static singleton? I think Chuy made a similar comment if I understand correctly.

Related to this, could the sourcing of configuration be moved up to a library that is used by the individual runtime bootstrappers? For example, allow the runtime boostrappers to decide where to obtain configuration from?

It would probably be a good thing to keep the option open for runtimes to source config from places other than files on the classpath. For example, if we create a Gradle-based test runtime, the config could be specified in the build files directly. Some cloud deployments may want to pull stuff from a store such as etcd or Consul. By having the config non-static, it also avoids any issues with having to cleanup values between test runs as well. 

I'm around tomorrow to discuss in more detail if people want.

rtannenbaum:
Hey Jim, Thanks for those suggestions. That’s a good idea to allow extensions to pull configuration from other places than the class path, like etcd.

I was planning to have extensions retrieve their own custom settings from ExtensionContext.getConfiguration. Then I guess, those could be parsed either from yaml in class path or some other place like etcd that the extension would specify.

For the ‘known’/required settings though, I like having the structure of ApiSettings and CommonSettings which is automatically validated when parsed, and doesn’t allow for typos or type mismatches when reading those known settings. Do you still prefer to have those known settings (e.g. cloud, environment) obtained from ExtensionContext.getConfigurtion()?

That’s a good point about statics causing issues for tests. We could certainly make it non-static.

jimmarino:
Ah OK I see. +1 on typed common settings. Could we get those from
ExtensionContext so they could be sourced from non-YAML locations? We could
even have methods on extension context to get the settings directly so they
are more discoverable.



On Fri, Mar 23, 2018 at 3:30 AM, Rachel Tannenbaum <notifications@github.com

rtannenbaum:
Hey @jimmarino , let me know what you think of this impl, thanks! I updated the PR description/comment to reflect the changes.

jimmarino:
FYI JCenter (dependency download) was having trouble today so that may be why the build is failing.

rtannenbaum:
Thanks. Looks like a bad merge actually, let me see if I can fix it.

rtannenbaum:
hey @holachuy can you take another look when you get a chance? Thanks!

portability-api-launcher/src/main/java/org/dataportabilityproject/api/launcher/extension/SettingsExtension.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+package org.dataportabilityproject.api.launcher.extension;
+
+import org.dataportabilityproject.api.launcher.AbstractExtension;
+import org.dataportabilityproject.api.launcher.CommonSettings;
+
+/**
+ * Extension that provides {@link CommonSettings} and extension-specific settings.
+ *
+ * <p>Used by both the worker server, and the API server (by inheritance in
```
holachuy: This is a bit confusing.

portability-api-launcher/src/main/java/org/dataportabilityproject/api/launcher/ExtensionContext.java
```diff
@@ -43,11 +43,13 @@
...
   /**
-   * Returns the configuration parameter for the key or the default value if not found.
+   * Returns the configuration value for {@code setting} or the default value if not found.
    *
-   * @param key the parameter key
+   * @param setting the parameter key
    * @param defaultValue the default value. Null may be passed.
    */
-  <T> T getConfiguration(String key, T defaultValue);
+  <T> T getSetting(String setting, T defaultValue);
```
holachuy: It this just  pass through to settingsExtension each time? If so, do we need to specify a method or constructor that requires (SettingsExtension) method of some type?

holachuy: When will a developer use this vs getConfiguration?

rtannenbaum:
New PR: https://github.com/google/data-transfer-project/pull/304/files

PR #291, opened 2018-03-23
Quick fix on result type in Google Photos Exporter

olsona:


extensions/data-transfer/portability-data-transfer-google/src/main/java/org/dataportabilityproject/datatransfer/google/photos/GooglePhotosExporter.java
```diff
@@ -122,6 +122,9 @@ public GooglePhotosExporter() {
       }
 
       ResultType resultType = ResultType.CONTINUE;
+      if (nextPageData == null || continuationData.getContainerResources().isEmpty()) {
```
bwillard: I have a question here, if there is container should the result type be Continue or end?

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/dataportabilityproject/datatransfer/google/photos/GooglePhotosExporter.java
```diff
@@ -122,6 +122,9 @@ public GooglePhotosExporter() {
...
+        resultType = ResultType.END;
```
bwillard: It might be cleaner to do this in the if block on line 109.

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/dataportabilityproject/datatransfer/google/photos/GooglePhotosExporter.java
```diff
@@ -159,7 +162,12 @@ public GooglePhotosExporter() {
       }
 
       PhotosContainerResource containerResource = new PhotosContainerResource(null, photos);
-      return new ExportResult<>(ResultType.END, containerResource, continuationData);
+
+      ResultType resultType = ResultType.CONTINUE;
+      if (nextPageData == null) {
```
bwillard: Here too, it might be cleaner to put this higher up in the if block too.

PR #293, opened 2018-03-26
Flickr auth and wireup through the binaries 

seehamrun:
#213 

This also adds InitialAuthData to the JobMap since its required for oauth1 
Also fix PortabilityCopier to use JobId instead of hardcoded number (needed for any importer/exporters that use the job store).


portability-worker/src/main/java/org/dataportabilityproject/worker/PortabilityInMemoryDataCopier.java
```diff
@@ -87,7 +87,7 @@ private void copyHelper(
 
     logger.debug("Starting import");
     // TODO, use job Id?
-    importer.get().importItem("1", importAuthData, exportResult.getExportedData());
```
holachuy: Wow, great catch!

PR #296, opened 2018-03-27
Add in Azure Vault for secrets

jimmarino:
Work in progress on getting to run in Azure...

### DIFF
--- a/extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureKeyVaultStore.java
+++ b/extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureKeyVaultStore.java
@@ -12,11 +12,7 @@ import java.util.Objects;
 
 import static java.lang.System.getenv;
 
-/**
- * A secrets store backed by Azure Key Vault.
- *
- * <p>
- */
+/** A secrets store backed by Azure Key Vault. */
 public class AzureKeyVaultStore implements AppCredentialStore {
   private static final String VAULT_ADDRESS = "https://%s.vault.azure.net/";
 
@@ -26,6 +22,7 @@ public class AzureKeyVaultStore implements AppCredentialStore {
   public AzureKeyVaultStore(String keyVaultName) {
     vaultUrl = String.format(VAULT_ADDRESS, keyVaultName);
 
+    // TODO should env vars be sourced from config?
     String azureClientId = getenv("AZURE_CLIENT_ID");
     Objects.requireNonNull(azureClientId, "AZURE_CLIENT_ID not set in environment");
 
--- a/extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureKeyVaultStore.java
+++ b/extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureKeyVaultStore.java
@@ -1,0 +1,57 @@
+package org.dataportabilityproject.cloud.microsoft.cosmos;
+
+import com.microsoft.azure.AzureEnvironment;
+import com.microsoft.azure.credentials.ApplicationTokenCredentials;
+import com.microsoft.azure.keyvault.KeyVaultClient;
+import com.microsoft.azure.keyvault.models.SecretBundle;
+import org.dataportabilityproject.spi.cloud.storage.AppCredentialStore;
+import org.dataportabilityproject.types.transfer.auth.AppCredentials;
+
+import java.io.IOException;
+import java.util.Objects;
+
+import static java.lang.System.getenv;
+
+/**
+ * A secrets store backed by Azure Key Vault.
+ *
+ * <p>
+ */
+public class AzureKeyVaultStore implements AppCredentialStore {
+  private static final String VAULT_ADDRESS = "https://%s.vault.azure.net/";
+
+  private final String vaultUrl;
+  private final KeyVaultClient vaultClient;
+
+  public AzureKeyVaultStore(String keyVaultName) {
+    vaultUrl = String.format(VAULT_ADDRESS, keyVaultName);
+
+    String azureClientId = getenv("AZURE_CLIENT_ID");
+    Objects.requireNonNull(azureClientId, "AZURE_CLIENT_ID not set in environment");
+
+    String azureTenantId = getenv("AZURE_TENANT_ID");
+    Objects.requireNonNull(azureTenantId, "AZURE_TENANT_ID not set in environment");
+
+    String azureClientSecret = getenv("AZURE_CLIENT_SECRET");
+    Objects.requireNonNull(azureClientSecret, "AZURE_CLIENT_SECRET not set in environment");
+
+    ApplicationTokenCredentials credentials =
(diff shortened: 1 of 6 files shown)


extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureKeyVaultStore.java
```diff
@@ -0,0 +1,57 @@
...
+    vaultUrl = String.format(VAULT_ADDRESS, keyVaultName);
+
+    String azureClientId = getenv("AZURE_CLIENT_ID");
+    Objects.requireNonNull(azureClientId, "AZURE_CLIENT_ID not set in environment");
+
+    String azureTenantId = getenv("AZURE_TENANT_ID");
+    Objects.requireNonNull(azureTenantId, "AZURE_TENANT_ID not set in environment");
+
+    String azureClientSecret = getenv("AZURE_CLIENT_SECRET");
+    Objects.requireNonNull(azureClientSecret, "AZURE_CLIENT_SECRET not set in environment");
```
rtannenbaum: What is this client secret used for? Is this a credential to access Azure services? Do you plan on moving to a more secure mechanism for storing this? For the GCP impl we store "service account" credentials as a Kubernetes secret. if you are using Kubernetes you could probably do that too.

If Azure supports it, I think role accounts are more secure than storing creds.

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureKeyVaultStore.java
```diff
@@ -0,0 +1,57 @@
...
+
+    ApplicationTokenCredentials credentials =
+        new ApplicationTokenCredentials(
+            azureClientId, azureTenantId, azureClientSecret, AzureEnvironment.AZURE);
+    vaultClient = new KeyVaultClient(credentials);
+  }
+
+  @Override
+  public AppCredentials getAppCredentials(String keyName, String secretName) throws IOException {
+    SecretBundle secretBundle = vaultClient.getSecret(vaultUrl, "/secrets/" + keyName);
```
rtannenbaum: Nice. Out of curiosity, are these stored encrypted?

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/MicrosoftCloudExtension.java
```diff
@@ -0,0 +1,61 @@
...
+    // create the vault
+    vaultStore = new AzureKeyVaultStore(vaultName);
+
+    // create the connection to Cosmos DB
+    ObjectMapper mapper = context.getTypeManager().getMapper();
+
+    String username = vaultStore.getSecret(JOB_STORE_USER_NAME);
+    Objects.requireNonNull(username, "JobStore username not set in vault");
+
+    String password = vaultStore.getSecret(JOB_STORE_PASSWORD);
```
rtannenbaum: Same comment here about storing credentials

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureKeyVaultStore.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.io.IOException;
+import java.util.Objects;
+
+import static java.lang.System.getenv;
+
+/**
+ * A secrets store backed by Azure Key Vault.
+ *
+ * <p>
```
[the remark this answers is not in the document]
jimmarino: The reason I put that in there is the Google formatter converts the class Javadoc into a single line comment which is really weird for class-level Javadoc. Do you know how to turn that off?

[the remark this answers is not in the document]
rtannenbaum: Oh weird. I'm used to seeing single line comments if it all fits on one line, even for classes, but didn't know the formatter enforces that. Not sure how to turn that off. I wouldn't mind single line personally.

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureKeyVaultStore.java
```diff
@@ -0,0 +1,57 @@
...
+public class AzureKeyVaultStore implements AppCredentialStore {
+  private static final String VAULT_ADDRESS = "https://%s.vault.azure.net/";
+
+  private final String vaultUrl;
+  private final KeyVaultClient vaultClient;
+
+  public AzureKeyVaultStore(String keyVaultName) {
+    vaultUrl = String.format(VAULT_ADDRESS, keyVaultName);
+
+    String azureClientId = getenv("AZURE_CLIENT_ID");
```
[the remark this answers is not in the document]
jimmarino: They don't but my thinking was they would be supplied by whatever controls the process since they are sensitive. For example, they could be a Kubernetes secret. We could have it sourced from props and override to env but that seems potentially dangerous. What do you think?

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureKeyVaultStore.java
jimmarino: Yeah the env variable would be provided by whatever controls the process, e.g. a Kubernetes would source it from a secret (secrets can be set as env variables by Kubernetes)

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureKeyVaultStore.java
jimmarino: In Azure they are

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureKeyVaultStore.java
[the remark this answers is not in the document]
rtannenbaum: OK cool, that sounds good. Maybe just add a comment reflecting that this should come from somewhere secure like a Kubernetes secret, so someone writing a distribution knows to do that and not set a raw cred in as an env variable in the container, for example.

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureKeyVaultStore.java
rtannenbaum: Nice, I'll have to see if GCP has something like this, we do the encryption manually

Issue #289, opened 2018-03-23
Limit Oauth Scopes requested in Google Services


seehamrun:
limit them to just the data type being used and read/write permissions only if importing

PR #297, opened 2018-03-27
limit scopes requested in Google Auth

seehamrun:
#289

### DIFF
--- a/extensions/auth/portability-auth-google/src/main/java/org/dataportabilityproject/auth/google/GoogleAuthDataGenerator.java
+++ b/extensions/auth/portability-auth-google/src/main/java/org/dataportabilityproject/auth/google/GoogleAuthDataGenerator.java
@@ -27,12 +27,10 @@ import com.google.api.services.people.v1.PeopleServiceScopes;
 import com.google.api.services.tasks.TasksScopes;
 import com.google.common.base.Charsets;
 import com.google.common.base.Preconditions;
-import com.google.common.collect.ImmutableListMultimap;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.ListMultimap;
 import com.google.common.io.BaseEncoding;
 import java.io.IOException;
-import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import org.dataportabilityproject.spi.gateway.auth.AuthDataGenerator;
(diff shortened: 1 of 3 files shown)


extensions/auth/portability-auth-google/src/main/java/org/dataportabilityproject/auth/google/GoogleAuthDataGenerator.java
```diff
@@ -44,10 +50,38 @@
  * authorization code and posts it against the AD API to obtain a token for querying the Graph API.
  */
 public class GoogleAuthDataGenerator implements AuthDataGenerator {
-  // TODO: Reduce requested scopes by service and authorization mode (readwrite/read)
-  private static final ImmutableCollection<String> SCOPES =
-      ImmutableSet.of(CalendarScopes.CALENDAR, PeopleServiceScopes.CONTACTS, TasksScopes.TASKS);
+  // The scopes necessary to import or export each supported data type.
+  // Scopes for EXPORT should contain READONLY permissions
+  private static final Map<String, ListMultimap<AuthMode, String>> DATA_TYPE_SCOPES =
```
holachuy: Very nice! Since there's so many levels to this map, can we have two map constants, EXPORT_DATA_SCOPE_TYPES and IMPORT_DATA_SCOPE TYPES?

extensions/auth/portability-auth-google/src/main/java/org/dataportabilityproject/auth/google/GoogleAuthDataGenerator.java
```diff
@@ -44,10 +50,38 @@
...
+      ImmutableMap.<String, ListMultimap<AuthMode, String>>builder()
+          .put(
+              "calendar",
+              ImmutableListMultimap.<AuthMode, String>builder()
+                  .putAll(AuthMode.IMPORT, Arrays.asList(CalendarScopes.CALENDAR))
```
holachuy: I think if you call .put() you won't need the Arrays.asList()

extensions/auth/portability-auth-google/src/main/java/org/dataportabilityproject/auth/google/GoogleAuthDataGenerator.java
seehamrun: Done

extensions/auth/portability-auth-google/src/main/java/org/dataportabilityproject/auth/google/GoogleAuthDataGenerator.java
seehamrun: Done. I used ImmutableList.of() instead.

Issue #290, opened 2018-03-23
Limit Oauth Scopes requested in Microsoft Services


seehamrun:
limit them to just the data type being used and read/write permissions only if importing

PR #299, opened 2018-03-27
limit oauth scopes requested in microsoft

seehamrun:
#290

### DIFF
--- a/extensions/auth/portability-auth-microsoft/src/main/java/org/dataportabilityproject/auth/microsoft/MicrosoftAuthDataGenerator.java
+++ b/extensions/auth/portability-auth-microsoft/src/main/java/org/dataportabilityproject/auth/microsoft/MicrosoftAuthDataGenerator.java
@@ -1,6 +1,8 @@
 package org.dataportabilityproject.auth.microsoft;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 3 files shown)


extensions/auth/portability-auth-microsoft/src/main/java/org/dataportabilityproject/auth/microsoft/MicrosoftAuthDataGenerator.java
```diff
@@ -45,26 +68,34 @@
...
+      ObjectMapper mapper,
+      String transferDataType,
+      AuthMode mode) {
     this.redirectPath = redirectPath;
     this.clientIdSupplier = clientIdSupplier;
     this.clientSecretSupplier = clientSecretSupplier;
     httpClient = client;
     this.mapper = mapper;
+    this.scopes =
+        mode == AuthMode.EXPORT
```
holachuy: Do we have to throw an error of it ends up null or do a precondtion check?

seehamrun: Done - Presumably this shouldnt ever happen since only the AuthServiceExtension creates these, but added it as a catch.

PR #301, opened 2018-03-28
Parse configuration from SettingsExtension, with a YAML implementation

rtannenbaum:
- Create SettingsExtension, ApiSettingsExtension, WorkerSettingsExtension which are used by ExtensionContext's. ExtensionContext's then expose typed and generic settings to extensions.
- Create Guice flag binder module which automatically binds any flags in SettingsExtensions as `@Named` annotations for use in core classes.
- Add YAML implementation of ApiSettingsExtension and WorkerSettingsExtension, and use it in demo-server.

Tested `@Named` annotations are bound correctly when running demo-server.

#197


rtannenbaum:
New PR: https://github.com/google/data-transfer-project/pull/304/files

PR #304, opened 2018-03-28
Parse configuration from SettingsExtension, with a YAML implementation

rtannenbaum:
- Add SettingsExtension which just exposes the generic getSetting(name, default) method
- Add a YamlSettingsExtension implementation that parses from YAML files into a Map<String, Object>. Use this in demo-server.
- A future EtcdSettingsExtension could fetch settings from etcd each time (to get dynamic updates)
- Add required typed settings to ApiExtensionContext and WorkerExtensionContext, with validation, in addition to the generic getSetting(). Extensions may grab settings from these methods.
- ReferenceApiModule and WorkerModule bind all required flags from their respective ExtensionContext's via FlagBindingModule and the `@Flag` annotation. New flags added to either context will automatically be bound as `@Named` annotations. This exposes all required settings via Guice for the core classes (not extensions).
- Remove old YAML validation. Validation of required flags is now done in the ExtensionContext's.

Tested `@Named` annotations are bound correctly when running demo-server.

#197


portability-worker/src/main/java/org/dataportabilityproject/worker/WorkerModule.java
```diff
@@ -61,6 +58,9 @@
 
   @Override
   protected void configure() {
+    // binds flags from ExtensionContext to @Named annotations
+    super.configure();
```
holachuy: What if we did FlagBindingModule.parse(context) instead of in the construtor

rtannenbaum: Done (I think - let me know if this isn't what you meant)

rtannenbaum:
Thanks for all the reviews & making this simpler!

Issue #191, opened 2018-03-08
Remove PortabilityJob.to/fromMap in favor of JSON serialization


rtannenbaum:
cc @jimmarino as discussed on Slack and @holachuy for the discussion on #187

rtannenbaum:
This came up in #308, it should probably be up to the JobStore implementation how it wants to de/serialize PortabilityJob. See the Azure store for an example. We should tweak the Google & local impls to do the same and then remove the map methods.

seehamrun:
Came up again in #670 we should really fix this and switch to JSON :)

lisad:
We reviewed this issue in maintainers' meeting as we go through old issues - we believe this could be fixed but is low-priority so it's fine to just keep in the issue list.

PR #308, opened 2018-03-30
Microsoft storage extension updates

jimmarino:
Fully implements storage and secrets handling using Azure services.


rtannenbaum:
tagging #275

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureTableStore.java
```diff
@@ -0,0 +1,285 @@
...
+import java.io.InputStream;
+import java.net.URISyntaxException;
+import java.security.InvalidKeyException;
+import java.util.Iterator;
+import java.util.UUID;
+
+import static com.microsoft.azure.storage.table.TableQuery.generateFilterCondition;
+
+/** Uses the Azure Cosmos DB Table Storage API to persist job data. */
+public class AzureTableStore implements JobStore {
```
rtannenbaum: this replaces CosmosStore, right? Anyway to make github realize it's a rename?

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureTableStore.java
```diff
@@ -0,0 +1,285 @@
...
+
+  @Override
+  public void updateJob(UUID jobId, PortabilityJob job) throws IOException {
+    updateJob(jobId, job, null);
+  }
+
+  @Override
+  public void updateJob(UUID jobId, PortabilityJob job, JobUpdateValidator validator)
+      throws IOException {
+    Preconditions.checkNotNull(jobId, "Job not persisted");
```
rtannenbaum: nit: I think this debugging message is wrong (for jobId null check)

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureTableStore.java
```diff
@@ -0,0 +1,285 @@
...
+    Preconditions.checkNotNull(job, "Job is null");
+    try {
+
+      CloudTable table = tableClient.getTableReference(JOB_TABLE);
+
+      String serializedJob = configuration.getMapper().writeValueAsString(job);
+      DataWrapper wrapper =
```
rtannenbaum: passing the jobID as the "rowKey" here makes jobID the key, right? Would be useful to add a comment about this

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureTableStore.java
```diff
@@ -0,0 +1,285 @@
...
+  @Override
+  public void updateJob(UUID jobId, PortabilityJob job, JobUpdateValidator validator)
+      throws IOException {
+    Preconditions.checkNotNull(jobId, "Job not persisted");
+    Preconditions.checkNotNull(job, "Job is null");
+    try {
+
+      CloudTable table = tableClient.getTableReference(JOB_TABLE);
+
+      String serializedJob = configuration.getMapper().writeValueAsString(job);
```
rtannenbaum: can this serialization be moved into PortabilityJob easily (or can you leave a TODO?) That would take care of issue #191

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureTableStore.java
```diff
@@ -0,0 +1,285 @@
...
+    try {
+
+      CloudTable table = tableClient.getTableReference(JOB_TABLE);
+
+      TableOperation retrieve =
+          TableOperation.retrieve(
+              configuration.getPartitionKey(), jobId.toString(), DataWrapper.class);
+      TableResult result = table.execute(retrieve);
+      DataWrapper wrapper = result.getResultAsType();
+      return configuration.getMapper().readValue(wrapper.getSerialized(), PortabilityJob.class);
```
rtannenbaum: Same here for the de-serialization side

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureTableStore.java
```diff
@@ -0,0 +1,285 @@
...
+    try {
+      remove(jobId, JOB_DATA_TABLE);
+    } catch (IOException e) {
+      throw new MicrosoftStorageException("Unable to remove data for job: " + jobId);
+    }
+  }
+
+  @Override
+  public void create(UUID jobId, String key, InputStream stream) {
+    try {
```
rtannenbaum: can there be any race conditions here? We use a db transaction for the Google store, maybe something similar can be done here?

Same thing in updateJob

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/DataWrapper.java
```diff
@@ -0,0 +1,35 @@
+package org.dataportabilityproject.cloud.microsoft.cosmos;
+
+import com.microsoft.azure.storage.table.TableServiceEntity;
+
+/** Wraps Job data for serialization to Cosmos DB */
+public class DataWrapper extends TableServiceEntity {
```
rtannenbaum: can this be package private?

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/MicrosoftCloudExtension.java
```diff
@@ -8,45 +8,116 @@
...
+ *       desired.
+ * </ul>
+ *
+ * Azure Setup.
  *
- * <p>
+ * <p>The following services must be setup on Azure:
+ *
+ * <ul>
+ *   <li>Active Directory: Active Directory must be enabled with a tenant
```
rtannenbaum: FYI for the GCP stuff, we scripted this to make it easier for someone to follow all the setup steps

(See config/gcp/setup_gke_environment.sh)

Though I don't think config/ is the best place for it any more. I'm going to look into moving GCP setup stuff into the google cloud extension, and move anything specific to our old GCP test project into a demo-google-deployment distribution (#202)

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/TableStoreConfiguration.java
```diff
@@ -0,0 +1,79 @@
+package org.dataportabilityproject.cloud.microsoft.cosmos;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.base.Preconditions;
+
+/** Configuration required to initialize the Azure job store */
+public class TableStoreConfiguration {
```
rtannenbaum: can this be package private?

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/TableStoreConfiguration.java
```diff
@@ -0,0 +1,79 @@
...
+
+  public String getBlobKey() {
+    return blobKey;
+  }
+
+  public ObjectMapper getMapper() {
+    return mapper;
+  }
+
+  public static class Builder {
```
rtannenbaum: FYI, totally optional but if you are interested, AutoValue does builders pretty easily (see JobAuthorization for an example)

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureTableStore.java
jimmarino: Git doesn't explicitly track renames but instead computes it by comparing the SHA hashes of the file contents so that would be difficult. I also think this is not a rename since one class that provided a piece of functionality is being removed and another class added.

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureTableStore.java
jimmarino: Yep cut-and-paste error. I'll fix it.

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureTableStore.java
jimmarino: I think each implementation should be responsible for serializing the job in a specific way, e.g. as a String or some binary format. If we move this into PortabilityJob, that latter will need to be aware of the ObjectMapper and it seems the relationship should really be the other war.

To fix #191, we should probably remove the Map conversion methods and then tweak the local and the Google implementations.

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureTableStore.java
jimmarino: I doubt there is a race condition here since the combination of jobId and key are unique and the behavior is idempotent if the creation happens twice .

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/DataWrapper.java
jimmarino: No because the Cosmos API implementations needs to reflectively instantiate types when deserializing.

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/MicrosoftCloudExtension.java
jimmarino: OK. I was planning on having the final config put into the documentation as well. It was a chore figuring out how to configure the gazillion parts of Azure  :-0

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/TableStoreConfiguration.java
jimmarino: OK

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/TableStoreConfiguration.java
jimmarino: I'll take a look. One thing is the autovalue annotation plugin is not playing nicely with IDEA's incremental compilation and Gradle since multiple generated output files are resulting in duplicates. I think this may be a bug in IDEA but haven't had time to look.

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureTableStore.java
rtannenbaum: interesting, did not know that

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureTableStore.java
jimmarino: Yeah, that would be one problem. I also think conceptually it should be the job of the store to decide how to serialize.

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/MicrosoftCloudExtension.java
jimmarino: Azure has a CLI that even runs in the browser, so I'll have to try something like what you said once all of the moving parts get organized.

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureTableStore.java
rtannenbaum: Makes sense, sounds good

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureTableStore.java
rtannenbaum: Oh ok in that case, if Cosmos doesn't support transactions for this case I agree we need to redesign. 

Another option is use a different type of hosted storage that acts more like a queue, where the store itself can be responsible for ensuring each entity is only given out to one consumer. Maybe Pubsub?

Another option, I'm not sure if this is possible but if we can enforce "final" state on the encryption key fields in the db, maybe the db operation could fail when you try and update a job's keys when it already has non-null keys assigned (from another worker).

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureKeyVaultStore.java
```diff
@@ -19,37 +16,42 @@
...
 
   public String getSecret(String key) {
-    SecretBundle secretBundle = vaultClient.getSecret(vaultUrl, "/secrets/" + key);
+    String normalizedKey = normalize(key);
+    SecretBundle secretBundle = vaultClient.getSecret(vaultUrl, normalizedKey);
     return secretBundle == null ? null : secretBundle.value();
   }
+
+  /** Normalizes a key to adhere to Azure Vault naming requirements. */
+  private String normalize(String key) {
```
holachuy: static

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureTableStore.java
```diff
@@ -0,0 +1,285 @@
...
+    }
+  }
+
+  @Override
+  public void createJob(UUID jobId, PortabilityJob job) throws IOException {
+    Preconditions.checkNotNull(jobId, "Job id is null");
+    Preconditions.checkNotNull(job, "Job is null");
+
+    try {
+
```
holachuy: extra lines here and below this method

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureTableStore.java
```diff
@@ -0,0 +1,285 @@
...
+      create(jobId, JOB_TABLE, job.jobAuthorization().state().name(), job);
+
+    } catch (JsonProcessingException e) {
+      throw new IOException(e);
```
holachuy: Any message we want to add here such as jobId or "Unable to create" or something to provide more context to this exception.

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureTableStore.java
```diff
@@ -0,0 +1,285 @@
...
+        }
+
+        validator.validate(previousJob, job);
+      }
+
+      TableOperation insert = TableOperation.insertOrReplace(wrapper);
+      table.execute(insert);
+
+    } catch (JsonProcessingException | StorageException | URISyntaxException e) {
+      throw new IOException(e);
```
holachuy: Same here, more messaging with this exception.

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureTableStore.java
```diff
@@ -0,0 +1,285 @@
...
+          TableOperation.retrieve(
+              configuration.getPartitionKey(), jobId.toString(), DataWrapper.class);
+      TableResult result = table.execute(retrieve);
+      DataWrapper wrapper = result.getResultAsType();
+
+      TableOperation delete = TableOperation.delete(wrapper);
+      table.execute(delete);
+
+    } catch (StorageException | URISyntaxException e) {
+      throw new IOException(e);
```
holachuy: Same here

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/DataWrapper.java
```diff
@@ -0,0 +1,35 @@
...
+  private String serialized;
```
holachuy: Both of these can be final

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureTableStore.java
jimmarino: One way to implement the last scenario you mentioned is with a "compare-and-set" style operation. I looked into this approach (Cassandra has something similar) and discovered Cosmos has a way of doing this.

The compare is done against the expected value of something. Both workers read, then compare-and-set. One will fail and go away while the other proceeds.

Failure recovery will be an interesting case. We'll have to decide whether duplication is tolerable. For example, when a worker dies, does the transfer start over or pick up from the last known item? If it's the latter, some form of checkpointing would be required.

I'll look into the compare-and-set once I have everything containerized on Azure.

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/TableStoreConfiguration.java
jimmarino: OK thx.

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureKeyVaultStore.java
jimmarino: Quick Q: why would this be static?

holachuy: Primarily since it doesn't depend on member variables, we can mark it as static and indicate that.

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/DataWrapper.java
jimmarino: The fields can't because Azure needs to reflectively set them. The API has a bit of a strange design.

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureTableStore.java
jimmarino: OK

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureTableStore.java
jimmarino: OK

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureTableStore.java
jimmarino: OK

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/DataWrapper.java
```diff
@@ -0,0 +1,35 @@
+package org.dataportabilityproject.cloud.microsoft.cosmos;
+
+import com.microsoft.azure.storage.table.TableServiceEntity;
+
+/** Wraps Job data for serialization to Cosmos DB */
```
holachuy: Maybe a link to the api docs for the specific requirements of this class. Maybe we want to call it CosmoDataWrapper?

extensions/cloud/portability-cloud-microsoft/src/main/java/org/dataportabilityproject/cloud/microsoft/cosmos/AzureTableStore.java
jimmarino: The failure recovery will be an interesting issue. I wonder if storing the key on the client will present an issue since users  could click away before the failure occurs or an interruption could occur.

If we externalize the key creation like we discussed in removing JobMetadata that will make the other scenario you mentioned easier.

PR #321, opened 2018-04-06
Dockerize and enable the client to build/run with Gradle

jimmarino:
A couple of changes as preconditions are included:

1. .gitignores are cleaned up and consolidated in the root
2. The client build output is moved from  "../../resources/" outside the repo root to "build/resources/" in the client project. 

*****This last change will break the Google build scripts but is needed so Dockerfiles can be created from within the repo context (Docker won't allow inclusion of files from outside the root context)****

This PR enables:

1. Building the client web app from within Gradle. This can be done by using an automatically installed Node distribution or using a pre-installed Node distro on the host

2. Running the client web app in dev mode from Gradle

3. Building Dockerfiles and Docker images for the demo with Gradle.

For example, to run a full build, install Node automatically, and create Docker images for the demo:
./gradlew -PcloudType=local clean check :distributions:demo-server:installToolsAndDockerize

To avoid the Node install (after the install is run for the first time):

./gradlew -PcloudType=local clean check :distributions:demo-server:dockerize
 
There are some other things I'll document ASAP. In the meantime, it would be good if people can try this PR out and comment as it is iterated.


client/build.gradle
```diff
@@ -0,0 +1,95 @@
+/**
+ * This subproject builds the Angular client UI web app using Node.
```
holachuy: Nice find!

distributions/demo-server/src/config/client/nginx.conf
```diff
@@ -0,0 +1,65 @@
+user  nginx;
```
holachuy: License header and maybe a quick line of documentation

distributions/demo-server/build.gradle
```diff
@@ -96,9 +108,84 @@ shadowJar {
...
+    group = 'docker'
+    destFile project.file("${buildDir}/apiserver/Dockerfile")
+    from "${javaDockerContainer}"
+    exposePort 8080
+    copyFile("build/libs/demo-server-all.jar", "/app/demo-server-all.jar")
+    defaultCommand("java", "-jar", "/app/demo-server-all.jar")
+
+    // Read in secrets and set them as environment variables.
+    // NB: This is for demo purposes only. Production systems should manage and set secrets in a secure way.
+    def secretsFile = file(Paths.get(System.getProperty('user.home')).resolve('.gradle/dataportability.secrets.properties'))
```
holachuy: Should we create a constant at the beginning of the script for this path and document it for developers, e.g. dataportability.secrets.properties

distributions/demo-server/build.gradle
```diff
@@ -96,9 +108,84 @@ shadowJar {
...
+    if (secretsFile.exists()) {
+        def props = new Properties()
+        secretsFile.withInputStream { props.load(it) }
+        props.each { k, v -> environmentVariable(k, v) }
     }
+
+}
+
+task createClientDockerfile(type: Dockerfile) {
```
holachuy: Just for clarification, this means we're running nginx within the docker container to serve the static client files after they're generated?

distributions/demo-server/src/config/client/nginx.conf
jimmarino: Sure will do

distributions/demo-server/build.gradle
jimmarino: Yeah this was one of the things I wanted to discuss further. Originally for the local cloud extension we had secretes sourced from the main settings file in the .gradle home. I think it would be better if we actually separate all of the config related to dataportability out of that and into a sidefile that sits in the .gradle home directory along side the main settings file. That way, people can share their dataportability settings (if appropriate) without having to mix in ohter things they may have in the settings file.

@rtannenbaum This can be where ENV settings are stored that get baked into Dockerfiles for demo and other non-production (i.e. "non-sensitive") deployments.

distributions/demo-server/build.gradle
jimmarino: Yeah another thing I was going to go into detail on. This is just for the demo. 

NGINX serves the webapp content as well as proxies to the API server in the other container. This requires a Docker network bridge to be created and the two docker images be bound to the network. One problem I have atm is if the client image is booted prior to the apiserver, NGINX throws an error as it attempts to contact the other container during initialization. I need to look into having that done lazily.

With the Docker plugin we can also have the build create the network bridge and the containers launched locally (or probably even in the cloud). We can look to do that once the basics are in place.

holachuy: Feel free to package the client as you see fit for demo purposes. As long as we can deploy the API without the client as well.  The initial thought is that the client would not be part of a deployment, as most use cases for hosting for DTP will be done behind an existing provider's UI.

distributions/demo-server/src/config/client/nginx.conf
holachuy: License header still doesn't appear for me.

jimmarino: Sorry added now

distributions/demo-server/build.gradle
rtannenbaum: Nice. I agree and was thinking the same thing re checking in local settings rather than in the .gradle home where they can't be shared. (Or at least a template of it if we don't want to share local app secrets)

jimmarino: I just added a note to Slack on a proposed approach to handle other config like URLs so we don't need to check them.

distributions/demo-server/build.gradle
```diff
@@ -95,9 +107,84 @@ shadowJar {
...
+    destFile project.file("${buildDir}/client/Dockerfile")
+    from "nginx"
+    exposePort 80
+    copyFile("src/config/client/nginx.conf", "/etc/nginx")
+    copyFile("build/webapp/html", "/usr/share/nginx/html")
+    copyFile("build/webapp/test-keys/server.crt", "/etc/ssl/certs")
+    copyFile("build/webapp/test-keys/server.key", "/etc/ssl")
+}
+
+task buildApiServerImage(type: DockerBuildImage) {
```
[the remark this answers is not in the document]
jimmarino: Yep I'll do that in the next PR that will shift some things around.

jimmarino:
Yep, I'll document ASAP. I'm still walking through the steps on Azure and I'll use that once it settles.

Issue #217, opened 2018-03-12
Migrate RememberTheMilk


olsona:

seehamrun:
Rmbrthemilk -> RmbrTheMilk works as expected, though we have to manually go back to our application after we authenticate (because remember the milk hasn't responded to our service ticket request to auto redirect).

PR #326, opened 2018-04-09
fix rmbr the milk exporter, when notes are null, dont access notes data.

seehamrun:
Tested with successful job from Remember the Milk -> Remember the Milk 

There are some bugs with the flow from Google -> Rmbr the milk and from Rmbr the milk -> Google, debugging those and following up in a separate PR. 

#217

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-rememberthemilk/src/main/java/org/dataportabilityproject/transfer/rememberthemilk/tasks/RememberTheMilkTasksExporter.java
+++ b/extensions/data-transfer/portability-data-transfer-rememberthemilk/src/main/java/org/dataportabilityproject/transfer/rememberthemilk/tasks/RememberTheMilkTasksExporter.java
@@ -95,11 +95,11 @@ public class RememberTheMilkTasksExporter implements Exporter<AuthData, TaskCont
     for (TaskList taskList : taskLists) {
       if (taskList.taskseries != null) {
         for (TaskSeries taskSeries : taskList.taskseries) {
-          // figure out what to do with notes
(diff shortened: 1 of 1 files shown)


extensions/data-transfer/portability-data-transfer-rememberthemilk/src/main/java/org/dataportabilityproject/transfer/rememberthemilk/tasks/RememberTheMilkTasksExporter.java
```diff
@@ -95,7 +95,9 @@ private ExportResult exportTask(
     for (TaskList taskList : taskLists) {
       if (taskList.taskseries != null) {
         for (TaskSeries taskSeries : taskList.taskseries) {
-          tasks.add(new TaskModel(oldListId, taskSeries.name, taskSeries.notes.toString()));
+          // figure out what to do with notes
+          String notesStr = taskSeries.notes == null ? "" : taskSeries.notes.toString();
```
holachuy: Add TODO. 
Also, do these have any data at all?

seehamrun: Done - Not sure if notes will ever have data. They are included in the RememberTheMilkAPI but all the examples don't have anything in the notes tag. 

Not sure if they're populated by other services - needs some more investigation but haven't been able to get a good response from google tasks - filed #327 to track Google Tasks fixes.

PR #330, opened 2018-04-10
Introduce instagram transfer extension

holachuy:
Initial empty implementations #285


settings.gradle
```diff
@@ -30,6 +30,10 @@ include ':extensions:data-transfer:portability-data-transfer-flickr'
 include ':extensions:auth:portability-auth-google'
 include ':extensions:data-transfer:portability-data-transfer-google'
 
+// Instagram Auth and Transfer Extensions
+//include ':extensions:auth:portability-auth-instagram'
```
seehamrun: Can we remove the portability-auth-instagram and add it back in when we implement it?

holachuy: Done

PR #335, opened 2018-04-10
update documentation and scripts for running web application locally

seehamrun:
Let me know if these changes make sense - I updated the gradle script to use the port 3000 for the webapp so we dont have to change any local yaml files, but if theres strong interest, can change it to 3030.


holachuy:
I'm going to submit these in case others are trying out the project as well as for others to modify if necessary.

jimmarino:
Sorry got this in the morning my time. Thanks!

PR #345, opened 2018-04-11
fix google tasks

seehamrun:
Cleanup Google tasks. Tested with Google -> Google

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/dataportabilityproject/datatransfer/google/tasks/GoogleTasksImporter.java
+++ b/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/dataportabilityproject/datatransfer/google/tasks/GoogleTasksImporter.java
@@ -86,10 +86,12 @@ public class GoogleTasksImporter implements Importer<TokensAndUrlAuthData, TaskC
     tempTasksData = jobStore.findData(TempTasksData.class, jobId);
 
     for (TaskModel oldTask : data.getTasks()) {
+      // TODO: The TaskModel doesn't contain information about completion, which means these all are
+      // new uncompleted tasks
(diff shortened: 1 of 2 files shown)


extensions/data-transfer/portability-data-transfer-google/src/main/java/org/dataportabilityproject/datatransfer/google/tasks/GoogleTasksImporter.java
```diff
@@ -86,10 +86,12 @@ public ImportResult importItem(
     tempTasksData = jobStore.findData(TempTasksData.class, jobId);
 
     for (TaskModel oldTask : data.getTasks()) {
+      // TODO: The TaskModel doesn't contain information about completion, which means these all are
+      // new uncompleted tasks
       Task newTask = new Task().setTitle(oldTask.getText()).setNotes(oldTask.getNotes());
-      String newTaskId = tempTasksData.lookupNewTaskListId(newTask.getId());
+      String newTaskListId = tempTasksData.lookupNewTaskListId(oldTask.getTaskListId());
```
holachuy: I almost want to call the method lookupNewTaskListIdFromOldTaskListId

Or maybe lookupTaskListMapping?

seehamrun: We have the same naming convention in photos too >.< 
I can send a followup PR to rename them if we feel strongly

holachuy: Yeah, in addition to naming do we want to have an abstract base class or interface for naming that enforces a specific convention. We seem to have similar functionality in different places.

PR #353, opened 2018-04-12
Google Photos tests

olsona:


extensions/data-transfer/portability-data-transfer-google/src/main/java/org/dataportabilityproject/datatransfer/google/photos/GooglePhotosExporter.java
```diff
@@ -111,31 +115,32 @@ public GooglePhotosExporter(GoogleCredentialFactory credentialFactory) {
...
+      albumUrl = new URL(String.format(URL_ALBUM_FEED_FORMAT, startItem, maxResults));
       albumFeed = getOrCreatePhotosService(authData).getFeed(albumUrl, UserFeed.class);
     } catch (ServiceException | IOException e) {
       return new ExportResult<>(ResultType.ERROR, e.getMessage());
     }
 
     PaginationData nextPageData = null;
-    if (albumFeed.getEntries().size() == MAX_RESULTS) {
-      int nextPageStart = startItem + MAX_RESULTS;
+    List<AlbumEntry> entries = albumFeed.getAlbumEntries();
```
seehamrun: I couldn't quite figure out why, but when I was testing the Google photos flow, getAlbumEntries returned me an empty list always, so there was never any data to export. 
Is this resolved? Or (more likely) was there another reason this might be empty?

Issue #236, opened 2018-03-13
Migrate build and docker bash scripts to gradle


rtannenbaum:
The old bash script, bin/build_and_run_jar.sh, is responsible for copying resources like secrets and config into our jar.

However it's not easily supported on Windows and we're now on Gradle, and have a gradle build file which can also dynamically copy in resources.

The gradle build can already be invoked on either platform, so we should migrate the build script to our gradle build file.

rtannenbaum:
The docker script (build_docker_image.sh) can also be moved to gradle, using the gradle docker plugin as Jim did in the demo-server

PR #355, opened 2018-04-12
Dockerize transfer server gradle build for demo Google distribution

rtannenbaum:
#236


distributions/demo-google-deployment/transfer/build.gradle
```diff
@@ -26,29 +45,69 @@ repositories {
...
-}
-
-def copyResources() {
     copy {
         from '../resources/'
         into 'src/main/resources'
     }
 }
+
+// Note: This service acct creds file should be used for local testing only. In production, we use
```
holachuy: Do we need to make this flag driven?  Or, ideally, can we have a local and prod deploy script that overrides this?

distributions/demo-google-deployment/transfer/build.gradle
```diff
@@ -26,29 +45,69 @@ repositories {
...
+task createTransferServerDockerfile(type: Dockerfile) {
+    description = 'Builds the Transfer Server Dockerfile'
+    group = 'docker'
+    dependsOn copyTransferServerDeps
+    destFile project.file("${buildDir}/transfer/Dockerfile")
+    from "gcr.io/google-appengine/openjdk:8"
+    exposePort 8082 // Port the TRANSFER server is accessed from
+    copyFile("build/libs/transfer-all.jar", "/app/transfer.jar")
+
+    // Note: Debug port, project ID, and service_acct_creds.json set in the image are for local
```
holachuy: Same here

jimmarino:
Looks good. I'll look into how to create a task that rolls up other tasks a flags to build a set of docker distributions as we discussed.

distributions/demo-google-deployment/transfer/build.gradle
```diff
@@ -26,29 +45,69 @@ repositories {
...
     compile project(':extensions:data-transfer:portability-data-transfer-google')
+    compile project(':extensions:data-transfer:portability-data-transfer-instagram')
     compile project(':extensions:data-transfer:portability-data-transfer-microsoft')
-    compile project(':extensions:cloud:portability-cloud-google')
-    compile project(':extensions:config:portability-config-yaml')
+    compile project(':extensions:data-transfer:portability-data-transfer-flickr')
+    compile project(':extensions:data-transfer:portability-data-transfer-rememberthemilk')
 }
 
 addCloudExtensionDependency(project)
```
seehamrun: Should we explicitly include the Google Cloud extension here or leave it up to the flag? Does it ever make sense to build this with any other cloud type?

distributions/demo-google-deployment/transfer/build.gradle
rtannenbaum: Yeah, I was going to come back to this once I start looking into the prod setup. I have a note at the top about the next step being to support non-local environments, I will come back to that in the next PR

distributions/demo-google-deployment/transfer/build.gradle
rtannenbaum: Same - will come back to this in the next PR, will have to refactor the local setup out somehow

distributions/demo-google-deployment/transfer/build.gradle
rtannenbaum: I actually brought this up in Slack too, that this should always be cloudType Google. cloudType is referenced other places in the main build file, so I thought it would be better to set cloudType to Google since we will need that anyway, and allow addCloudExtensionDependency to use that. That also allows for addCloudExtensionDependency to be refactored to a common place and not need to hardcode the cloud extension for each distro.

Unfortunately, setting cloudType is harder than it seems, Jim is going to come back to that in his PR, but for now we're just setting it on the command line to ./gradlew

seehamrun: sgtm, optionally if its possible, can we check that the cloudType in the param is Google and fail the build?

seehamrun:
Looks like the travis build is failing, since there's no gcp project defined: 

"
* Where:
Build file '/home/travis/build/google/data-transfer-project/distributions/demo-google-deployment/transfer/build.gradle' line: 100
* What went wrong:
A problem occurred evaluating project ':distributions:demo-google-deployment:transfer'.
> Could not get unknown property 'gcpProject' for task ':distributions:demo-google-deployment:transfer:createTransferServerDockerfile' of type com.bmuschko.gradle.docker.tasks.image.Dockerfile.
" 

Is it possible to set a default?

distributions/demo-google-deployment/transfer/build.gradle
rtannenbaum: that's a good idea but I think it may fail the Travis build depending how it is done. Since Jim is coming right back to this I left a TODO, since it will always be google in his impl.

rtannenbaum:
@seehamrun good catch - fixed

rtannenbaum:
Thanks all for the good comments!

PR #360, opened 2018-04-13
Split GCP demo API build into separate tasks for local and prod 

rtannenbaum:
Split GCP demo API build into separate tasks for local and prod (#202) (as discussed in PR #354)

Also copy index.html and environment-specific settings into the jar (#236)

Tested inspecting src/main/resources for the appropriate config files and env/ override files (see path constants in YamlSettingsExtention.java)

Jim mentioned he is going to change how api.yaml and gitignore'ing of settings works. I wanted to get this in for now but am happy to update it once we change that.


rtannenbaum:
If this looks OK, I'll split up local and prod the same way for the transfer build

distributions/demo-google-deployment/api/build.gradle
```diff
@@ -57,35 +58,73 @@ dependencies {
...
 }
 
-// TODO: Include index.html in jar
 
 addCloudExtensionDependency(project)
 
 mainClassName = 'org.dataportabilityproject.api.ApiMain'
 
+task copyEnvResources(type: Copy) {
+    description = 'Copies environment-specific resources and configuration into the jar, ' +
```
holachuy: Could we eventually remove index.html from being copied in case the deployment is using rpc instead of http

distributions/demo-google-deployment/api/build.gradle
```diff
@@ -57,35 +58,73 @@ dependencies {
...
-task copyApiServerDeps {
+task createApiServerDockerfile(type: Dockerfile) {
+    description = 'Builds the API Server Dockerfile'
+    group = 'docker'
+    destFile project.file("${buildDir}/api/Dockerfile")
+    from "gcr.io/google-appengine/openjdk:8"
+    exposePort 8080 // Port the API server is accessed from
+    copyFile("build/libs/api-all.jar", "/app/api.jar")
+    defaultCommand("java",
+            "-Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=5005",
```
holachuy: TODO: make debug optional or flag driven

distributions/demo-google-deployment/api/build.gradle
rtannenbaum: I added some comments to the README clarifying how this works and how this is just one way DTP may be deployed on GCP. I think the RPC stuff will need further discussion so I'll leave it out of this demo google deployment (perhaps that could be another demo google deployment later on).

distributions/demo-google-deployment/api/build.gradle
rtannenbaum: Done

PR #364, opened 2018-04-16
Split GCP demo transfer build into separate tasks for local and prod

### LINKED ISSUE #202: Move GCP-specific config into Google distribution

rtannenbaum:
Split GCP demo transfer build into separate tasks for local and prod

Refactors the remaining GCP config into resources/config and resources/config/\<environment name\> (the latter is gitignored to hide demo domain names). Closes #202. This will probably be reworked soon as Jim is moving baseUrl and baseApiUrl out of api.yaml. 

Copies environment-specific settings into the transfer jar (#236). Tested that the correct resources are included:
- log4j.properties
- config/common.yaml
- config/env/common.yaml

Cleans up how this is done for the api jar as well. Tested the correct resources are included:
- log4j.properties
- config/index.html
- config/api.yaml
- config/common.yaml
- config/env/api.yaml
- config/env/common.yaml


distributions/demo-google-deployment/api/build.gradle
```diff
@@ -63,20 +63,40 @@ addCloudExtensionDependency(project)
 
 mainClassName = 'org.dataportabilityproject.api.ApiMain'
 
-task copyEnvResources(type: Copy) {
-    description = 'Copies environment-specific resources and configuration into the jar, ' +
-            'including index.html'
+task copyIndexHtml(type: Copy) {
+    description = 'Copies index.html into the jar'
+    ext.env = project.hasProperty('env') ? env : 'local'
+    into 'src/main/resources/config'
```
seehamrun: Does the build_and_deploy_static_content script move the index file in resource/config/enviornments?

distributions/demo-google-deployment/api/build.gradle
```diff
@@ -95,7 +115,7 @@ task createApiServerDockerfile(type: Dockerfile) {
     copyFile("build/libs/api-all.jar", "/app/api.jar")
     // TODO: make debug flag driven
     defaultCommand("java",
-            "-Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=5005",
+            "-Xrunjdwp:transport=dt_socket,server=y,suspend=n",
```
seehamrun: Looks like the adress got dropped from this, is that intentional?

distributions/demo-google-deployment/transfer/build.gradle
```diff
@@ -23,16 +23,21 @@ import com.bmuschko.gradle.docker.tasks.image.Dockerfile
...
- * resulting images work when deployed to GCP.
+ * <p>TODO(#202): Test deploying a prod image to GCP
  *
  * <p>TODO(@jimmarino): Automatically set cloudType=google so it is not needed on the commandline
  *
- * <ul>To run:
+ * <ul>To run locally:
  *  <li>./gradlew -PcloudType=google -PgcpProject=<your-project-name> \
- *      :distributions:demo-google-deployment:transfer:dockerize
+ *      :distributions:demo-google-deployment:transfer:dockerizeLocal
```
seehamrun: Should we update this in the Documentation.md file as well?

distributions/demo-google-deployment/transfer/build.gradle
```diff
@@ -60,56 +65,106 @@ addCloudExtensionDependency(project)
...
+task createTransferServerDockerfile(type: Dockerfile) {
+    description = 'Builds the Transfer Server Dockerfile'
+    group = 'docker'
+    destFile project.file("${buildDir}/transfer/Dockerfile")
+    from "gcr.io/google-appengine/openjdk:8"
+    exposePort 8080 // Port the transfer server is accessed from
+    copyFile("build/libs/transfer-all.jar", "/app/transfer.jar")
+    // TODO: make debug flag driven
+    defaultCommand("java",
+            "-Xrunjdwp:transport=dt_socket,server=y,suspend=n",
```
seehamrun: same question here regarding debugt port

distributions/demo-google-deployment/api/build.gradle
seehamrun: Or wait, is this for the production env? Do we need this flag at all then?

distributions/demo-google-deployment/api/build.gradle
rtannenbaum: Good catch, it does not, just updated it

distributions/demo-google-deployment/api/build.gradle
rtannenbaum: Exactly - this is the production image so I didn't want to expose debugging. I'm removing the TODO here (and leaving it for the local task)

distributions/demo-google-deployment/transfer/build.gradle
rtannenbaum: Where do you think it should go? Developer.md has instructions for demo-server but not demo-google-deployment currently. I'll add a comment about demo-google-deployment there.

distributions/demo-google-deployment/transfer/build.gradle
rtannenbaum: Done

distributions/demo-google-deployment/transfer/build.gradle
seehamrun: Oh youre right, I got confused with which distro was being run here. Maybe we should include this in the developer documentation under the google-demo-deployment (pointing to that in the developer.md)?  Maybe the README here: https://github.com/google/data-transfer-project/blob/master/distributions/demo-google-deployment/README.md ? Feel free to leave as is or create an issue to track the documentation changes we need.

distributions/demo-google-deployment/api/build.gradle
holachuy: Later on down the line, we will probably not include this for RPC only serving so we may want to make it optional

distributions/demo-google-deployment/api/build.gradle
```diff
@@ -93,9 +113,8 @@ task createApiServerDockerfile(type: Dockerfile) {
     from "gcr.io/google-appengine/openjdk:8"
     exposePort 8080 // Port the API server is accessed from
     copyFile("build/libs/api-all.jar", "/app/api.jar")
-    // TODO: make debug flag driven
     defaultCommand("java",
-            "-Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=5005",
+            "-Xrunjdwp:transport=dt_socket,server=y,suspend=n",
```
holachuy: Should these be driven by a flag?

distributions/demo-google-deployment/api/build.gradle
rtannenbaum: Ack. I'd rather figure that out in a separate distribution, leave this one as an HTTP-serving one

distributions/demo-google-deployment/api/build.gradle
rtannenbaum: I have a TODO to do so for the dockerizeLocal task. I left it out of this task since it's for production and we shouldn't expose a debug port.

distributions/demo-google-deployment/transfer/build.gradle
rtannenbaum: Good point, I"ll add something there once I get this working deployed to GCP, I'll document how to run it both locally and for prod (as part of #236)

Issue #362, opened 2018-04-16
Smugmug transfer extension


seehamrun:

seehamrun:
Whats left: 

* Move importer side to use the scribe API for signing post requests
* Fully test export. 
* The export only gets public data - this is likely an issue with the auth data generator and needs debugging
* Can't get smugmug -> flickr working due to unrecongized file type, and smugmug -> google photos renders a black image instead of the actual image (not clear if the issue is on the exporter or the importer side, since for public images, the photo resources look correct.)

seehamrun:
Still can't get smugmug upload to work, going to put on the backlog for now and disable from the demo-server 

Updated list of what's left / errors I'm running into:
* Importer side uses scribe API and album creation works, but image upload doesn't and just has an "unauthorized" error (and no details as to why).
  * I thought this might be because we mark the album as private and then try to upload photos, but making albums public didn't fix things either. 
* AuthDataGenerator now has access to private data and correct permissions, but it still doesnt look like we get "hidden" images within albums (we do get private albums though, haven't tested private albums with content..). 
* public images uploaded to other services (Google photos) are still corrupt (black empty image), even though I am able to get to the content fetchable url and download successfully - This might be the same reason why Flickr is rejecting the photo with "unrecognized file type".

PR #366, opened 2018-04-16
Smugmug Exporter

seehamrun:
Wire up the auth data to requests made by the smugmug exporter. This switches to using the scribe API for making oauth requests, and does some type changes and modifications to the existing SmugMugInterface.

This appears to be working as there are data models and resources filled in and passed onto the importer chosen, im still working out some kinks in the flow from smugmug -> flickr, regarding the file types, so once I get down to that cause, i'll send a followup PR, but just sending this out for review to make my future changes to the importer easier to understand. 

#362


extensions/data-transfer/portability-data-transfer-smugmug/src/main/java/org/dataportabilityproject/transfer/smugmug/photos/SmugMugPhotosImporter.java
```diff
@@ -119,33 +141,32 @@ void importSingleAlbum(UUID jobId, String folder, PhotoAlbum inputAlbum) throws
...
-            ImmutableMap.of(
-                    "X-Smug-AlbumUri", "/api/v2/album/" + newAlbumKey,
-                    "X-Smug-ResponseType", "json",
-                    "X-Smug-Version", "v2"),
-            new TypeReference<ImageUploadResponse>() {});
-
+        "http://upload.smugmug.com/",
+        content,
+        // Headers from: https://api.smugmug.com/api/v2/doc/reference/upload.html
+        ImmutableMap.of(
```
holachuy: Maybe this map creation can be broken up into a different method that takes in the album key

extensions/data-transfer/portability-data-transfer-smugmug/src/main/java/org/dataportabilityproject/transfer/smugmug/photos/SmugMugInterface.java
```diff
@@ -34,54 +34,69 @@
...
+import org.scribe.builder.ServiceBuilder;
+import org.scribe.model.OAuthRequest;
+import org.scribe.model.Response;
+import org.scribe.model.Token;
+import org.scribe.model.Verb;
+import org.scribe.oauth.OAuthService;
 
 public class SmugMugInterface {
 
   // TODO(olsona): is this the mapper we want to use?
```
holachuy: Let's inject a mapper passed through from the TransferExtension here

extensions/data-transfer/portability-data-transfer-smugmug/src/main/java/org/dataportabilityproject/transfer/smugmug/photos/SmugMugPhotosImporter.java
seehamrun: Yea, I was planning on moving this with more substantial changes when switching the postRequest function to using the scribe API and signing in a followup PR. I didn't actually modify the Importer yet aside from the type changes necessary for compiling.

extensions/data-transfer/portability-data-transfer-smugmug/src/main/java/org/dataportabilityproject/transfer/smugmug/photos/SmugMugInterface.java
seehamrun: Done.

PR #372, opened 2018-04-17
slight refactoring of smugmug interface

seehamrun:
Use the actual response instead of the wrapper type.
Make all smugmug keys and internal urls private
createAlbum() to wrap around the post request.
createPhoto() to wrap around the photo request

Note: Importer doesnt use the auth data yet. 
#362


extensions/data-transfer/portability-data-transfer-smugmug/src/main/java/org/dataportabilityproject/transfer/smugmug/photos/SmugMugInterface.java
```diff
@@ -70,22 +88,88 @@
...
+        .getResponse();
   }
 
-  SmugMugResponse<SmugMugUserResponse> makeUserRequest() throws IOException {
-    return makeRequest(USER_URL, new TypeReference<SmugMugResponse<SmugMugUserResponse>>() {});
+  /* Creates an album with albumName provided. */
+  SmugMugAlbumResponse createAlbum(String albumName) throws IOException {
+    // Set up album
+    Map<String, String> json = new HashMap<>();
+    String niceName = "Copy-of-" + albumName.replace(' ', '-');
```
rtannenbaum: Why copy-of? Shouldn't we use the same album name as the original? Unless we are copying to the same account, i guess..

extensions/data-transfer/portability-data-transfer-smugmug/src/main/java/org/dataportabilityproject/transfer/smugmug/photos/SmugMugInterface.java
```diff
@@ -70,22 +88,88 @@
...
+    checkState(response.getResponse().getAlbum() != null, "Album is null");
+
+    return response.getResponse();
   }
 
-  <T> SmugMugResponse<T> makeRequest(String url, TypeReference<SmugMugResponse<T>> typeReference)
-      throws IOException {
+  /* Uploads the resource at photoUrl to the albumId provided
+   * The albumId must exist before calling upload, else the request will fail */
+  ImageUploadResponse uploadImage(String photoUrl, String albumId) throws IOException {
```
rtannenbaum: uploadImage is specific to the importer only, right? I think it should belong in the importer then

extensions/data-transfer/portability-data-transfer-smugmug/src/main/java/org/dataportabilityproject/transfer/smugmug/photos/SmugMugPhotosExporter.java
```diff
@@ -93,48 +89,54 @@ public SmugMugPhotosExporter(
     IdOnlyContainerResource resource =
         (IdOnlyContainerResource) exportInformation.getContainerResource();
 
+    SmugMugInterface smugMugInterface;
+
+    try {
+      smugMugInterface = getOrCreateSmugMugInterface(authData);
```
rtannenbaum: can smugmugInterface be initialized in the constructor? can we use the member var instead of passing it to the export methods?

extensions/data-transfer/portability-data-transfer-smugmug/src/main/java/org/dataportabilityproject/transfer/smugmug/photos/SmugMugInterface.java
seehamrun: This is what was done in the FlickrImporter too (and I recall also in google photos before I removed album support).

extensions/data-transfer/portability-data-transfer-smugmug/src/main/java/org/dataportabilityproject/transfer/smugmug/photos/SmugMugInterface.java
seehamrun: It is, but since it does a lot of specific  smugmug http header and body modifications, I think it made more sense to leave it with the smugmug interface (because also that is what will be signing the post request).

extensions/data-transfer/portability-data-transfer-smugmug/src/main/java/org/dataportabilityproject/transfer/smugmug/photos/SmugMugPhotosExporter.java
seehamrun: Ideally yes, esp if each job/worker will only handle one data transfer and then exit. But the issue right now is that they are reused (and the extensions don't provide a brand new instance each time getExporter() or getImporter() is called). Each service instance is specific to the authdata passed in and makes requests specifically for that user. 

This is the same pattern that is done in the Google services. For Flickr, we just call a function to set the request context to the new user auth data. This seemed better so that we don't accidentally forget to reset the context.

holachuy: Yeah, the auth data is request time (and should be as we may need to refresh expired tokens) but some sdk's and client require it at construction time. There may be better refactoring here, e.g. use a factory to wrap this functionality or some other pattern but we figured we'd do that once we get all the code migrated.

extensions/data-transfer/portability-data-transfer-smugmug/src/main/java/org/dataportabilityproject/transfer/smugmug/photos/SmugMugInterface.java
holachuy: Yeah, from other migration projects we have experience with, it functions as a way of distinguishing imported content. Another approach is to put all content in a new top level album or tag each one.

PR #380, opened 2018-04-18
Expose common and local settings. Hide other environment settings.

rtannenbaum:
This provides an example yaml config setup (local) without exposing domain names for our demo app


distributions/demo-google-deployment/resources/config/environments/qa/index.html
```diff
@@ -1 +0,0 @@
-<!doctype html><html><head><meta charset="utf-8"><title>PortabilityDemo</title><base href="/"><meta name="viewport" content="width=device-width,initial-scale=1"><link rel="icon" type="image/x-icon" href="favicon.ico"><link href="static/styles.eb6697daa4a627e6f5c6.bundle.css" rel="stylesheet"/></head><body><app-root>Loading...</app-root><script type="text/javascript" src="static/inline.31e1fb380eb7cf3d75b1.bundle.js"></script><script type="text/javascript" src="static/polyfills.197f206e141f3b499361.bundle.js"></script><script type="text/javascript" src="static/vendor.9dc5b91a0a5845e9dfb9.bundle.js"></script><script type="text/javascript" src="static/main.e0305332c78bebdc7588.bundle.js"></script></body></html>
```
seehamrun: Did we ever decide if we wanted to check in the index.html? I guess if the build/deploy script also updates the bucket contents, there's no need. But otherwise wouldn't it be possible for someone to build and deploy a docker image with a different index.html than what's available in the cloud bucket?

seehamrun:
sorry for the delayed response, I'm not sure why I never got an email about this review :(

rtannenbaum:
This is pretty out of date and probably obsolete. We should update demo-google-deployment to reflect the recent handler refactors.

PR #381, opened 2018-04-18
Google photos importer test

olsona:


libraries/transfer/src/main/java/org/dataportabilityproject/transfer/ImageStreamProvider.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package org.dataportabilityproject.transfer;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.HttpURLConnection;
+import java.net.URL;
+
+public class ImageStreamProvider {
```
holachuy: It seems like we can just call this UrlInputStreamReader or something like that since it doesn't have any logic directly related to images and can probably be used in other fetching scenarios.

extensions/data-transfer/portability-data-transfer-google/src/test/java/org/dataportabilityproject/datatransfer/google/photos/GooglePhotosImporterTest.java
```diff
@@ -15,5 +15,81 @@
...
+import org.dataportabilityproject.transfer.ImageStreamProvider;
+import org.dataportabilityproject.types.transfer.models.photos.PhotoModel;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Matchers;
+
 public class GooglePhotosImporterTest {
+
+  private String PHOTO_TITLE = "Model photo title";
```
holachuy: these constants can be static final

extensions/data-transfer/portability-data-transfer-google/build.gradle
```diff
@@ -28,7 +29,7 @@ dependencies {
     compile("com.google.apis:google-api-services-gmail:${googleGmailVersion}")
     compile "com.google.apis:google-api-services-people:${googlePeopleVersion}"
```
seehamrun: nit include the ( here too

libraries/transfer/src/main/java/org/dataportabilityproject/transfer/ImageStreamProvider.java
```diff
@@ -0,0 +1,35 @@
...
+
+  /**
+   * Gets an input stream to an image, given its URL.
+   */
+  public InputStream get(String urlStr) throws IOException {
```
seehamrun: Just curious, why can't this be a static utility function?

olsona: So I can mock it.

extensions/data-transfer/portability-data-transfer-google/build.gradle
olsona: Done.

libraries/transfer/src/main/java/org/dataportabilityproject/transfer/ImageStreamProvider.java
olsona: Done.

extensions/data-transfer/portability-data-transfer-google/src/test/java/org/dataportabilityproject/datatransfer/google/photos/GooglePhotosImporterTest.java
olsona: Done.

Issue #384, opened 2018-04-19
Implement temp data methods in GoogleJobStore


holachuy:

lisad:
Obsolete - JZacsh has been fixing issues in this part of the code much more recently

PR #385, opened 2018-04-19
Google Job Store temp data implementation

holachuy:
Google Job Store temp data implementation #384

### DIFF
--- a/extensions/cloud/portability-cloud-google/src/main/java/org/dataportabilityproject/cloud/google/GoogleJobStore.java
+++ b/extensions/cloud/portability-cloud-google/src/main/java/org/dataportabilityproject/cloud/google/GoogleJobStore.java
@@ -15,6 +15,7 @@
  */
 package org.dataportabilityproject.cloud.google;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.cloud.Timestamp;
 import com.google.cloud.datastore.Blob;
 import com.google.cloud.datastore.BooleanValue;
@@ -30,6 +31,7 @@ import com.google.cloud.datastore.StringValue;
 import com.google.cloud.datastore.StructuredQuery.PropertyFilter;
 import com.google.cloud.datastore.TimestampValue;
 import com.google.cloud.datastore.Transaction;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
 import com.google.inject.Inject;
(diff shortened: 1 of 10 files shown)


portability-spi-cloud/src/main/java/org/dataportabilityproject/spi/cloud/storage/JobStore.java
```diff
@@ -72,7 +72,7 @@
    */
   UUID findFirst(JobAuthorization.State jobState);
 
-  default <T extends DataModel> void create(UUID jobId, T model) {
+  default <T extends DataModel> void create(UUID jobId, T model) throws IOException {
```
seehamrun: Looks like the travis build is failing since some of the classes that use this dont catch the exception: 
> flickr/src/main/java/org/dataportabilityproject/datatransfer/flickr/photos/FlickrPhotosImporter.java:89: error: unreported exception IOException; must be caught or declared to be thrown
>       jobStore.create(jobId, tempPhotosData);

I think this also has to do with some of the discussion in #176, so tagging that explicitly.

extensions/cloud/portability-cloud-google/src/main/java/org/dataportabilityproject/cloud/google/GoogleJobStore.java
```diff
@@ -251,4 +265,73 @@ private Key getKey(UUID jobId) {
...
+        .set(CREATED_FIELD, Timestamp.now())
+        .set(model.getClass().getName(), serialized)
+        .build();
+
+    try {
+      transaction.put(entity);
+    } catch (DatastoreException e) {
+      transaction.rollback();
+      throw new IOException(
+          "Could not create initial record for jobID: " + jobId + ". Record: " + entity, e);
```
rtannenbaum: Can you log data type here as well? or log the key as you do below, since that should include job ID and data type

extensions/cloud/portability-cloud-google/src/main/java/org/dataportabilityproject/cloud/google/GoogleJobStore.java
```diff
@@ -251,4 +265,73 @@ private Key getKey(UUID jobId) {
...
+
+  @Override
+  public <T extends DataModel> T findData(Class<T> type, UUID jobId) {
+    Key key = getDataKey(jobId, type);
+    Entity entity = datastore.get(key);
+    if (entity == null) {
+      return null;
+    }
+    try {
+      return  objectMapper.readValue(entity.getBlob(type.getName()).toByteArray(), type);
```
rtannenbaum: nit: I think there's an extra space in here after return

PR #389, opened 2018-04-20
Update script to remove legacy path references

holachuy:
Also added a couple of checks #387

### DIFF
--- a/distributions/demo-google-deployment/bin/encrypt_and_upload_secrets.sh
+++ b/distributions/demo-google-deployment/bin/encrypt_and_upload_secrets.sh
@@ -103,14 +103,15 @@ if [[ ! -e $ENCRYPTED_SECRETS_DIR ]]; then
   echo -e "Directory $ENCRYPTED_SECRETS_DIR does not already exist... creating it"
   mkdir $ENCRYPTED_SECRETS_DIR
   echo -e "Encrypting secrets..."
+  # TODO: Make this list configurable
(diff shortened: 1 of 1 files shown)


distributions/demo-google-deployment/bin/encrypt_and_upload_secrets.sh
```diff
@@ -81,26 +98,26 @@ else
...
   echo -e "Encrypting secrets..."
   encrypt_secret "FLICKR"
   encrypt_secret "GOOGLE"
-  encrypt_secret "INSTAGRAM"
-  encrypt_secret "MICROSOFT"
-  encrypt_secret "RTM"
-  encrypt_secret "SHUTTERSTOCK"
-  encrypt_secret "SMUGMUG"
-  encrypt_secret "JWT"
+  #encrypt_secret "INSTAGRAM"
```
rtannenbaum: can this either be uncommented, or better yet, can it encrypt based on the contents of the dir? Fine to leave a TODO also

seehamrun: Why are these other ones commented out? Should we also allow a list of secrets to update as a cmd line arg? (and default to all of them if no arg is specified)

seehamrun: just saw @rtannenbaum's comment, either way is fine!

PR #394, opened 2018-04-25
Update Developer.md

olsona:


Documentation/Developer.md
```diff
@@ -62,7 +62,7 @@ limitations under the License.
 
 ## Creating the Docker Network
 * `sudo service docker start`
-* `docker network create dataportability`
+* `docker network create dataportability` - note you only need to do this once
```
seehamrun: Does this apply to the sudo service command above too?

olsona: Honestly, not sure.  I've been stopping and starting each time.  @rtannenbaum, can you weigh in?  When is starting/stopping necessary?

PR #395, opened 2018-04-26
Fix local demo google docker cred copy task

holachuy:
This allows the task to build the local transfer tasking running against test GCP to complete


rtannenbaum:
fixes #392

holachuy:
Updated documentation per offline conversation. Also removed suspend on startup.

Issue #376, opened 2018-04-18
GCP - Workers should keep polling if they fail to claim a job


rtannenbaum:
Somewhere in the migration, this broke, and the workers now terminate when they fail to claim a job. They immediately restart, so it still works, but they should just continue to keep polling and not terminate.

PR #399, opened 2018-05-01
Fix bug where worker would die after failing to claim a job

olsona:
When a worker failed to claim a job, it would throw and exception and die.  This fix updates it so that claiming a job is a matter of success/failure, instead of success/exception, and if a worker fails to claim a job, it can continue to poll.

### DIFF
--- a/portability-transfer/src/main/java/org/dataportabilityproject/transfer/JobPollingService.java
+++ b/portability-transfer/src/main/java/org/dataportabilityproject/transfer/JobPollingService.java
@@ -86,9 +86,8 @@ class JobPollingService extends AbstractScheduledService {
     // encrypt and store the private key on the client.
     // Note: tryToClaimJob may fail if another transfer worker beat us to it. That's ok -- this transfer
     // worker will keep polling until it can claim a job.
-    boolean claim = tryToClaimJob(jobId, keyPair);
-    logger.debug("Claim: " + claim);
-    if (claim) {
+    boolean claimed = tryToClaimJob(jobId, keyPair);
+    if (claimed) {
       logger.debug(
           "Updated job {} to CREDS_ENCRYPTION_KEY_GENERATED, publicKey length: {}",
           jobId,
@@ -96,7 +95,8 @@ class JobPollingService extends AbstractScheduledService {
     }
   }
 
-  /** Claims {@link PortabilityJob} {@code jobId} and updates it with our public key in storage. */
+  /** Claims {@link PortabilityJob} {@code jobId} and updates it with our public key in storage.
+   * Returns true if the claim was successful; otherwise it returns false. */
   private boolean tryToClaimJob(UUID jobId, KeyPair keyPair) {
     // Lookup the job so we can append to its existing properties.
     PortabilityJob existingJob = store.findJob(jobId);
--- a/portability-transfer/src/main/java/org/dataportabilityproject/transfer/JobPollingService.java
+++ b/portability-transfer/src/main/java/org/dataportabilityproject/transfer/JobPollingService.java
@@ -62,14 +62,14 @@ class JobPollingService extends AbstractScheduledService {
 
   @Override
   protected Scheduler scheduler() {
-    return AbstractScheduledService.Scheduler.newFixedDelaySchedule(0, 20, TimeUnit.SECONDS);
+    return AbstractScheduledService.Scheduler.newFixedDelaySchedule(0, 2, TimeUnit.SECONDS);
   }
 
   /**
    * Polls for an unassigned job, and once found, initializes the global singleton job metadata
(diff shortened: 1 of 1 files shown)


portability-transfer/src/main/java/org/dataportabilityproject/transfer/JobPollingService.java
```diff
@@ -60,6 +60,7 @@ protected void runOneIteration() throws Exception {
     }
   }
 
+  // TODO: the delay should be more easily configurable
```
rtannenbaum: Can you create & reference an issue for this?

portability-transfer/src/main/java/org/dataportabilityproject/transfer/JobPollingService.java
```diff
@@ -83,22 +84,26 @@ private void pollForUnassignedJob() throws IOException {
...
     // encrypt and store the private key on the client.
-    // Note: claimJob may fail if another transfer worker beat us to it. That's ok -- this transfer
+    // Note: tryToClaimJob may fail if another transfer worker beat us to it. That's ok -- this transfer
     // worker will keep polling until it can claim a job.
-    claimJob(jobId, keyPair);
-    logger.debug(
-        "Updated job {} to CREDS_ENCRYPTION_KEY_GENERATED, publicKey length: {}",
-        jobId,
-        publicKey.getEncoded().length);
+    boolean claim = tryToClaimJob(jobId, keyPair);
```
rtannenbaum: nit: claimed?

portability-transfer/src/main/java/org/dataportabilityproject/transfer/JobPollingService.java
```diff
@@ -83,22 +84,26 @@ private void pollForUnassignedJob() throws IOException {
...
+    logger.debug("Claim: " + claim);
```
rtannenbaum: Can you consolidate this log with the one below, and make it clearer so you don't have to look up what "claim: false" means? i.e. "successfully claimed blah" or "failed to claim blah"

portability-transfer/src/main/java/org/dataportabilityproject/transfer/JobPollingService.java
```diff
@@ -83,22 +84,26 @@ private void pollForUnassignedJob() throws IOException {
...
+    if (claim) {
+      logger.debug(
+          "Updated job {} to CREDS_ENCRYPTION_KEY_GENERATED, publicKey length: {}",
+          jobId,
+          publicKey.getEncoded().length);
+    }
   }
 
   /** Claims {@link PortabilityJob} {@code jobId} and updates it with our public key in storage. */
```
rtannenbaum: Can you update doc here to reflect the new status returned?

portability-transfer/src/main/java/org/dataportabilityproject/transfer/JobPollingService.java
```diff
@@ -125,19 +130,20 @@ private void claimJob(UUID jobId, KeyPair keyPair) throws IOException {
           (previous, updated) ->
               Preconditions.checkState(
                   previous.jobAuthorization().state() == JobAuthorization.State.CREDS_AVAILABLE));
-    } catch (IllegalStateException e) {
-      throw new IOException(
+    } catch (IllegalStateException | IOException e) {
```
rtannenbaum: where is an IOException thrown?

portability-transfer/src/main/java/org/dataportabilityproject/transfer/JobPollingService.java
olsona: Done

portability-transfer/src/main/java/org/dataportabilityproject/transfer/JobPollingService.java
olsona: Not sure if it makes sense to log it at all, actually.  Should be clear from the other logs what happened.

portability-transfer/src/main/java/org/dataportabilityproject/transfer/JobPollingService.java
olsona: Done

portability-transfer/src/main/java/org/dataportabilityproject/transfer/JobPollingService.java
olsona: store.updateJob can throw an IOException.  Should an IOException be treated differently?

rtannenbaum: ah ok, my bad

olsona:
Fixes #376

Issue #400, opened 2018-05-01
Make job polling interval configurable


olsona:
It should be easier to configure the job polling frequency.  Right now you have to go into JobPollingService.java.

mohasaid:
I could try to fix it, which would be the best option to achieve job polling frequency value from? Gradle.properties?

burakfahri:
I can try to fix it, Mohasaid has asked for it but nobody has answered. Which would be the best option to poll frequency value? From Gradle.properties or under the resources or something else?

olsona:
Sure, if you can get it done in the next day or so that'd be pretty sweet!  We were thinking of making it environment-specific, so in a transfer.yaml file under resources/config would be great.

olsona:
#574

PR #401, opened 2018-05-02
Rogue newline fix

olsona:
Fixes bug where keys pulled by GoogleAppCredentialStore would have a newline appended, causing auth to fail.

### DIFF
--- a/extensions/cloud/portability-cloud-google/src/main/java/org/dataportabilityproject/cloud/google/GoogleAppCredentialStore.java
+++ b/extensions/cloud/portability-cloud-google/src/main/java/org/dataportabilityproject/cloud/google/GoogleAppCredentialStore.java
@@ -125,8 +125,6 @@ final class GoogleAppCredentialStore implements AppCredentialStore {
 
   private byte[] getRawBytes(String blobName) {
     Bucket bucket = storage.get(bucketName);
-    System.out.println("Blob name: " + blobName);
-    System.out.println("Bucket: " + bucketName);
     return bucket
         .get(blobName)
         .getContent();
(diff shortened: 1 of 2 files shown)


extensions/cloud/portability-cloud-google/src/main/java/org/dataportabilityproject/cloud/google/GoogleAppCredentialStore.java
```diff
@@ -125,15 +125,19 @@ public AppCredentials getAppCredentials(String keyName, String secretName) throw
 
   private byte[] getRawBytes(String blobName) {
     Bucket bucket = storage.get(bucketName);
-    return bucket.get(blobName).getContent();
+    System.out.println("Blob name: " + blobName);
```
holachuy: Let's remove the System.outs

portability-transfer/src/main/java/org/dataportabilityproject/transfer/JobPollingService.java
```diff
@@ -60,6 +60,8 @@ protected void runOneIteration() throws Exception {
     }
   }
 
+  // TODO: the delay should be more easily configurable
```
holachuy: Do you want this file in this PR or separate?

olsona: Separate, I made an issue #400

extensions/cloud/portability-cloud-google/src/main/java/org/dataportabilityproject/cloud/google/GoogleAppCredentialStore.java
olsona: Done

Issue #346, opened 2018-04-11
Complete TasksModel


seehamrun:
Currently this doesn't handle completion date/status (and probably other things).

olsona:
Related bugs: #429, #430, #432

lisad:
Not currently a high priority for anybody to work on.

PR #434, opened 2018-06-06
Adding task completion to importers

olsona:
Added task completion status and date to Google importer
Added task completion status (date is not available) to RTM importer
Renamed TaskAddResponse to TaskUpdateResponse, since the same format is used for responding to various task update requests

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-rememberthemilk/src/main/java/org/dataportabilityproject/transfer/rememberthemilk/tasks/RememberTheMilkService.java
+++ b/extensions/data-transfer/portability-data-transfer-rememberthemilk/src/main/java/org/dataportabilityproject/transfer/rememberthemilk/tasks/RememberTheMilkService.java
@@ -89,9 +89,8 @@ class RememberTheMilkService {
 
   public void completeTask(String timeline, String listId, int seriesId, int taskId)
       throws IOException {
-    // NB: This does not transfer the complete time!  There's no method for that.  It just sets a
-    // task as being complete, and I believe that the completion time is the time that the request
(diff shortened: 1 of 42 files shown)


olsona:
#346

extensions/data-transfer/portability-data-transfer-rememberthemilk/src/main/java/org/dataportabilityproject/transfer/rememberthemilk/tasks/RememberTheMilkService.java
```diff
@@ -83,8 +83,29 @@ public TaskSeries createTask(String name, String timeline, String listId) throws
...
-    TaskAddResponse taskAddResponse = makeRequest(params, TaskAddResponse.class);
-    return taskAddResponse.list.taskseries.get(0);
+    TaskUpdateResponse taskUpdateResponse = makeRequest(params, TaskUpdateResponse.class);
+    return taskUpdateResponse.list.taskseries.get(0);
+  }
+
+  public void completeTask(String timeline, String listId, int seriesId, int taskId)
+      throws IOException {
+    // NB: This does not transfer the complete time!  There's no method for that.  It just sets a
+    // task as being complete, and I believe that the completion time is the time that the request
```
holachuy: In other words, RTM does not support completion time in their api?

olsona: As far as I can tell, that is the case.

PR #436, opened 2018-06-14
Add demo support for import/export of derived data

jimmarino:
This adds demo support for transferring derived data.

### DIFF
--- a/extensions/auth/portability-auth-microsoft/src/main/java/org/dataportabilityproject/auth/microsoft/MicrosoftAuthDataGenerator.java
+++ b/extensions/auth/portability-auth-microsoft/src/main/java/org/dataportabilityproject/auth/microsoft/MicrosoftAuthDataGenerator.java
@@ -5,10 +5,6 @@ import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
-import java.io.IOException;
-import java.util.List;
-import java.util.Map;
-import java.util.function.Supplier;
 import okhttp3.MediaType;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
@@ -20,6 +16,11 @@ import org.dataportabilityproject.spi.api.auth.AuthServiceProviderRegistry.AuthM
 import org.dataportabilityproject.spi.api.types.AuthFlowConfiguration;
 import org.dataportabilityproject.types.transfer.auth.AuthData;
 import org.dataportabilityproject.types.transfer.auth.TokenAuthData;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Supplier;
 
 /**
  * Provides configuration for conducting an OAuth flow against the Microsoft AD API. Returned tokens
@@ -51,6 +52,7 @@ public class MicrosoftAuthDataGenerator implements AuthDataGenerator {
           .put("mail", ImmutableList.of("user.read", "Mail.Read"))
           .put("contacts", ImmutableList.of("user.read", "Contacts.Read"))
           .put("calendar", ImmutableList.of("user.read", "Calendars.Read"))
+          .put("derived-data", ImmutableList.of("user.read", "Files.Read.All"))
           .build();
 
   private final String redirectPath;
--- a/extensions/auth/portability-auth-microsoft/src/main/java/org/dataportabilityproject/auth/microsoft/MicrosoftAuthDataGenerator.java
+++ b/extensions/auth/portability-auth-microsoft/src/main/java/org/dataportabilityproject/auth/microsoft/MicrosoftAuthDataGenerator.java
@@ -52,6 +52,7 @@ public class MicrosoftAuthDataGenerator implements AuthDataGenerator {
           .put("mail", ImmutableList.of("user.read", "Mail.Read"))
           .put("contacts", ImmutableList.of("user.read", "Contacts.Read"))
           .put("calendar", ImmutableList.of("user.read", "Calendars.Read"))
+          // needed because derived exported as one drive files
           .put("derived-data", ImmutableList.of("user.read", "Files.Read.All"))
           .build();
 
--- a/extensions/data-transfer/portability-data-transfer-derived-demo/src/main/java/org/dataportabilityproject/transfer/derived/DerivedDemoImporter.java
+++ b/extensions/data-transfer/portability-data-transfer-derived-demo/src/main/java/org/dataportabilityproject/transfer/derived/DerivedDemoImporter.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2018 The Data-Portability Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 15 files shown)


extensions/auth/portability-auth-microsoft/src/main/java/org/dataportabilityproject/auth/microsoft/MicrosoftAuthDataGenerator.java
```diff
@@ -51,6 +52,7 @@
           .put("mail", ImmutableList.of("user.read", "Mail.Read"))
           .put("contacts", ImmutableList.of("user.read", "Contacts.Read"))
           .put("calendar", ImmutableList.of("user.read", "Calendars.Read"))
+          .put("derived-data", ImmutableList.of("user.read", "Files.Read.All"))
```
rtannenbaum: The Files.Read.All isn't obvious to me, can you document why that scope is used for derived data?

extensions/data-transfer/portability-data-transfer-derived-demo/src/main/java/org/dataportabilityproject/transfer/derived/DerivedDemoImporter.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.dataportabilityproject.spi.transfer.provider.ImportResult;
+import org.dataportabilityproject.spi.transfer.provider.Importer;
+import org.dataportabilityproject.transfer.microsoft.spi.types.MicrosoftDerivedData;
+import org.dataportabilityproject.types.transfer.auth.TokenAuthData;
+
+import java.util.UUID;
+
+/** Simulates importing derived data. For demo purposes only! */
+public class DerivedDemoImporter implements Importer<TokenAuthData, MicrosoftDerivedData> {
```
rtannenbaum: if it's not too much work, maybe the data type should be generic and not microsoft, since this is a demo extension. e.g. DemoDerivedData

extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/dataportabilityproject/transfer/microsoft/MicrosoftTransferExtension.java
```diff
@@ -11,25 +11,35 @@
...
 import org.dataportabilityproject.transfer.microsoft.contacts.MicrosoftContactsImporter;
+import org.dataportabilityproject.transfer.microsoft.derived.MicrosoftDerivedDataExporter;
 import org.dataportabilityproject.transfer.microsoft.photos.MicrosoftPhotosExporter;
 import org.dataportabilityproject.transfer.microsoft.photos.MicrosoftPhotosImporter;
 import org.dataportabilityproject.transfer.microsoft.transformer.TransformerService;
 import org.dataportabilityproject.transfer.microsoft.transformer.TransformerServiceImpl;
 
+/**
+ * Bootstraps the Microsoft data transfer services. Note the format of the exported contents are
+ * opaque; they may change without notice.
```
rtannenbaum: is this comment true of all data or just derived?

extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/dataportabilityproject/transfer/microsoft/MicrosoftTransferExtension.java
```diff
@@ -55,6 +65,10 @@ public String getServiceId() {
       return new MicrosoftPhotosExporter(BASE_GRAPH_URL, client, mapper, jobStore);
     }
 
+    if (derivedData && transferDataType.equals(DERIVED_DATA)) {
```
rtannenbaum: can you add a comment why the flag is here for derived data and not the other verticals?

extensions/auth/portability-auth-microsoft/src/main/java/org/dataportabilityproject/auth/microsoft/MicrosoftAuthDataGenerator.java
jimmarino: Sure

extensions/data-transfer/portability-data-transfer-derived-demo/src/main/java/org/dataportabilityproject/transfer/derived/DerivedDemoImporter.java
jimmarino: One thing is derived data is specific to the provider and probably won't have any data in common with other providers. We could have a placeholder type but it wouldn't have any data specified in it. The exporting extension would still need to supply a specific type that is understood by the import, so import extensions will still be dependent on the exported type.

extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/dataportabilityproject/transfer/microsoft/MicrosoftTransferExtension.java
jimmarino: Good catch, that's not true. I moved it out to the derived data exporter.

extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/dataportabilityproject/transfer/microsoft/MicrosoftTransferExtension.java
jimmarino: Sure

rtannenbaum: sorry wasn't clear, I meant why derived needs a flag but the other data types don't?

extensions/data-transfer/portability-data-transfer-derived-demo/src/main/java/org/dataportabilityproject/transfer/derived/DerivedDemoImporter.java
rtannenbaum: Placeholder is fine with me, or if you could just add a comment that Microsoft is used since that's the only one we have now, or something like that. Thanks!

jimmarino: OK I added a comment in there.

extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/dataportabilityproject/transfer/microsoft/MicrosoftTransferExtension.java
jimmarino: For now, I think so. I added a TODO about potentially adding a configuration option that allows builders/deployers to disable (or enable) transfer of certain data types. For example, "baseline" data types could be turned on by default while "experimental/optional" could be off.

It looks as if the number of config switches is quickly multiplying and it may be getting unwieldy to figure out how to specify all of them. One way around this would be to create a Gradle plugin that creates a configured build artifact. For example, someone would use the plugin in their own Gradle build to handle all of the configuration setup for the runtime. The Gradle build would have a project in it that depended on the relevant data portability modules and would have plugins applied to generate Docker images. In addition, the project would apply the "data portability plugin" and generate all of the required YAML, command line settings, etc. The plugin would provide one typed way to view and specify runtime settings. 

What do you think? That's probably something we should all discuss and think about in relation to the current config.

Thanks!

PR #437, opened 2018-06-18
Getting started on simple error handling

olsona:
If an export call fails, we check the error message to see whether the error is retryable or not, and how many retries make sense.


portability-transfer/src/main/java/org/dataportabilityproject/transfer/PortabilityInMemoryDataCopier.java
```diff
@@ -42,6 +43,9 @@
   private static final AtomicInteger COPY_ITERATION_COUNTER = new AtomicInteger();
   private static final Logger logger = LoggerFactory.getLogger(PortabilityInMemoryDataCopier.class);
 
+  private static final List<String> fatalRegexes = ImmutableList.of("*fatal*");
```
rtannenbaum: can fatalRegexes (the var name) be more descriptive? as a member of InMemoryDataCopier, I don't think there is much context

portability-transfer/src/main/java/org/dataportabilityproject/transfer/PortabilityInMemoryDataCopier.java
```diff
@@ -42,6 +43,9 @@
...
+  private static final int maxRetriesInit = 5;
```
rtannenbaum: seems weird that this would be called maxRetries"Init" when it never changes

portability-transfer/src/main/java/org/dataportabilityproject/transfer/PortabilityInMemoryDataCopier.java
```diff
@@ -121,16 +122,72 @@ private void copyHelper(
...
+              Optional.of(new ExportInformation(null, resource)));
+        }
+      }
+    }
+  }
+
+  private ExportResult runExportLogic(UUID jobId, AuthData exportAuthData,
+      Optional<ExportInformation> exportInformation) {
+    ExportResult<?> exportResult = exportHelper(jobId, exportAuthData, exportInformation);
+    if (exportResult.getType() == ResultType.ERROR) {
```
rtannenbaum: is it possible to simplify this to consolidate the checking exportResult.getType multiple times and the manual increment of retries? something like:

resultType = get initial result . type;
for (int attempt = 0; resultType == error; attempt++) {
   resultType = retried result.type
}

portability-transfer/src/main/java/org/dataportabilityproject/transfer/PortabilityInMemoryDataCopier.java
```diff
@@ -121,16 +122,72 @@ private void copyHelper(
...
         for (ContainerResource resource : continuationData.getContainerResources()) {
-          copyHelper(jobId, exportAuthData, importAuthData, Optional.of(new ExportInformation(null, resource)));
+          copyHelper(jobId, exportAuthData, importAuthData,
+              Optional.of(new ExportInformation(null, resource)));
+        }
+      }
+    }
+  }
+
+  private ExportResult runExportLogic(UUID jobId, AuthData exportAuthData,
```
rtannenbaum: how about just export() ?

portability-transfer/src/main/java/org/dataportabilityproject/transfer/PortabilityInMemoryDataCopier.java
```diff
@@ -121,16 +122,72 @@ private void copyHelper(
...
+
+  private ExportResult exportHelper(UUID jobId, AuthData exportAuthData,
+      Optional<ExportInformation> exportInformation) {
+    logger.debug("Starting export");
+    ExportResult<?> exportResult = exporter.get().export(jobId, exportAuthData, exportInformation);
+    logger.debug("Finishing export");
+    return exportResult;
   }
+
+  private ExceptionResponse checkRetry(String exceptionMessage) {
```
rtannenbaum: can we also check based on http status code?

portability-transfer/src/main/java/org/dataportabilityproject/transfer/PortabilityInMemoryDataCopier.java
```diff
@@ -121,16 +121,69 @@ private void copyHelper(
...
+  private ExportResult runExportLogic(UUID jobId, AuthData exportAuthData,
+      Optional<ExportInformation> exportInformation) {
+    ExportResult<?> exportResult = exportHelper(jobId, exportAuthData, exportInformation);
+    if (exportResult.getType() == ResultType.ERROR) {
+      ExceptionResponse response = checkRetry(exportResult.getMessage());
+      if (response.canRetry) {
+        int attempts = 1;
+        // TODO: what if different kinds of errors have different max retries, and we start with one
+        // kind of error and then start seeing another?
+        while (attempts < response.maxRetries && response.canRetry) {
```
rtannenbaum: can all the work to determine "retryable" (attempts plus error checking) be in its own method, either called here or in ExceptionResponse? I think that will help generalize later to be in a callable

portability-transfer/src/main/java/org/dataportabilityproject/transfer/PortabilityInMemoryDataCopier.java
```diff
@@ -121,16 +121,69 @@ private void copyHelper(
...
+  private ExceptionResponse checkRetry(String exceptionMessage) {
+    for (String fatalRegex : fatalRegexes) {
+      if (exceptionMessage.matches(fatalRegex)) {
+        return new ExceptionResponse(false, 0);
+      }
+    }
+    return new ExceptionResponse(true, 5);
+  }
+
+  private class ExceptionResponse {
```
rtannenbaum: can the name ExceptionResponse be more descriptive?

portability-transfer/src/main/java/org/dataportabilityproject/transfer/PortabilityInMemoryDataCopier.java
```diff
@@ -121,16 +121,69 @@ private void copyHelper(
...
+          }
         }
       }
     }
+    return exportResult;
+  }
+
+  private ExportResult exportHelper(UUID jobId, AuthData exportAuthData,
+      Optional<ExportInformation> exportInformation) {
+    logger.debug("Starting export");
```
rtannenbaum: if you're going to log, I think it would be more useful to include some more info

portability-transfer/src/main/java/org/dataportabilityproject/transfer/PortabilityInMemoryDataCopier.java
olsona: That's much cleaner, thank you!  Forgotten you could do that.

portability-transfer/src/main/java/org/dataportabilityproject/transfer/PortabilityInMemoryDataCopier.java
olsona: Done.

portability-transfer/src/main/java/org/dataportabilityproject/transfer/PortabilityInMemoryDataCopier.java
olsona: Is fatalErrorRegexes sufficiently descriptive?

portability-transfer/src/main/java/org/dataportabilityproject/transfer/PortabilityInMemoryDataCopier.java
olsona: I think that should wait until v1.

portability-transfer/src/main/java/org/dataportabilityproject/transfer/PortabilityInMemoryDataCopier.java
```diff
@@ -121,16 +124,57 @@ private void copyHelper(
...
+              Optional.of(new ExportInformation(null, resource)));
+        }
+      }
+    }
+  }
+
+  private ExportResult export(UUID jobId, AuthData exportAuthData,
+      Optional<ExportInformation> exportInformation) {
+    ExportResult<?> exportResult = exporter.get().export(jobId, exportAuthData, exportInformation);
+    for (int attempts = 0; exportResult.getType() == ResultType.ERROR; attempts++) {
```
rtannenbaum: can all this logic be moved into a helper method called validateResult() or retryIfNecessary() or something to that effect? I think that makes it more readable to not have the majority of the export() function be error handling

portability-transfer/src/main/java/org/dataportabilityproject/transfer/PortabilityInMemoryDataCopier.java
```diff
@@ -121,16 +124,57 @@ private void copyHelper(
...
+      ExceptionType response = checkRetry(exportResult.getMessage(), attempts);
+      if (response.canRetry) {
```
rtannenbaum: Can you make canRetry() a method so we don't directly access a member var?

portability-transfer/src/main/java/org/dataportabilityproject/transfer/PortabilityInMemoryDataCopier.java
```diff
@@ -121,16 +124,57 @@ private void copyHelper(
...
+        exportResult = exporter.get().export(jobId, exportAuthData, exportInformation);
+        if (exportResult.getType() != ResultType.ERROR) {
```
rtannenbaum: is this check still needed, won't we terminate in the for loop condition?

portability-transfer/src/main/java/org/dataportabilityproject/transfer/PortabilityInMemoryDataCopier.java
```diff
@@ -121,16 +124,57 @@ private void copyHelper(
...
+        }
+      }
+    }
+  }
+
+  private ExportResult export(UUID jobId, AuthData exportAuthData,
+      Optional<ExportInformation> exportInformation) {
+    ExportResult<?> exportResult = exporter.get().export(jobId, exportAuthData, exportInformation);
+    for (int attempts = 0; exportResult.getType() == ResultType.ERROR; attempts++) {
+      ExceptionType response = checkRetry(exportResult.getMessage(), attempts);
```
rtannenbaum: is ExceptionType response used for anything besides checking whether to retry? 

If not, I'm wondering if checkRetry can just return a boolean instead

olsona: That's what I was thinking.  I believe that ultimately something like ExceptionType will also be holding retries, but I think that can wait until v1.

portability-transfer/src/main/java/org/dataportabilityproject/transfer/PortabilityInMemoryDataCopier.java
olsona: Done.

portability-transfer/src/main/java/org/dataportabilityproject/transfer/PortabilityInMemoryDataCopier.java
olsona: Done.

portability-transfer/src/main/java/org/dataportabilityproject/transfer/PortabilityInMemoryDataCopier.java
olsona: I feel like it's simple enough to read now that I implemented your other suggestions.  What do you think?

portability-transfer/src/main/java/org/dataportabilityproject/transfer/PortabilityInMemoryDataCopier.java
```diff
@@ -42,6 +43,9 @@
   private static final AtomicInteger COPY_ITERATION_COUNTER = new AtomicInteger();
   private static final Logger logger = LoggerFactory.getLogger(PortabilityInMemoryDataCopier.class);
 
+  private static final List<String> fatalErrorRegexes = ImmutableList.of("*fatal*"); // TODO: make configurable
```
rtannenbaum: I think var name should still be capitalized since it's a private static final, if we want to follow google java style guide. like how COPY_ITERATION_COUNTER is cap'd.

portability-transfer/src/main/java/org/dataportabilityproject/transfer/PortabilityInMemoryDataCopier.java
rtannenbaum: yep, looks good

portability-transfer/src/main/java/org/dataportabilityproject/transfer/PortabilityInMemoryDataCopier.java
```diff
@@ -121,16 +124,42 @@ private void copyHelper(
...
+      }
+    }
+    return exportResult;
+  }
+
+  private boolean checkCanRetry(String exceptionMessage, int attempts) {
+    if (attempts >= maxAttempts) {
+      return false;
+    }
+    for (String fatalRegex : fatalErrorRegexes) {
```
holachuy: I would add a comment for clarity.

PR #442, opened 2018-06-25
First round of error handling, retry strategies

olsona:
Tried to implement simpler versions of RetryingCallable and RetryStrategy.  Shifted logic around so that RetryStrategy does a lot of computation of when things should be retried and when they shouldn't.


portability-transfer/src/main/java/org/dataportabilityproject/transfer/RetryStrategy.java
```diff
@@ -0,0 +1,72 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.dataportabilityproject.transfer;
+
+public abstract class RetryStrategy {
+
+  public boolean canTryAgain(Exception e, int tries) {
```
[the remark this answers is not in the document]
olsona: Done.

bwillard:
Looking pretty close, just a couple of small comments.

portability-spi-transfer/src/main/java/org/dataportabilityproject/spi/transfer/provider/ExportResult.java
```diff
@@ -1,23 +1,32 @@
...
+import com.google.common.base.Preconditions;
 import org.dataportabilityproject.spi.transfer.types.ContinuationData;
 import org.dataportabilityproject.types.transfer.models.DataModel;
 
-/** The result of an item export operation. */
+/**
+ * The result of an item export operation.
+ */
 public class ExportResult<T extends DataModel> {
+
```
rtannenbaum: nit: extra newlines at the beginning of classes?

portability-spi-transfer/src/main/java/org/dataportabilityproject/spi/transfer/provider/ExportResult.java
```diff
@@ -1,23 +1,32 @@
...
+  private Throwable throwable;
+
   /**
    * Ctor used to return error or retry results.
    *
    * @param type the result type
    * @param message the result message, if any
    */
   public ExportResult(ResultType type, String message) {
+    Preconditions.checkArgument(!type.equals(ResultType.ERROR), MUST_HAVE_THROWABLE);
```
rtannenbaum: since this is done a few times, I'd move this (and the string constant) to a new method that explains what we're doing / why in its name, like "verifyNonErrorResult" or "validateResultType"

portability-spi-transfer/src/main/java/org/dataportabilityproject/spi/transfer/provider/ExportResult.java
```diff
@@ -1,23 +1,32 @@
...
+
   public static final ExportResult CONTINUE = new ExportResult(ResultType.CONTINUE);
   public static final ExportResult END = new ExportResult(ResultType.CONTINUE);
+  private static final String MUST_HAVE_THROWABLE = "ExportResult with ResultType = ERROR must hold a throwable";
+
   private ResultType type;
   private String message;
   private T exportedData;
   private ContinuationData continuationData;
+  private Throwable throwable;
```
rtannenbaum: nit: can you document that this should be null if nothing was thrown during the export?

portability-spi-transfer/src/main/java/org/dataportabilityproject/spi/transfer/provider/ExportResult.java
```diff
@@ -1,23 +1,32 @@
 package org.dataportabilityproject.spi.transfer.provider;
 
+import com.google.common.base.Preconditions;
 import org.dataportabilityproject.spi.transfer.types.ContinuationData;
 import org.dataportabilityproject.types.transfer.models.DataModel;
 
-/** The result of an item export operation. */
+/**
+ * The result of an item export operation.
```
rtannenbaum: nit: this is after all retries, right? can you document that?

portability-spi-transfer/src/main/java/org/dataportabilityproject/spi/transfer/provider/ExportResult.java
```diff
@@ -74,19 +102,26 @@ public Object getContinuationData() {
...
+     * Indicates the operation was successful and no more items are available.
+     */
     END,
-    /** Indicates an unrecoverable error was raised. */
-    ERROR,
-
-    /** Indicates a recoverable error was raised. */
-    RETRY
+    /**
+     * Indicates an unrecoverable error was raised.
```
rtannenbaum: does this also include the case where all retries have failed? if so, I think that should be documented, i.e. unrecoverable error or all retries failed

portability-transfer/src/main/java/org/dataportabilityproject/transfer/CallableExporter.java
```diff
@@ -0,0 +1,49 @@
...
+
+import com.google.inject.Provider;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.concurrent.Callable;
+import org.dataportabilityproject.spi.transfer.provider.ExportResult;
+import org.dataportabilityproject.spi.transfer.provider.Exporter;
+import org.dataportabilityproject.spi.transfer.types.ExportInformation;
+import org.dataportabilityproject.types.transfer.auth.AuthData;
+
```
rtannenbaum: nit: documentation for public classes

portability-transfer/src/main/java/org/dataportabilityproject/transfer/PortabilityInMemoryDataCopier.java
```diff
@@ -86,31 +90,45 @@ private void copyHelper(
...
     // NOTE: order is important below, do the import of all the items, then do continuation
     // then do sub resources, this ensures all parents are populated before children get
     // processed.
     logger.debug("Starting export");
-    ExportResult<?> exportResult = export(jobId, exportAuthData, exportInformation);
-    logger.debug("Finished export");
-
-    if (exportResult.getType().equals(ResultType.ERROR)) {
-      logger.warn("Error happened during export: {}", exportResult.getMessage());
+    CallableExporter callableExporter = new CallableExporter(exporter, jobId, exportAuthData,
```
rtannenbaum: nit: As far as this copier is concerned it wants a retrying exporter, retrying importer, etc. So I'd move this retrying logic into the CallableExporter class and call it RetryingExporter.

portability-transfer/src/main/java/org/dataportabilityproject/transfer/PortabilityInMemoryDataCopier.java
```diff
@@ -86,31 +90,45 @@ private void copyHelper(
...
+        library,
+        Clock.systemUTC());
+    ExportResult<?> exportResult;
+    try {
+      exportResult = exportRetryingCallable.call();
+    } catch (RetryException e) {
+      logger.warn("Error happened during export: {}", e);
       return;
     }
+    logger.debug("Finished export");
```
rtannenbaum: If we aggregate these logs and don't look up individual worker logs one by one, it would be much more useful to include the job ID in all log statements

portability-transfer/src/main/java/org/dataportabilityproject/transfer/RetryException.java
```diff
@@ -0,0 +1,35 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.dataportabilityproject.transfer;
+
+public class RetryException extends Exception {
+  private final int attempts;
```
rtannenbaum: can this be documented? i.e. I'm wondering does attempts represent the # we've already made, the number we're allowed to make, etc?

portability-transfer/src/main/java/org/dataportabilityproject/transfer/RetryStrategy.java
```diff
@@ -0,0 +1,93 @@
...
+  public long getRemainingIntervalMillis(int tries, long elapsedMillis) {
+    long intervalMillis = getNextIntervalMillis(tries);
+    Preconditions.checkArgument(intervalMillis >= 0, "No retries left");
+    return intervalMillis - elapsedMillis;
+  }
+
+  /**
+   * {@link RetryStrategy} that allows retries on regular intervals
+   */
+  public static class SimpleRetryStrategy extends RetryStrategy {
```
rtannenbaum: nice Retry strategies! although a little weird for them to extend the abstract class inside that class's definition, I haven't seen that before. Can they each be their own class in a retry package perhaps? and I think RetryStrategy can be an interface since no member vars

portability-transfer/src/main/java/org/dataportabilityproject/transfer/RetryingCallable.java
```diff
@@ -0,0 +1,73 @@
...
+package org.dataportabilityproject.transfer;
+
+import static java.lang.Thread.currentThread;
+
+import java.time.Clock;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.concurrent.Callable;
+
+public class RetryingCallable<T> implements Callable<T> {
```
rtannenbaum: I think this can be package private

portability-transfer/src/main/java/org/dataportabilityproject/transfer/RetryStrategy.java
olsona: Done.

portability-transfer/src/main/java/org/dataportabilityproject/transfer/RetryingCallable.java
olsona: Not if it's in another package.  What's more important?

portability-transfer/src/main/java/org/dataportabilityproject/transfer/RetryException.java
olsona: Should be clearer now.

portability-transfer/src/main/java/org/dataportabilityproject/transfer/PortabilityInMemoryDataCopier.java
olsona: Wouldn't that result in a lot of code duplication, since the same logic is used in CallableExporter and CallableImporter?

portability-transfer/src/main/java/org/dataportabilityproject/transfer/PortabilityInMemoryDataCopier.java
olsona: Done.

portability-spi-transfer/src/main/java/org/dataportabilityproject/spi/transfer/provider/ExportResult.java
olsona: Done.

portability-transfer/src/main/java/org/dataportabilityproject/transfer/CallableExporter.java
olsona: Got started on documentation.  Will expand it as I continue to work.

portability-transfer/src/main/java/org/dataportabilityproject/transfer/RetryingCallable.java
```diff
@@ -0,0 +1,72 @@
...
+  public T call() throws RetryException {
+    Instant start = clock.instant();
+    while (true) {
+      attempts++;
+      try {
+        return callable.call();
+      } catch (InterruptedException e) {
+        currentThread().interrupt();
+        throw new RetryException(attempts, mostRecentException != null ? mostRecentException : e);
+      } catch (Exception e) {
```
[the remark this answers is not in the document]
olsona: This should be better now.

portability-spi-transfer/src/main/java/org/dataportabilityproject/spi/transfer/provider/ExportResult.java
```diff
@@ -1,23 +1,31 @@
...
 public class ExportResult<T extends DataModel> {
+
   public static final ExportResult CONTINUE = new ExportResult(ResultType.CONTINUE);
   public static final ExportResult END = new ExportResult(ResultType.CONTINUE);
+
   private ResultType type;
   private String message;
   private T exportedData;
   private ContinuationData continuationData;
+  private Throwable throwable; // Should be null unless an error was thrown during export
```
rtannenbaum: actually, what do you think of using an Optional?

portability-spi-transfer/src/main/java/org/dataportabilityproject/spi/transfer/provider/ExportResult.java
rtannenbaum: is it possible to change the setting in your IDE?

portability-spi-transfer/src/main/java/org/dataportabilityproject/spi/transfer/provider/ExportResult.java
rtannenbaum: oh I see. It seems odd to store the result in an ExportResult if it's a success or unrecoverable error, but not retried error. I would have ExportResult store all results (success and all failure modes).

That is also something that could be useful to store (in a db) later on.

portability-transfer/src/main/java/org/dataportabilityproject/transfer/PortabilityInMemoryDataCopier.java
rtannenbaum: you could refactor to a common place, just didn't think it needed to be a part of the copier, seems like it belongs more in the Callable*porter's

rtannenbaum: seems like now there is some duplication in the copier, and it's stuff the copier doesn't need to know about

PR #443, opened 2018-06-26
Error handling - cleaning up constructors

olsona:


portability-spi-transfer/src/main/java/org/dataportabilityproject/spi/transfer/provider/ImportResult.java
```diff
@@ -32,7 +20,7 @@ public ImportResult(ResultType type, String message) {
    */
   public ImportResult(Throwable throwable) {
     this.type = ResultType.ERROR;
-    this.throwable = throwable;
+    this.throwable = Optional.of(throwable);
   }
 
   /**
```
rtannenbaum: Cool, would just initialize the throwable to absent

PR #444, opened 2018-06-28
Switch to new client-rest UI and remove old UI

jimmarino:


seehamrun:
Trying to bring up the new rest API - is there any new documentation on how to do this?

jimmarino:
The build should work exactly the same as before. You can launch the client either using one of the Gradle tasks or from the command line:

 ng serve --ssl --port 3000 --proxy-config proxy.conf.json

The command line changed slightly since it only uses HTTPS and not HTTP.

Let me know if you run into an issue getting it started.

seehamrun:
Since this is not entirely complete yet, lets hold off on deleting the old one. 

A couple of things I noticed: 

* Only oauth2 is implemented. 
  * flickr and smugmug use oauth1
  * remember the milk has a legacy auth handler
* Looks like somewhere along the line import and export services are flipped.  Example: when doing a photos transfer the list of export services and import services looks to be flipped around (instagram is an export only service). 
image

rtannenbaum:
Hey @jimmarino I figured this PR was probably stale so I made this one, https://github.com/google/data-transfer-project/pull/503. I'll remove the old oauth handlers next in a separate PR.

Issue #449, opened 2018-06-30
Rework local secrets to support a publishable docker image


bwillard:
We would like to have a docker image we can publish that folks can just add their own API keys to to try out DTP.

Right now for local deployments we bake the secrets into the JAR, so we couldn't publish the JAR.

A better way would be to publish a docker image with no secrets, and then have an easy way to use our image a a base and add in your own secrets/api keys and launch the image.

PR #454, opened 2018-07-06
Switch how secrets for local/demo-server deployments are handled.

bwillard:
This addresses #449 and will allow us to start publishing a docker image to docker hub so folks can run the demo without building anything.


distributions/README.md
```diff
@@ -9,9 +9,12 @@ We welcome contributors to add configuration for additional platforms.
 
 ## Local development
 
-For local development (only), secrets may be stored in gradle properties.
-Add any test secrets to your local `.gradle/dataportability.secrets.properties` file. A
-sample is provided in `distributions/common/dataportability.secrets.example.properties`.
+For local development (only), secrets may be stored in environment variables.
```
holachuy: Maybe a quick example. We also have to ensure this works for the windows case

distributions/demo-server/.gitignore
```diff
@@ -0,0 +1 @@
+env.secrets
```
holachuy: Should we add **/env.secrets in case folks put it somewhere deeper in the directory?

distributions/demo-server/env.secrets.template
```diff
@@ -0,0 +1,31 @@
+# see https://docs.docker.com/compose/env-file/
+# copy this file to env.secrets and fill in your own app credentials, see:
```
holachuy: should we put the directory in here too

distributions/demo-server/.gitignore
bwillard: Sure done.

distributions/demo-server/env.secrets.template
bwillard: Done.

distributions/README.md
bwillard: Done.

Issue #461, opened 2018-07-09
Support credential handling in client


holachuy:
Support Create Job, Reserve Worker, StartJob, LookupJob endpoints for

PR #468, opened 2018-07-10
Rename DataTransfer[Request/Response]->CreateJob[Request/Response]

rtannenbaum:
Resolves some TODOs

This might change soon for #461, but I didn't want to confuse myself in the process of reworking the CreateJob endpoint

Tested a photos transfer with demo-server

### DIFF
--- a/portability-types-client/src/test/java/org/dataportabilityproject/types/client/transfer/DataGetTransferTest.java
+++ b/portability-types-client/src/test/java/org/dataportabilityproject/types/client/transfer/DataGetTransferTest.java
@@ -20,7 +20,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import org.junit.Assert;
 import org.junit.Test;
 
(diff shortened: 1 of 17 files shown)


portability-types-client/src/test/java/org/dataportabilityproject/types/client/transfer/DataGetTransferTest.java
```diff
@@ -20,7 +20,7 @@
 import org.junit.Assert;
 import org.junit.Test;
 
-/* Test for DataTransferRequest */
+/* Test for CreateJobRequest */
 public class DataGetTransferTest {
```
holachuy: Update the class name too?

rtannenbaum:
I re-did this in https://github.com/google/data-transfer-project/pull/471, merging in the package rename was too big a pain in this PR.

I got the bad filename in there too, thanks

PR #474, opened 2018-07-16
Configure Travis to deploy new docker images to Docker hub every merge to master

bwillard:


docker_push.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+#Script used to push new docker images to docker hug by travis.
+
+./gradlew -PcloudType=local :distributions:demo-server:dockerize
+echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
```
rtannenbaum: can you document how these env variables are set?

holachuy:
Does push to docker hub mean each commit is a published version, can we distinguish between latest and an RC or prod version?

Just hate for folks to pull and run latest when it could be broken.

docker_push.sh
bwillard: Done.

rtannenbaum:
@bwillard I think we need to distinguish between push/commit and merge though, for example a PR can contain 10 commits but only 1 merge into master. Is it possible to make it only build the new image on merge PR to master? We probably don't need it for every commit

bwillard:
@rtannenbaum sorry, yes that is what this does is only run the docker push on the merge to master, not on every commit.

rtannenbaum:
Ah perfect, thanks. Can you update PR description? LGTM

PR #479, opened 2018-07-18
Switch to new UI

jimmarino:
This PR supersedes the previous one. It switches the build to default to the new UI; adds support for OAuth1 and Legacy Auth in new UI; and fixes a bug where export and import services were transposed.

Note that multiple callback protocols are handled by the same Angular auth component on the client so all extensions may use the same "/callback" base URL. If an extension uses a completely custom auth protocol, the UI can be extended with a custom URL.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,6 +55,8 @@ testem.log
 /e2e/*.js
 /e2e/*.map
 
+client-rest/dist/
+
 # System Files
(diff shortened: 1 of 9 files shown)


jimmarino:
Thanks. @seehamrun when you get a chance, can you try it out to make sure the issues you found are corrected?

PR #483, opened 2018-07-19
Start on logging

olsona:
Right now, this updates the console logger to use the new EncryptingLayout.  The EncryptingLayout doesn't encrypt yet, but it does prepend every log statement with a jobId.

The logger update happens within JobProcessor.  It could happen earlier, in the initialization of JobMetadata, but I feel a little more comfortable with the update at this point so we can be sure that everything is working (i.e., the assignment of a worker to the jobId).  

We could, I believe, only encrypt a FileLogger and leave the ConsoleLogger as is, i.e., unencrypted.  That would make debugging in the moment easier.  

Next step is to add encryption to the EncryptingLayout.


portability-transfer/src/main/java/org/datatransferproject/transfer/logging/EncryptingLayout.java
```diff
@@ -0,0 +1,32 @@
...
+  private final UUID jobId;
+
+  public EncryptingLayout(UUID jobId) {
+    this.jobId = jobId;
+  }
+
+  @Override
+  public String format(LoggingEvent event) {
+    // NB: copied from SimpleLayout.format()
+    return String.format("[%s] [%s]: %s - %s%s", new ISO8601DateFormat().format(event.timeStamp),
```
bwillard: Can this be a pattern that is passed in instead of hard coded?

portability-transfer/src/main/java/org/datatransferproject/transfer/JobProcessor.java
```diff
@@ -129,4 +133,15 @@ private AuthData deSerialize(String serialized) throws IOException {
       throw new IOException("Unable to deserialize AuthData", e);
     }
   }
+
+  private void setupConsoleLogger(UUID jobId) {
```
bwillard: I think hard coding the layout/appender is the wrong way to go, it should come from the config file.

portability-api-launcher/build.gradle
```diff
@@ -0,0 +1,3 @@
+dependencies {
+    compile("org.slf4j:slf4j-log4j12:${slf4jVersion}")
```
holachuy: I'm wondering if we need this in the api-launcher API launcher build file.  Isn't it only setup to work in the TransferWorker for now?

portability-transfer/src/main/java/org/datatransferproject/transfer/logging/EncryptingLayout.java
```diff
@@ -0,0 +1,32 @@
...
+import java.util.UUID;
+import org.apache.log4j.Layout;
+import org.apache.log4j.helpers.ISO8601DateFormat;
+import org.apache.log4j.spi.LoggingEvent;
+
+public class EncryptingLayout extends Layout {
+
+  static UUID jobId;
+
+  public static void setJobId(UUID inputJobId) {
```
holachuy: Let's check that it is not already set and generate an error if it is attempted to be set twice

portability-transfer/src/main/java/org/datatransferproject/transfer/WorkerModule.java
```diff
@@ -85,6 +83,7 @@ protected void configure() {
     bind(AsymmetricKeyGenerator.class).toInstance(asymmetricKeyGenerator);
     bind(InMemoryDataCopier.class).to(PortabilityInMemoryDataCopier.class);
     bind(ObjectMapper.class).toInstance(context.getTypeManager().getMapper());
+    bind(Layout.class).toInstance(context.getLayout());
```
holachuy: If we're using the EncryptingLayout each time, should just we just keep it out of the context alltogether?

portability-api-launcher/build.gradle
olsona: I needed it in there for ExtensionContext.  I agree it feels weird in there.

portability-transfer/src/main/java/org/datatransferproject/transfer/WorkerModule.java
olsona: We're not going to be, I just haven't settled on a way to inject it from the config yet.

portability-transfer/src/main/java/org/datatransferproject/transfer/logging/EncryptingLayout.java
olsona: Do you mean different kinds of patterns for EncryptingLayout, or different kinds of patterns for different layouts?

portability-transfer/src/main/java/org/datatransferproject/transfer/logging/EncryptingLayout.java
olsona: Done.

portability-api/src/main/java/org/datatransferproject/api/ApiExtensionContext.java
```diff
@@ -58,7 +59,7 @@ public ApiExtensionContext(TypeManager typeManager, SettingsExtension settingsEx
   }
 
   @Override
-  public Logger getLogger() {
+  public Layout getLayout() {
```
bwillard: Why do we need to expose this to clients?  Seems like this is an internal implementation detail.

portability-transfer/src/main/java/org/datatransferproject/transfer/JobProcessor.java
```diff
@@ -129,4 +137,15 @@ private AuthData deSerialize(String serialized) throws IOException {
       throw new IOException("Unable to deserialize AuthData", e);
     }
   }
+
+  private void setupConsoleLogger(Layout layout) {
```
bwillard: I don't think we should be setting up the appender in code, why do we need to?

portability-transfer/src/main/java/org/datatransferproject/transfer/WorkerMain.java
```diff
@@ -26,6 +26,8 @@
 import com.google.inject.Injector;
 import java.util.List;
 import java.util.ServiceLoader;
+import java.util.UUID;
```
bwillard: Do you still need these changes?

portability-transfer/src/main/java/org/datatransferproject/transfer/WorkerModule.java
bwillard: Ya I am confused why we need to build this at all...

portability-transfer/src/main/java/org/datatransferproject/transfer/logging/EncryptingLayout.java
bwillard: I mean this kind of string seems like it should come from the log4j.properties file, but that be left as a TODO

olsona: It can't be in the log4j properties file, since that pattern layout doesn't have room for arbitrary variables.  But it can go somewhere else.

portability-api/src/main/java/org/datatransferproject/api/ApiExtensionContext.java
```diff
@@ -16,6 +16,7 @@
 package org.datatransferproject.api;
 
 import com.google.common.base.Preconditions;
+import org.apache.log4j.Layout;
```
bwillard: Nit: is this still needed?

portability-transfer/src/main/java/org/datatransferproject/transfer/JobProcessor.java
```diff
@@ -20,14 +20,18 @@
 import com.google.common.io.BaseEncoding;
 import com.google.inject.Inject;
 import java.io.IOException;
+import java.util.Enumeration;
```
bwillard: Are these import changes still needed?

portability-transfer/src/main/java/org/datatransferproject/transfer/WorkerExtensionContext.java
```diff
@@ -19,13 +19,15 @@
 import com.google.common.base.Preconditions;
 import java.util.HashMap;
 import java.util.Map;
+import org.apache.log4j.Layout;
```
bwillard: Imports here too

portability-transfer/src/main/java/org/datatransferproject/transfer/logging/EncryptingLayout.java
```diff
@@ -0,0 +1,36 @@
...
+    Preconditions.checkState(jobId == null, "jobId has already been set");
+    jobId = inputJobId;
+  }
+
+  @Override
+  public String format(LoggingEvent event) {
+    // NB: copied from SimpleLayout.format()
+    return String.format("[%s] [%s]: %s - %s%s",
```
bwillard: Cna we either change to read this from the config file or leave a TODO to do that?

PR #496, opened 2018-07-23
Removed redundant script docker_push.sh

cyyber:
docker_push.sh is already in scripts directory (https://github.com/google/data-transfer-project/tree/master/scripts)


rtannenbaum:
hey @bwillard  is this script actually WAI? I don't see any updates to our demo image

PR #501, opened 2018-07-24
Add bwillard's example to the Integration doc

holachuy:


rtannenbaum:
nice, we should copy it into the docs too so it can be kept up to date..

Issue #511, opened 2018-07-25
Angular tests broken / unimplemented in client-rest/


rtannenbaum:
I only noticed this since I was looking into unrelated Travis issues, but the Travis test that looks like it passes doesn't actually do much: https://travis-ci.org/google/data-transfer-project/jobs/408238778

For comparison, here is a run from the tests we had implemented in the old client/: https://travis-ci.org/google/data-transfer-project/jobs/404991441

calumcalder:
Seems to have been fixed in #524

PR #512, opened 2018-07-25
Remove jdk transport extension (may add it back later if refactored)

rtannenbaum:
Tested running a photos transfer with the demo server


rtannenbaum:
More context: we are switching to the new restful model implemented by the jettyrest transport extension, where OAuth callbacks are separate from the API server. (In our demo server, they are handled by our Angular client after being forwarded back from nginx). In other implementations, they may be handled by other frontends.

Since the old jdk transport extension implemented OAuth handlers there and is inconsistent, we're removing it to avoid confusion, but may add it back if we refactor it to follow the new model. The more transport options we provide, the better, there is nothing inherently wrong with using 'jdk' (i.e. sun's HTTP server)

rtannenbaum:
@holachuy I updated the comment to reflect our offline discussion

holachuy:
Thanks! Can we check that all tests pass before submitting, in case anything depended on this?

rtannenbaum:
Java tests and demo image build are passing via Travis (I also tested a local run with the demo image) but the Angular tests actually need to be updated for the new rest client, so nothing to test there yet (https://github.com/google/data-transfer-project/issues/511)

PR #517, opened 2018-07-27
Start refactor to support handling all creds client-side during the auth flow

rtannenbaum:
Start refactor to support handling all creds client-side during the auth flow.

Since the API jobs are (by design) currently running code from arbitrary integrations, this refactor will allow all raw creds handling to be moved out of the API to more vetted/secure clients.

Changes in this PR:
- Simplify OAuth redirect flow to be entirely client-side. Instead of having the API prepare the import auth URL after the export auth has completed, now both the export and import URLs are obtained before the auth flow, and the redirects all happen on the client. The export and import URLs obtained from the API are stored in session storage along with the other client state.
- Introduce ReserveWorker and GetReservedWorker endpoints. Polling for a reserved worker is moved from the API to the client using these endpoints.

To do:
- #518: Move creds encryption from the API to the client
- #519: Handle code/token exchange which currently still resides in the API. Can either move this to client, too, or remove the (arbitrary) AuthExtensions in favor of a single standardized (and vetted) library.

Tested a photos migration locally.

#461


portability-api/src/main/java/org/datatransferproject/api/action/transfer/ReserveWorkerAction.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    @Override
+    public Class<ReserveWorker> getRequestType() {
+        return ReserveWorker.class;
+    }
+
+    @Override
+    public String handle(ReserveWorker reserveWorker) {
+        String id = reserveWorker.getId();
```
holachuy: Could this be null? Should we validate?

client-rest/src/app/transfer/auth-callback.component.ts
```diff
@@ -31,26 +27,29 @@ export class AuthCallbackComponent implements OnInit {
 
     ngOnInit() {
         let token = this.getToken();
-
         let transferId = this.progressService.transferId();
+        let importUrl = this.progressService.importUrl();
 
         if (Step.AUTHENTICATE_EXPORT === this.progressService.currentStep()) {
```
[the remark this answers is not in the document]
rtannenbaum: Done

client-rest/src/app/transfer/initiate-transfer.component.ts
```diff
@@ -33,15 +34,32 @@ export class InitiateTransferComponent implements OnInit {
     }
 
     initiate() {
-        let start: StartTransfer = {
+        let pollForWorkerKey = Observable.interval(1500)
```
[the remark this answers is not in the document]
rtannenbaum: Done

client-rest/src/app/transfer/initiate-transfer.component.ts
```diff
@@ -33,15 +34,32 @@ export class InitiateTransferComponent implements OnInit {
...
+          .switchMap(() => this.transferService.getReservedWorker({id: this.progressService.transferId()}))
+          .take(20)
```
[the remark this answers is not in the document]
rtannenbaum: Added an alert

portability-api/src/main/java/org/datatransferproject/api/action/transfer/ReserveWorkerAction.java
rtannenbaum: added validation (apparently Jackson does not honor the annotations we put in that say the field is required)

client-rest/src/app/transfer/initiate-transfer.component.ts
```diff
@@ -34,18 +34,26 @@ export class InitiateTransferComponent implements OnInit {
...
+                pollAttempt++;
                 if (reservedWorker.publicKey) {
-                  console.log("got worker with key: " + reservedWorker.publicKey);
+                  console.log("got transfer worker with public key: " + reservedWorker.publicKey);
                   pollForWorkerKey.unsubscribe();
                   this.progressService.workerReserved(reservedWorker.publicKey);
                   this.startTransferJob();
                   // TODO encrypt creds with worker public key
+                } else if (pollAttempt == maxPollAttempts) {
+                  alert(`Could not poll worker public key`);
```
holachuy: Maybe could not poll for worker information or something. We don't necessarily need to expose that we're polling for key information.

client-rest/src/app/transfer/initiate-transfer.component.ts
```diff
@@ -34,18 +34,26 @@ export class InitiateTransferComponent implements OnInit {
...
-          .take(20)
+          .take(maxPollAttempts)
           .subscribe(
             reservedWorker => {
-                console.log("polling for worker");
+                console.log("polling for assigned transfer worker");
+                pollAttempt++;
                 if (reservedWorker.publicKey) {
-                  console.log("got worker with key: " + reservedWorker.publicKey);
+                  console.log("got transfer worker with public key: " + reservedWorker.publicKey);
```
holachuy: // TODO: Remove debug statement in production

PR #530, opened 2018-08-01
Accept either raw or encrypted creds in API server

### LINKED ISSUE #528: Accept encrypted creds in StartTransferJobAction
Blocks #461 

see TODO here:
https://github.com/google/data-transfer-project/blob/cec5a7bd37049c81a44d9177753cd2420f927367/portability-api/src/main/java/org/datatransferproject/api/action/transfer/StartTransferJobAction.java#L57

I was thinking I would introduce a temporary flag where StartTransferJobAction could either accept raw or encrypted creds, and then remove once the client sends them encrypted (https://github.com/google/data-transfer-project/issues/527).

rtannenbaum:
Tested a photos transfer with the demo server

Fixes #528

### DIFF
--- a/client-rest/src/app/transfer/initiate-transfer.component.ts
+++ b/client-rest/src/app/transfer/initiate-transfer.component.ts
@@ -63,7 +63,9 @@ export class InitiateTransferComponent implements OnInit {
       let start: StartTransferJob = {
             id: this.progressService.transferId(),
             exportAuthData: this.progressService.exportAuthData(),
-            importAuthData: this.progressService.importAuthData()
+            importAuthData: this.progressService.importAuthData(),
+            // TODO: remove flag once client encrypts creds
+            authDataEncrypted: false
         };
         this.progressService.initiated();
         this.initiated = true;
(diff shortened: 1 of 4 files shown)


client-rest/src/app/transfer/initiate-transfer.component.ts
```diff
@@ -63,7 +63,9 @@ export class InitiateTransferComponent implements OnInit {
       let start: StartTransferJob = {
             id: this.progressService.transferId(),
             exportAuthData: this.progressService.exportAuthData(),
-            importAuthData: this.progressService.importAuthData()
+            importAuthData: this.progressService.importAuthData(),
+            // TODO: remove flag once client encrypts creds
+            authDataEncrypted: false
```
holachuy: Maybe debugAuthDataEncrypted?

rtannenbaum: I'm probably missing something but could you explain why 'debug'? I was trying to make it mean "whether the auth data is encrypted when sent to the API servers", not necessarily a debug thing

Either way, I think we'll get rid of the flag soon since Jim is looking into encrypting creds in the demo client now

PR #532, opened 2018-08-02
Rename some variables in creds encryption flow for clarification

rtannenbaum:
Maybe this only makes more sense in my head... we can feel free to leave it as is if you think it was clearer before


rtannenbaum:
Sure, I changed workerPublicKey back to authPublicKey. Makes sense, I need to stop calling it that in my head..

PR #535, opened 2018-08-07
Update Angular frontend set up on GCP

rtannenbaum:
Update demo-google-deployment to support the latest changes on GCP.

This routes callback/* to the Angular FE via index.html, and in Angular, we now store “authorizing” and “token” state in session storage during the OAuth flow. When we redirect back to index.html via callback, we detect that we are currently “authorizing” and route to the callback component.

Index.html is no longer served from the API server - it is served from the static bucket. To achieve this, the static bucket is changed to a "website bucket" with main page set to index.html. The bucket had to be renamed to the domain name in order to be a website bucket.

GCLB now defaults to the static bucket, and serves API requests via routing /api/* to the API backend service.

Also includes some flag clean up in the FE:
- Use environment.apiBaseUrl instead of hardcoded localhost:3000
- Remove unused environment.apiPostUrl
- Remove environment.apiUrl. Instead, derive from environment.apiBaseUrl ( + /api)

Tested deploying changes to GCP project and running demo-server locally, tested a photos transfer on both.


client-rest/src/environments/environment.ts
```diff
@@ -4,7 +4,7 @@
 
 export const environment = {
     production: false,
+    apiBaseUrl: 'https://localhost:3000',
```
holachuy: This makes sense but maybe the apiUrl below it needs a comment as to where that is used instead of assembling a url from the base

distributions/demo-google-deployment/bin/setup_gke_environment.sh
```diff
@@ -345,7 +345,8 @@ gcloud compute url-maps create ${LB_NAME} \
 --default-service ${API_BACKEND_SERVICE_NAME}
 gcloud compute url-maps add-path-matcher ${LB_NAME} \
 --default-service ${API_BACKEND_SERVICE_NAME} --path-matcher-name "static-bucket-mapping" \
---backend-bucket-path-rules "/static/*=${STATIC_BUCKET_NAME}"
+--backend-bucket-path-rules "/static/*=${STATIC_BUCKET_NAME},/index.html=${STATIC_BUCKET_NAME}" \
+--backend-service-path-rules "/api/*=${API_BACKEND_SERVICE_NAME}"
```
holachuy: Why is api here?

extensions/transport/portability-transport-jettyrest/src/main/java/org/datatransferproject/transport/jettyrest/http/JettyTransport.java
```diff
@@ -116,6 +120,58 @@ public void registerServlet(String path, Servlet servletContainer) {
...
+     */
+    @Override
+    public void handle(String target, Request baseRequest, HttpServletRequest request, HttpServletResponse response)
+            throws IOException
+    {
+      logger.warn("handling non-'api/*' request: " + request.getServletPath()
+              + ". Do NOT do this in production environments!");
+      if (request.getServletPath().equals("/")) {
+        String redirect = request.getServerName() + "/index.html";
+        redirect = (request.isSecure() || !request.getServerName().equals("localhost") ? "https://" : "http://")
```
holachuy: Can use use the useHttps variable to drive this rather than having different logic that determines if the api is running with https?

extensions/transport/portability-transport-jettyrest/src/main/java/org/datatransferproject/transport/jettyrest/http/JettyTransport.java
```diff
@@ -116,6 +120,58 @@ public void registerServlet(String path, Servlet servletContainer) {
...
+        // This is a hack to redirect OAuth callbacks back to the frontend so creds may be processed there.
+        // This should only be done for demo purposes. Secure implementations may not want to allow creds
+        // to reach here (i.e. the API server). We do this here due to limitations with the load balancer
+        // used in demo-google-deployment (we cannot handle the redirect there with the Google Cloud Load
+        // Balancer currently).
+        String redirect = request.getServerName() + "/index.html?";
+        redirect = (request.isSecure() || !request.getServerName().equals("localhost") ? "https://" : "http://")
+                + redirect;
+        for (Map.Entry<String, String[]> header : request.getParameterMap().entrySet()) {
+          if (header.getKey().equals("code")
```
holachuy: Maybe we can pass all params along?

rtannenbaum:
I pulled the backend changes out into #541

I'll come back to the frontend changes in this PR another time

client-rest/src/environments/environment.ts
rtannenbaum: Thanks, cleaned these up instead

distributions/demo-google-deployment/bin/setup_gke_environment.sh
rtannenbaum: Updated. Now that we use a "website bucket" for the static bucket, we can set the default page there to index.html and default the LB to the bucket. then we need /api/* to map to the API backend.

extensions/transport/portability-transport-jettyrest/src/main/java/org/datatransferproject/transport/jettyrest/http/JettyTransport.java
rtannenbaum: Outdated, we don't need this hack in the API any more with the new "website" static bucket.

extensions/transport/portability-transport-jettyrest/src/main/java/org/datatransferproject/transport/jettyrest/http/JettyTransport.java
rtannenbaum: Outdated, we don't need this hack in the API any more with the new "website" static bucket.

PR #539, opened 2018-08-13
Add InputStreams to Google Job Store

olsona:


olsona:
You're right, they should be separated.  I'll do that tonight, and check on the formatting.

Issue #537, opened 2018-08-10
Add support for InputStreams to Google Cloud Storage


olsona:

olsona:
#560

PR #540, opened 2018-08-13
Adding InputStreams to GoogleJobStore

olsona:
This will allow us to store InputStreams for objects that need auth to be accessed, like SmugMug photos.

Also propagates IOExceptions from JobStore through to importers.  Since everything is wrapped in RetryingCallable, this should be fine.

#537 #362

### DIFF
--- a/extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleJobStore.java
+++ b/extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleJobStore.java
@@ -72,44 +72,7 @@ public final class GoogleJobStore implements JobStore {
     return String.format("%s-%s", jobId, key);
   }
 
-  private static Map<String, Object> getProperties(Entity entity) {
-    if (entity == null) {
-      return null;
-    }
-    ImmutableMap.Builder<String, Object> builder = new ImmutableMap.Builder<>();
-    for (String property : entity.getNames()) {
-      // builder.put(property, entity.getValue(property));
-      if (entity.getValue(property) instanceof StringValue) {
-        builder.put(property, (String) entity.getString(property));
-      } else if (entity.getValue(property) instanceof LongValue) {
-        // This conversion is safe because of integer to long conversion above
-        builder.put(property, new Long(entity.getLong(property)).intValue());
-      } else if (entity.getValue(property) instanceof DoubleValue) {
-        builder.put(property, (Double) entity.getDouble(property));
-      } else if (entity.getValue(property) instanceof BooleanValue) {
-        builder.put(property, (Boolean) entity.getBoolean(property));
-      } else if (entity.getValue(property) instanceof TimestampValue) {
-        builder.put(property, (Timestamp) entity.getTimestamp(property));
-      } else {
-        Blob blob = entity.getBlob(property);
-        Object obj = null;
-        try {
-          try (ObjectInputStream in = new ObjectInputStream(blob.asInputStream())) {
-            try {
-              obj = in.readObject();
-            } catch (ClassNotFoundException e) {
-              e.printStackTrace();
-            }
-          }
-        } catch (IOException e) {
-          e.printStackTrace();
-        }
-        builder.put(property, obj); // BlobValue
-      }
-    }
 
-    return builder.build();
-  }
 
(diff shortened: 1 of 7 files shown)


extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleJobStore.java
```diff
@@ -333,4 +339,46 @@ private Key getDataKey(UUID jobId, String key) {
...
+    Key fullKey = getDataKey(jobId, key);
+    Entity entity = datastore.get(fullKey);
+    if (entity == null) {
+      return null;
+    }
+    try {
+      return objectMapper
+          .readValue(entity.getString(InputStream.class.getName()), InputStream.class);
+    } catch (IOException e) {
+      throw new RuntimeException("Failed to deserialize entityKey: " + fullKey, e);
```
holachuy: Since the create() method throws IOException, this one should probably rethrow this exception as well, perhaps wrapped in another IOException with the full key information.

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleJobStore.java
```diff
@@ -333,4 +337,46 @@ private Key getDataKey(UUID jobId, String key) {
       throw new RuntimeException("Failed to deserialized entityKey: " + entityKey, t);
     }
   }
+
+  @Override
+  public void create(UUID jobId, String key, InputStream stream) throws IOException {
+    Preconditions.checkNotNull(jobId);
```
rtannenbaum: can you make a helper method and refactor most of this? I think the only difference between this and the other create() is the objectMapper line

rtannenbaum:
Also can you add to the PR desc why we're adding the stream, I assume it's for Smugmug? Can you tag the issue? Thanks!

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleJobStore.java
```diff
@@ -63,13 +67,57 @@ public GoogleJobStore(Datastore datastore, ObjectMapper objectMapper) {
     this.objectMapper = objectMapper;
   }
 
+  @VisibleForTesting
+  static String getDataKeyName(UUID jobId, String key) {
+    return String.format("%s-%s", jobId, key);
+  }
+
+  private static Map<String, Object> getProperties(Entity entity) {
```
rtannenbaum: can this be moved back down since it is private?

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleJobStore.java
```diff
@@ -325,21 +302,22 @@ private Key getDataKey(UUID jobId, String key) {
...
   public void create(UUID jobId, String key, InputStream stream) throws IOException {
+    storeNewItem(jobId, key, stream);
+  }
+
+  @Override
+  public InputStream getStream(UUID jobId, String key) throws IOException {
+    return getItem(jobId, key, InputStream.class);
+  }
+
+  private void storeNewItem(UUID jobId, String key, Object object) throws IOException {
```
rtannenbaum: I think generalizing this input parameter to type Object is error prone since anything can be passed in.

How about instead we pass in a serialized model (type String), which is the result of ObjectMapper.writeAs...?

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleJobStore.java
olsona: Done.

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleJobStore.java
olsona: Done.

seehamrun:
Due to inactivity, Im closing this PR out - I think this ended up getting done in a different PR as well..

PR #541, opened 2018-08-14
Support serving API requests over HTTP, in addition to HTTPS

rtannenbaum:
This is necessary for our demo-google-deployment to work with the Jetty transport.

We can't serve HTTPS since don't have certs installed at our API servers; we have them installed at our LB instead.

Keep the default HTTPS for the demo server.

### DIFF
--- a/extensions/transport/portability-transport-jettyrest/src/main/java/org/datatransferproject/transport/jettyrest/http/JettyTransport.java
+++ b/extensions/transport/portability-transport-jettyrest/src/main/java/org/datatransferproject/transport/jettyrest/http/JettyTransport.java
@@ -48,36 +48,45 @@ public class JettyTransport {
   private static final String LOG_CLASS = "org.eclipse.jetty.util.log.class";
 
   private final KeyStore keyStore;
+  private final boolean useHttps;
 
   private int httpPort = 8080; // TODO configure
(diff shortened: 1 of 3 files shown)


extensions/transport/portability-transport-jettyrest/src/main/java/org/datatransferproject/transport/jettyrest/http/JettyTransport.java
```diff
@@ -48,36 +48,45 @@
...
-            server,
-            new SslConnectionFactory(sslContextFactory, "http/1.1"),
-            new HttpConnectionFactory(https));
-    sslConnector.setPort(httpPort);
-    server.setConnectors(new Connector[] {sslConnector});
+    if (useHttps) {
+      server = new Server();
+      SslContextFactory sslContextFactory = new SslContextFactory();
+      sslContextFactory.setKeyStore(keyStore);
+      sslContextFactory.setKeyStorePassword("password");
```
holachuy: TODO to parameterize these

rtannenbaum: Done

PR #548, opened 2018-08-17
Make constants upper case

rtannenbaum:
Tested a photos transfer in the demo server

### DIFF
--- a/portability-transfer/src/main/java/org/datatransferproject/transfer/WorkerModule.java
+++ b/portability-transfer/src/main/java/org/datatransferproject/transfer/WorkerModule.java
@@ -62,7 +62,7 @@ final class WorkerModule extends FlagBindingModule {
     try {
       return transferExtensions
           .stream()
-          .filter(ext -> ext.getServiceId().equals(service))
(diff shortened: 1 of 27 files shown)


rtannenbaum:
Yeah, agreed. I feel like these should be upper case regardless though since it's convention for constants. WDYT?

portability-transfer/src/main/java/org/datatransferproject/transfer/WorkerModule.java
```diff
@@ -62,7 +62,7 @@ private static TransferExtension findTransferExtension(
     try {
       return transferExtensions
           .stream()
-          .filter(ext -> ext.getServiceId().equals(service))
+          .filter(ext -> ext.getServiceId().equals(service.toLowerCase()))
```
holachuy: Can we write a test that ensures a mixed or lower case service name works here?

rtannenbaum: Good point, done. And it caught a bug :)

PR #564, opened 2018-08-22
Define callback URLs in the frontend, rather than API

rtannenbaum:
Different frontends may configure callbacks differently, and not necessarily use the "/callback/servicename" convention we had in the API server

Tested photos transfer locally with the demo server

### DIFF
--- a/client-rest/src/app/transfer/auth-callback.component.ts
+++ b/client-rest/src/app/transfer/auth-callback.component.ts
@@ -3,6 +3,7 @@ import {TransferService} from "./transfer.service";
 import {ProgressService, Step} from "../progress";
 import {ActivatedRoute, Router} from "@angular/router";
 import {transportError} from "../transport";
+import {environment} from "../../environments/environment";
 
 /**
  * Receives callbacks from external OAuth export and import services. Updates the application state with corresponding OAuth tokens.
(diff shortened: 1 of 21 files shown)


client-rest/src/app/transfer/auth-callback.component.ts
```diff
@@ -33,7 +34,12 @@ export class AuthCallbackComponent implements OnInit {
         if (Step.AUTHENTICATE_EXPORT === this.progressService.currentStep()) {
             // export auth step: received the export auth callback, generate the export auth data from the export token
             // and redirect to the import url. Use SSL as the token is sent cleartext.
-            this.transferService.generateAuthData({id: transferId, authToken: token, mode: "EXPORT"}).subscribe(
+            this.transferService.generateAuthData({
+              id: transferId,
+              authToken: token,
+              mode: "EXPORT",
+              callbackUrl: `${environment.apiBaseUrl}/callback/${this.progressService.exportService().toLowerCase()}`
```
holachuy: Do we expect to handle the service as part of the path in all clients?

rtannenbaum: per offline discussion, some clients may choose to use the same callback, with no service in the path

the demo client always includes service/ in the path

PR #565, opened 2018-08-23
Update logger used in demo-google-deployment

rtannenbaum:


distributions/demo-google-deployment/resources/log4j.properties
```diff
@@ -18,4 +18,4 @@
 
 log4j.rootLogger=DEBUG, A1
 log4j.appender.A1=org.apache.log4j.ConsoleAppender
-log4j.appender.A1.layout=org.datatransferproject.logging.EncryptingLayout
+log4j.appender.consoleAppender.layout=org.datatransferproject.logging.EncryptingLayout
```
olsona: consoleAppender isn't a defined appender.  The defined appenders are in line 19, log4j.rootLogger=DEBUG, A1.  

I suspect the real problem is that the logging library isn't actually being imported somewhere.

rtannenbaum:
Reverting this, I had a bad sync, Anna is right

PR #566, opened 2018-08-23
Client encryption extensibility

jimmarino:
This PR implements support for extending the server runtime with custom auth data encryption schemes via a worker security extension. Two schemes are provided out-of-the box: JWE and clear text.

JWE is enabled by default when building. To enable the cleartext extension:

1. Change the client-rest environment.ts entry 'encryptionScheme' to 'cleartext'
2. Build the server with the Gradle setting -PencryptionScheme=cleartext

Other schemes can be supported by supplying an extension module and changing the settings described above. 

Note that the server supports loading multiple security extension modules but the build is currently set to only enable one at a time for the demo server.

### DIFF
--- a/client-rest/src/app/transfer/initiate-transfer.component.ts
+++ b/client-rest/src/app/transfer/initiate-transfer.component.ts
@@ -60,6 +60,7 @@ export class InitiateTransferComponent implements OnInit {
 
     encryptAndStartTransfer() {
         // encrypt the export/import auth data pair as a JWE using the public key associated with the transfer job
+        // .cf https://tools.ietf.org/html/rfc7516#page-32
         let cryptographer = new Jose.WebCryptographer();
         (<any>cryptographer).setContentEncryptionAlgorithm("A128CBC-HS256"); // workaround missing method definition in Typescript type definition for WebCryptographer
         cryptographer.setKeyEncryptionAlgorithm("RSA-OAEP");
--- a/client-rest/src/app/transfer/initiate-transfer.component.ts
+++ b/client-rest/src/app/transfer/initiate-transfer.component.ts
@@ -60,7 +60,6 @@ export class InitiateTransferComponent implements OnInit {
 
     encryptAndStartTransfer() {
         // encrypt the export/import auth data pair as a JWE using the public key associated with the transfer job
-        // .cf https://tools.ietf.org/html/rfc7516#page-32
         let cryptographer = new Jose.WebCryptographer();
         (<any>cryptographer).setContentEncryptionAlgorithm("A128CBC-HS256"); // workaround missing method definition in Typescript type definition for WebCryptographer
         cryptographer.setKeyEncryptionAlgorithm("RSA-OAEP");
--- a/client-rest/src/app/transfer/initiate-transfer.component.ts
+++ b/client-rest/src/app/transfer/initiate-transfer.component.ts
@@ -6,6 +6,7 @@ import {TransferService} from "./transfer.service";
 import {StartTransferJob} from "../types";
 import {transportError} from "../transport";
 import {Jose, JoseJWE} from "jose-jwe-jws";
+import {environment} from "../../environments/environment";
 
 
 /**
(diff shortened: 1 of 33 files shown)


extensions/security/portability-security-cleartext/src/main/java/org/datatransferproject/security/cleartext/ClearTextAuthDataDecryptService.java
```diff
@@ -0,0 +1,46 @@
...
+  public boolean canHandle(String scheme) {
+    return "cleartext".equals(scheme);
+  }
+
+  @Override
+  public AuthDataPair decrypt(String encrypted, PrivateKey privateKey) {
+    try {
+      return objectMapper.readValue(encrypted, AuthDataPair.class);
+    } catch (IOException e) {
+      throw new SecurityException("Error decrypting auth tokens", e);
```
holachuy: Maybe: Error deserializing clear text auth tokens

portability-transfer/src/main/java/org/datatransferproject/transfer/JobProcessor.java
```diff
@@ -46,12 +49,18 @@
...
   private final InMemoryDataCopier copier;
+  private final Set<AuthDataDecryptService> decryptServices;
 
   @Inject
-  JobProcessor(JobStore store, ObjectMapper objectMapper, InMemoryDataCopier copier) {
+  JobProcessor(
+      JobStore store,
+      ObjectMapper objectMapper,
+      InMemoryDataCopier copier,
+      Set<AuthDataDecryptService> decryptServices) {
```
rtannenbaum: For a future change, what do you think about making the backend determine which scheme to accept? I think it would be more secure than allowing the frontend request to specify it and allow cleartext.

Am I understanding it correctly that the backend defines a set of allowed schemes, and the frontend specifies which one (of the allowed) on a per-request basis?

portability-transfer/src/test/java/org/datatransferproject/transfer/JobPollingServiceTest.java
```diff
@@ -134,6 +151,7 @@ public void pollingLifeCycle() throws Exception {
             .setImportService("DummyImportService")
             .setAndValidateJobAuthorization(
                 JobAuthorization.builder()
+                    .setEncryptionScheme("cleartext")
```
rtannenbaum: can you document somewhere that cleartext should be used for debug purposes only?

portability-transfer/src/main/java/org/datatransferproject/transfer/JobProcessor.java
rtannenbaum: Actually, scratch that, just realized backend can prevent cleartext from being allowed

portability-transfer/src/test/java/org/datatransferproject/transfer/JobPollingServiceTest.java
jimmarino: Sure, I added it in the UI console.

extensions/security/portability-security-cleartext/src/main/java/org/datatransferproject/security/cleartext/ClearTextAuthDataDecryptService.java
jimmarino: Yep, changed that.

client-rest/src/app/transfer/initiate-transfer.component.ts
```diff
@@ -51,13 +52,23 @@ export class InitiateTransferComponent implements OnInit {
                     if (reservedWorker.publicKey) {
                         pollForWorkerKey.unsubscribe();
                         this.progressService.workerReserved(reservedWorker.publicKey);
-                        this.encryptAndStartTransfer();
+                        if ("cleartext" === environment.encryptionScheme) {
```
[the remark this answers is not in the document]
jimmarino: Sure, I added it to the UI startup

PR #567, opened 2018-08-23
Add local store

olsona:
Add a local temp file store, to mirror the GoogleTempFileStore.


extensions/cloud/portability-cloud-local/src/main/java/org/datatransferproject/cloud/local/LocalTempFileStore.java
```diff
@@ -0,0 +1,47 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+public class LocalTempFileStore {
+
+  private String TEMP_DIR = "/tmp/";
+
+  void writeInputStream(String filename, InputStream inputStream) throws IOException {
+    File file = new File(TEMP_DIR + filename);
```
bwillard: I think if you just use getInputStream here it will result in less code.

PR #569, opened 2018-08-23
Adding another field to PhotoModel 

olsona:
Indicates whether photo is stored in temp storage or not


portability-types-transfer/src/main/java/org/datatransferproject/types/transfer/models/photos/PhotoModel.java
```diff
@@ -33,13 +35,15 @@ public PhotoModel(
       @JsonProperty("description") String description,
       @JsonProperty("mediaType") String mediaType,
       @JsonProperty("dataId") String dataId,
-      @JsonProperty("albumId") String albumId) {
+      @JsonProperty("albumId") String albumId,
+      @JsonProperty("inTempStore") boolean inTempStore) {
```
holachuy: TODO refactor out of model

PR #575, opened 2018-08-26
Updates for testing auth data generation, and some cleanups

rtannenbaum:
- add a job ID param temporarily to our export callback for testing purposes
- add some temporarily logging, also for testing purposes
- make ReserveWorker return a structured type rather than String

Tested a photos transfer in the demo server and demo-gcp-deployment

### DIFF
--- a/portability-api/src/main/java/org/datatransferproject/api/action/transfer/CreateTransferJobAction.java
+++ b/portability-api/src/main/java/org/datatransferproject/api/action/transfer/CreateTransferJobAction.java
@@ -17,6 +17,8 @@ import org.datatransferproject.spi.cloud.types.JobAuthorization;
 import org.datatransferproject.spi.cloud.types.PortabilityJob;
 import org.datatransferproject.types.client.transfer.CreateTransferJob;
 import org.datatransferproject.types.client.transfer.TransferJob;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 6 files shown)


portability-api/src/main/java/org/datatransferproject/api/action/transfer/CreateTransferJobAction.java
```diff
@@ -55,20 +59,25 @@
 
   @Override
   public TransferJob handle(CreateTransferJob request) {
+    // Create a new job and persist
+    UUID jobId = UUID.randomUUID();
```
holachuy: Where were we creating this before?

portability-api/src/main/java/org/datatransferproject/api/action/transfer/ReserveWorkerAction.java
```diff
@@ -35,12 +36,12 @@
...
 
   @Override
-  public String handle(ReserveWorker reserveWorker) {
+  public ReservedWorker handle(ReserveWorker reserveWorker) {
     String id = reserveWorker.getId();
     Preconditions.checkNotNull(id, "transfer job ID required for ReserveWorkerAction");
     UUID jobId = decodeJobId(id);
     updateStateToCredsAvailable(jobId);
-    return "";
+    return new ReservedWorker("");
```
holachuy: TODO to fill in with something?

PR #583, opened 2018-09-19
Add support for exporting Facebook photos

wmorland:
Adds an auth and data transfer extension for Facebook enabling the export of photos (including albums).


wmorland:
I need to sort out the CLA plus some rebase issues. Apologies, will resubmit.

Issue #581, opened 2018-09-18
Support photos that aren't in albums


olsona:
Not all photos exist within albums, and as such they are not transferred.

PR #585, opened 2018-09-21
Transferring albumless photos

olsona:
Basics of transferring albumless photos.

Due to API limitations, this is an inefficient process.  First, we get a list of all media items in a user's library.  We then get a list of all albums, and a list of media items in each album.  We subtract the set of all album-assigned photos to get the set of albumless photos and export those as part of the initial export step.

There is a **lot** of room for improvement here.  We don't yet address what would happen if any of those sets are too big for memory, and we don't take any advantage of having a list of all albums and their contents at the beginning.  This is just Step 1.

#581

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosExporter.java
+++ b/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosExporter.java
@@ -21,14 +21,18 @@ import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
+import java.util.stream.Collectors;
+import javax.annotation.Nullable;
 import org.datatransferproject.datatransfer.google.common.GoogleCredentialFactory;
 import org.datatransferproject.datatransfer.google.photos.model.AlbumListResponse;
 import org.datatransferproject.datatransfer.google.photos.model.GoogleAlbum;
 import org.datatransferproject.datatransfer.google.photos.model.GoogleMediaItem;
-import org.datatransferproject.datatransfer.google.photos.model.MediaItemSearchResponse;
+import org.datatransferproject.datatransfer.google.photos.model.MediaItemListResponse;
 import org.datatransferproject.spi.transfer.provider.ExportResult;
 import org.datatransferproject.spi.transfer.provider.ExportResult.ResultType;
 import org.datatransferproject.spi.transfer.provider.Exporter;
@@ -50,7 +54,6 @@ public class GooglePhotosExporter
 
   static final String ALBUM_TOKEN_PREFIX = "album:";
   static final String PHOTO_TOKEN_PREFIX = "media:";
-  static final String DEFAULT_ALBUM_ID = "default";
 
   private final GoogleCredentialFactory credentialFactory;
   private volatile GooglePhotosInterface photosInterface;
@@ -70,7 +73,13 @@ public class GooglePhotosExporter
   public ExportResult<PhotosContainerResource> export(UUID jobId, TokensAndUrlAuthData authData,
       Optional<ExportInformation> exportInformation) throws IOException {
     if (!exportInformation.isPresent()) {
-      return exportAlbums(authData, Optional.empty());
+      ExportResult<PhotosContainerResource> albumResult = exportAlbums(authData, Optional.empty());
+      List<PhotoModel> albumlessPhotos = getAlbumlessPhotos(authData);
+      if (albumlessPhotos != null && albumlessPhotos.size() > 0) {
+        return addMorePhotos(albumResult, albumlessPhotos);
+      } else {
+        return albumResult;
+      }
     } else {
       StringPaginationToken paginationToken =
           (StringPaginationToken) exportInformation.get().getPaginationData();
@@ -80,7 +89,7 @@ public class GooglePhotosExporter
       if (idOnlyContainerResource != null) {
         // export more photos
         return exportPhotos(
-            authData, idOnlyContainerResource.getId(), jobId, Optional.ofNullable(paginationToken));
+            authData, idOnlyContainerResource.getId(), Optional.ofNullable(paginationToken));
       } else {
         // export more albums if there are no more photos
         return exportAlbums(authData, Optional.ofNullable(paginationToken));
@@ -92,8 +101,9 @@ public class GooglePhotosExporter
    * Note: not all accounts have albums to return.  In that case, we just return an empty list of
    * albums instead of trying to iterate through a null list.
    */
-  private ExportResult<PhotosContainerResource> exportAlbums(
-      TokensAndUrlAuthData authData, Optional<PaginationData> paginationData) throws IOException {
+  @VisibleForTesting
+  ExportResult<PhotosContainerResource> exportAlbums(TokensAndUrlAuthData authData,
+      Optional<PaginationData> paginationData) throws IOException {
     Optional<String> paginationToken = Optional.empty();
     if (paginationData.isPresent()) {
       String token = ((StringPaginationToken) paginationData.get()).getToken();
@@ -107,9 +117,9 @@ public class GooglePhotosExporter
     albumListResponse = getOrCreatePhotosInterface(authData).listAlbums(paginationToken);
 
     PaginationData nextPageData = null;
-    if (!Strings.isNullOrEmpty(albumListResponse.getNextPageToken())) {
-      nextPageData = new StringPaginationToken(
-          ALBUM_TOKEN_PREFIX + albumListResponse.getNextPageToken());
+    String token = albumListResponse.getNextPageToken();
+    if (!Strings.isNullOrEmpty(token)) {
+      nextPageData = new StringPaginationToken(ALBUM_TOKEN_PREFIX + token);
     }
 
     ContinuationData continuationData = new ContinuationData(nextPageData);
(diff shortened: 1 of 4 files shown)


extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosExporter.java
```diff
@@ -73,30 +82,34 @@ public GooglePhotosExporter(GoogleCredentialFactory credentialFactory, JsonFacto
   public ExportResult<PhotosContainerResource> export(UUID jobId, TokensAndUrlAuthData authData,
       Optional<ExportInformation> exportInformation) throws IOException {
     if (!exportInformation.isPresent()) {
+      populateContainedPhotosList(jobId, authData);
```
bwillard: Maybe add a comment about why we are doing this?

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosExporter.java
```diff
@@ -73,30 +82,34 @@ public GooglePhotosExporter(GoogleCredentialFactory credentialFactory, JsonFacto
...
+    }
+    StringPaginationToken paginationToken =
+        (StringPaginationToken) exportInformation.get().getPaginationData();
+    IdOnlyContainerResource idOnlyContainerResource =
+        (IdOnlyContainerResource) exportInformation.get().getContainerResource();
+
+    boolean containerResourcePresent = idOnlyContainerResource != null;
+    boolean paginationDataPresent = paginationToken != null;
+
+    if (!containerResourcePresent
```
bwillard: Want to add some comments about the expected flow here?

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosExporter.java
```diff
@@ -167,38 +174,125 @@ public GooglePhotosExporter(GoogleCredentialFactory credentialFactory, JsonFacto
...
     return new ExportResult<>(resultType, containerResource, continuationData);
   }
 
+
+  @VisibleForTesting
+  void populateContainedPhotosList(UUID jobId, TokensAndUrlAuthData authData)
+      throws IOException {
+    // Get list of all album ids
+    List<String> albumIds = new LinkedList<>();
+    AlbumListResponse albumListResponse = getOrCreatePhotosInterface(authData)
```
bwillard: I think if you swapped this to a do {}while() you could skip duplicating some code.

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosExporter.java
```diff
@@ -167,38 +174,125 @@ public GooglePhotosExporter(GoogleCredentialFactory credentialFactory, JsonFacto
...
+          .listAlbums(Optional.of(token));
+      albumIds.addAll(
+          Arrays.stream(albumListResponse
+              .getAlbums())
+              .map(GoogleAlbum::getId)
+              .collect(Collectors.toList()));
+    }
+
+    // Get list of all ids belonging to photos belonging to albums
+    TempPhotosData tempPhotosData = jobStore
```
bwillard: This seems confusing to me, why create it once and then update it bellow?

Isn't the logic that we just write this data once?  Why not construct the data to store and then just insert it?

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosExporter.java
```diff
@@ -167,38 +174,125 @@ public GooglePhotosExporter(GoogleCredentialFactory credentialFactory, JsonFacto
...
+        .findData(jobId, createCacheKey(), TempPhotosData.class);
+    if (tempPhotosData == null) {
+      tempPhotosData = new TempPhotosData(jobId);
+      jobStore.create(jobId, createCacheKey(), tempPhotosData);
+    }
+    for (String albumId : albumIds) {
+      MediaItemSearchResponse albumMediaItemResponse = getOrCreatePhotosInterface(authData)
```
bwillard: Is there a reason to re-crete the photos interface every time? If so we should document it.

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosExporter.java
```diff
@@ -167,38 +174,125 @@ public GooglePhotosExporter(GoogleCredentialFactory credentialFactory, JsonFacto
...
+        albumMediaItemResponse = getOrCreatePhotosInterface(authData)
+            .listMediaItems(Optional.of(albumId), Optional.of(token));
+        if (albumMediaItemResponse.getMediaItems().length > 0) {
+          tempPhotosData.addAllContainedPhotoIds(
+              Arrays.stream(albumMediaItemResponse.getMediaItems())
+                  .map(GoogleMediaItem::getId)
+                  .collect(Collectors.toList()));
+        }
+      }
+    }
```
bwillard: I think a do {} while() will be cleaner here too.

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosExporter.java
```diff
@@ -167,38 +174,125 @@ public GooglePhotosExporter(GoogleCredentialFactory credentialFactory, JsonFacto
...
+
     return new ExportResult<>(resultType, containerResource, continuationData);
   }
 
+
+  @VisibleForTesting
+  void populateContainedPhotosList(UUID jobId, TokensAndUrlAuthData authData)
+      throws IOException {
+    // Get list of all album ids
+    List<String> albumIds = new LinkedList<>();
```
bwillard: Any reason to have albumIds as a separate variable, this seems like just two nested loops is a little cleaner right?

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosExporter.java
```diff
@@ -167,38 +174,125 @@ public GooglePhotosExporter(GoogleCredentialFactory credentialFactory, JsonFacto
...
+              Arrays.stream(albumMediaItemResponse.getMediaItems())
+                  .map(GoogleMediaItem::getId)
+                  .collect(Collectors.toList()));
+        }
+      }
+    }
+    jobStore.update(jobId, createCacheKey(), tempPhotosData);
+  }
+
+  private Optional<String> getPaginationToken(Optional<PaginationData> paginationData) {
```
bwillard: Maybe rename this to getPhotosPagnationToken?  right now I would think it would work with albums too.

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosExporter.java
olsona: I believe this clarifies some things.

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosExporter.java
olsona: I believe that pattern is to allow for the possibility that auth data might expire.  We haven't written any code to address that yet, however.

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosExporter.java
olsona: Much cleaner, thank you.

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosExporter.java
olsona: I think in general we want to pull the temp data and add to it, but since there won't be any temp data to start with for this case, it should be fine.

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosExporter.java
```diff
@@ -190,57 +204,39 @@ public GooglePhotosExporter(GoogleCredentialFactory credentialFactory, JobStore
...
+      for (GoogleAlbum album : albumListResponse.getAlbums()) {
+        String albumId = album.getId();
+        String photoToken = null;
+        do {
+          containedMediaSearchResponse = getOrCreatePhotosInterface(authData)
+              .listMediaItems(Optional.of(albumId), Optional.ofNullable(photoToken));
+          for (GoogleMediaItem mediaItem : containedMediaSearchResponse.getMediaItems()) {
+            tempPhotosData.addContainedPhotoId(mediaItem.getId());
+          }
+        } while ((photoToken = containedMediaSearchResponse.getNextPageToken()) != null);
```
bwillard: I would avoid assignments in conditionals, they get confusing.

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosExporter.java
```diff
@@ -190,57 +204,39 @@ public GooglePhotosExporter(GoogleCredentialFactory credentialFactory, JobStore
...
       }
-    }
-    jobStore.update(jobId, createCacheKey(), tempPhotosData);
+    } while ((albumToken = albumListResponse.getNextPageToken()) != null);
```
bwillard: Here too

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosExporter.java
olsona: Done.

PR #587, opened 2018-09-24
Add support for exporting Facebook photos

wmorland:
Adds an auth and data transfer extension for Facebook enabling the export of photos (including albums).


extensions/data-transfer/portability-data-transfer-facebook/src/main/java/org/datatransferproject/transfer/facebook/photos/FacebookPhotosExporter.java
```diff
@@ -0,0 +1,103 @@
...
+  public ExportResult<PhotosContainerResource> export(
+      UUID jobId, TokensAndUrlAuthData authData, Optional<ExportInformation> exportInformation) {
+    Preconditions.checkNotNull(authData);
+
+    ArrayList<PhotoAlbum> exportAlbums = new ArrayList<>();
+
+    // Get albums
+    Iterable<List<Album>> connection = getOrCreatePhotosInterface(authData).getAlbums();
+
+    for (List<Album> albums : connection) {
```
holachuy: Maybe we can add a TODO to paginate through albums so we don't have to hold them all in memory.

portability-types-transfer/src/main/java/org/datatransferproject/types/transfer/models/photos/PhotoModel.java
```diff
@@ -85,4 +86,17 @@ public String toString() {
         .add("inTempStore", inTempStore)
         .toString();
   }
+  
+  @Override
+  public boolean equals(Object o) {
```
holachuy: Thanks!

extensions/data-transfer/portability-data-transfer-facebook/src/test/java/org/datatransferproject/transfer/facebook/photos/FacebookPhotosExporterTest.java
```diff
@@ -0,0 +1,112 @@
...
+
+  private FacebookPhotosExporter facebookPhotosExporter;
+  private UUID uuid = UUID.randomUUID();
+
+  @Before
+  public void setUp() {
+    FacebookPhotosInterface photosInterface =
+        new FacebookPhotosInterface() {
+          @Override
+          public Iterable<List<Album>> getAlbums() {
```
holachuy: We have used mockito for configuring this kind of behavior. Here is an example in case you find it useful if this gets unwieldy 

https://github.com/google/data-transfer-project/blob/8fbeadf3ef7ac47d489abac1c9862d300d9a89b6/extensions/data-transfer/portability-data-transfer-flickr/src/test/java/org/datatransferproject/datatransfer/flickr/photos/FlickrPhotosExporterTest.java#L111

extensions/data-transfer/portability-data-transfer-facebook/src/main/java/org/datatransferproject/transfer/facebook/photos/FacebookPhotosExporter.java
wmorland: Added

extensions/data-transfer/portability-data-transfer-facebook/src/test/java/org/datatransferproject/transfer/facebook/photos/FacebookPhotosExporterTest.java
wmorland: Makes sense, switched to Mockito.

extensions/data-transfer/portability-data-transfer-facebook/src/main/java/org/datatransferproject/transfer/facebook/photos/FacebookPhotosExporter.java
sarjo23: import com.restfb.exception.FacebookOAuthException; 
 import org.datatransferproject.spi.transfer.types.CopyExceptionWithFailureReason; 
 import org.datatransferproject.spi.transfer.types.SessionInvalidatedException; 
 import org.datatransferproject.spi.transfer.types.UnconfirmedUserException; 
 import org.datatransferproject.spi.transfer.types.UserCheckpointedException; 
  
 public class FacebookTransferUtils { 
  
   public static FacebookOAuthException handleFacebookOAuthException(FacebookOAuthException e) 
       throws CopyExceptionWithFailureReason { 
     String message = e.getMessage(); 
     if (message != null && message.contains("the user is not a confirmed user")) { 
       throw new UnconfirmedUserException( 
           "The user account is not confirmed or deactivated", e); 
     } else if (message != null && message.contains("code 190, subcode 459")) { 
       // Throw out exception for known user checkpointed error from Graph API 
       throw new UserCheckpointedException("The user has been checkpointed", e); 
     } else if (message != null && message.contains("code 190, subcode 460")) { 
       throw new SessionInvalidatedException("The user session has been invalidated", e); 
     } else if (message != null && message.contains("code 190, subcode 463")) { 
       throw new SessionInvalidatedException("The user session has expired", e); 
     } else { 
       return e; 
     } 
   }

sarjo23: > Maybe we can add a TODO to paginate through albums so we don't have to hold them all in memory.

sarjo23: > Maybe we can add a TODO to paginate through albums so we don't have to hold them all in memory.

sarjo23: > > Maybe we can add a TODO to paginate through albums so we don't have to hold them all in memory.
> 
>

portability-types-transfer/src/main/java/org/datatransferproject/types/transfer/models/photos/PhotoModel.java
sarjo23: - @****)

Issue #580, opened 2018-09-18
Publish artifacts to Maven Central


wmorland:
We would really like to have the artifacts published in Maven Central so that they can be easily imported and tracked and each user of the project doesn't have to set up their own build pipeline to stay up to date. This could be done either directly via their Open Source Software Repository Hosting https://central.sonatype.org/pages/producers.html or via another repository such as jcenter https://github.com/bintray/gradle-bintray-plugin#maven-central-sync

This would require some additions to the gradle scripts as well as registering with Maven Central and setting up a key for signing the artifacts.

jimmarino:
Hi William,

I'm looking into this now via JCenter. Let me write-up a description of what I'm thinking later today or earl morning tomorrow and we can discuss via this thread or Slack.

wmorland:
Hi Jim, thank you for laying everything out so clearly.

I can only speak for our use case which seems to fit into number 2 of the goals. We need it for hooking into proprietary cloud infrastructure rather than a particular need for private data transfer or auth extensions. It is preferable for us to have a private repository that depends on the public DTP artifacts compared to maintaining a private fork or running a private build pipeline + dependency management.

- I believe the reason Maven Central is preferred as the repository is that JCenter does not require signed binaries (I may be wrong on this).
- There isn't a need for a starter project from us but I can see how it might be useful.

jimmarino:
That's a good point about singing artifacts. I've started a PR with configuration that produces Maven artifacts here:

https://github.com/google/data-transfer-project/pull/582

I'll iterate on that and add in support for signing. Once we have that and JCenter and Maven Central repos to publish to, we should be able to easily replicate from the former to the latter.

PR #591, opened 2018-09-26
Publish Maven-based artifacts

jimmarino:
This is an updated PR that adds support for publishing Maven-based artifacts to the build and uploading them to Bintray.

Relates to https://github.com/google/data-transfer-project/issues/580

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,4 @@
 .gradle/
-gradle/
-**/gradlew
-**/gradlew.bat
 *.log
 
 secrets.csv
--- a/build.gradle
+++ b/build.gradle
@@ -173,7 +173,7 @@ def getPomConfig() {
             developer {
                 id 'dtp'
                 name 'Data Transfer Project'
-                email "dev@null.com"
+                email "portability-maintainers@googlegroups.com"
             }
         }
 
--- a/build.gradle
+++ b/build.gradle
@@ -91,84 +91,99 @@ configure(sourceProjects()) {
     }
 }
 
-// Setup publishing configuration for all subprojects that have applied the Maven Publish plugin
-subprojects {
-
-    plugins.withId('maven-publish') {
-
-        apply plugin: 'io.spring.bintray'
-
-        // Configure artifact publication
-        def pomConfig = {
-            licenses {
-                license {
-                    name 'Apache 2.0'
-                    url 'https://www.apache.org/licenses/LICENSE-2.0.html'
-                    distribution 'repo'
-                }
-            }
-            developers {
-                developer {
-                    id 'dtp'
-                    name 'Data Transfer Project'
-                    email "dev@null.com"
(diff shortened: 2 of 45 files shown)


jimmarino:
Thanks Will, I fixed that and pushed!

build.gradle
```diff
@@ -70,8 +79,112 @@ configure(sourceProjects()) {
...
+                name 'Apache 2.0'
+                url 'https://www.apache.org/licenses/LICENSE-2.0.html'
+                distribution 'repo'
+            }
+        }
+        developers {
+            developer {
+                id 'dtp'
+                name 'Data Transfer Project'
+                email "dev@null.com"
```
holachuy: Should we put in portability-maintainers@googlegroups.com

client-rest/build.gradle
```diff
@@ -94,4 +96,39 @@ task installTools {
...
+// creates the distribution ZIP containing the compiled Angular application
+task createDistribution(type: Zip) {
+    from "dist/portability-demo"
+    include "*"
+    include "*/*"
+    archiveName "client-rest.zip"
+    destinationDir(file("${buildDir}/libs"))
+}
+
+def distFile = file("${buildDir}/libs/client-rest.zip")
```
holachuy: Does this only publish the distribution directory?

.gitignore
```diff
@@ -1,7 +1,4 @@
 .gradle/
-gradle/
```
holachuy: Will these need to no longer be ignored for publishing reasons?

client-rest/build.gradle
jimmarino: The build creates an Angular-compiled web app, zips it and publishes a ZIP-based artifact to Maven/JCenter. That will allow any Gradle, Maven or Ivy-based build system to use built-in dependency resolution to download it as a standard artifact. The starter repo will do this and then unzip the contents to the Docker image.

.gitignore
jimmarino: They actually should not be ignored because we can't upgrade the Gradle version if they are. I had to upgrade Gradle to resolve a bug in the version we were using. That means we will need to be careful about changes to that directory but the only way that will happen is if someone explicitly upgrades Gradle or directly mucks around in it.

holachuy: Should we add upgrade instructions to the README?

client-rest/build.gradle
holachuy: Does this include sources or only the compiled code? I'm not sure how javascript distributions work.

holachuy: Oh and can we add a comment or documentation

.gitignore
wmorland: We can always just link to https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:upgrading_wrapper

PR #592, opened 2018-10-09
Make ReserveWorkerAction return a ReservedWorker response (instead of an empty string)

seehamrun:


portability-api/src/main/java/org/datatransferproject/api/action/transfer/ReserveWorkerAction.java
```diff
@@ -35,12 +36,12 @@
...
 
   @Override
-  public String handle(ReserveWorker reserveWorker) {
+  public ReservedWorker handle(ReserveWorker reserveWorker) {
     String id = reserveWorker.getId();
     Preconditions.checkNotNull(id, "transfer job ID required for ReserveWorkerAction");
     UUID jobId = decodeJobId(id);
     updateStateToCredsAvailable(jobId);
-    return "";
+    return new ReservedWorker("");
```
holachuy: Let's comment this for now or inside the constructor of the ReservedWorker

seehamrun: Done. Also addded a TODO to make this a status instead

rtannenbaum: 👍

PR #621, opened 2018-11-22
Allow jobs to have preconfigured ExportInformation

wmorland:
This adds an optional ExportInformation to the PortabilityJob specification to allow specifying Albums for an export. The FacebookPhotosExporter supports this by reading albums from a PhotosContainerResource on the ExportInformation and then exporting those albums and their photos but no other data.

Currently this works by the logic differentiating between PhotosContainerResource and IdOnlyContainerResource getting handed to the exporter. I'm open to other suggestions for how to handle this. It was partly a tradeoff between creating new concepts and flows and reusing existing  constructs.


rtannenbaum:
@wmorland I'm going to pass this one on to @holachuy since I think he offered to take it in the maintainers meeting, also he's more familiar with the container logic than I am

wmorland:
I've updated against master and sorted out some imports that caused the build to break after the merge. It now just needs an approval again please. \cc @jimmarino @holachuy

PR #624, opened 2018-11-29
Add job lifetime hooks interface

BrentChesny:
This introduces some changes as to how the end of a job is handled. Current behaviour is to simply return when the data copier fails because of a RetryException for example. We introduce the ExportFailedException and ImportFailedException to throw here instead. 

We also add a new JobHooks interface with two  methods that are called at the start and end of a job. A cloud extension can then provide an implementation of this interface and decide how they want to handle this. I've also written a default implementation of these hooks, which mimics the current behaviour of removing a job from the jobstore. The interface also allows a cloud extension to handle success and failure differently.


jimmarino:
Hi Brent,

The lifecycle hook looks fine to me. A question I had was regarding the additional import and export exceptions. Would it be better to return a status from the copy method and signal failure rather than having an exception? It may be cleaner in the client to then decide what to do as part of the normal as opposed to exceptional flow. Another option would to have a single exception that can detail what side of the import/export operation failed.

BrentChesny:
Hey Jim,

I thought about having some sort of status returned from the copier, but the recursive nature of the copier makes that a bit messy (i.e. we should probably only return success if all recursive calls also return success, and checking that could get awkward). I guess having a single CopyException would be the better option then? The message can clarify if it was the import or export that failed.

Let me know what you think.

jimmarino:
That sounds good to me.

BrentChesny:
Oops, should probably not have force pushed here.

jimmarino:
Hi Brent (and others interested),

I was merging work I am doing on replacing SLF4J with an extension mechanism and ran into an issue. I see that the default impl of JobHooks is bound in CloudExtensionModule. If a cloud extension does not use Guice (or derive from that module), an error will be thrown since JobProcessor requires it to be injected.

This also got me thinking that someone may want to use an extension deriving from CloudExtensionModule and user their own impl of JobHooks. 

One way we could solve this is to make JobHooks and extension interface similar to the other extension types and rely on ServiceLoader to resolve implementations (we could support multiplexing as well) when the Worker module is created. The worker module could then bind the JobHooks extensions into its Guice injector context. This is similar to how I am implementing the Monitor mechanism.

What do you think of this approach or do you have an alternative that could work better?

I'm going to have a rather complex merge due to all of the places where SLF4J is referenced (this issue came up as I am merging), so my thought is to introduce a workaround to avoid the issue in the merge. After the merge is working, approved, and committed, we can look at coming up with a solution to the JobHooks issue.

Let me know your thoughts.

BrentChesny:
Hey Jim,

I was under the impression that every cloud extension would use Guice and derive from CloudExtensionModule, but that may not be the case then? The main idea here was indeed that a cloud extension can create their own implementation of JobHooks, and override the binding in their subclass of CloudExtensionModule.

If you think it makes sense to make JobHooks an extension on its own, then I'm open to that. To me it seemed more like it should be part of the cloud extension though.

jimmarino:
Hi Brent,

Yeah, the idea was Guice would be optional for extensions. To ensure that continues to work, I created the MS extensions without the use of Guice (also, the wiring is simple in the MS extensions, so Guice is a bit of overkill there).

Looking at JobHooks, it seems more like a lifecycle interface for Worker/transfer operations. For example, in addition to resource cleanup (which seems like a cloud extension thing), some deployments may want to collect metrics that are sent to a collector not related to how jobs are persisted (e.g. through multiplexing). Would it make sense to have that interface in the Transfer SPI module?

BrentChesny:
> Yeah, the idea was Guice would be optional for extensions.

Ah, I didn't realize that. 

Yeah, it would make sense to move it to the Transfer SPI module. We can look into that after you merge the logger extension.

PR #628, opened 2018-12-10
Monitor

jimmarino:
This PR replaces SLF4J logging with a monitor extension using the JDK ServiceLoader mechanism. Monitor implementations can send logging and system events to their sink of choice, e.g. SLF4J, Log4J, a cloud logging service, etc.

Unfortunately this PR is huge as logging involved my parts of the system.

JobHooks was also modified to work with cloud extensions that do not use Guice. An injection error was thrown in these runtime configurations when a worker process was started. To resolve this, JobHooks was moved to the worker SPI module (it seems more about jobs). I also moved the callback to JobStore.remove() from DefaultJobHook to JobProcessor since having the invocation in DefaultJobHook would have coupled the Worker SPI to the Cloud SPI. In the end, I also made the JobHook methods default methods so DefaultJobHook was no longer needed.

In an ensuing PR, we can make JobHook an extension that is loaded by ServiceLoader when a worker process is initialized.

*** In the meantime, the fact that JobHook is not yet extensible and is moved from the cloud extension may cause issues for people that were relying on their own cloud extension and JobHook implementation. It should be a relatively small task to make JobHook extensible so this should not be an issue for long.


BrentChesny:
Thanks for taking care of the JobHooks part. Do you think calling JobStore.remove() is sensible default behaviour though? I think setting the status to COMPLETED or ERROR is more desirable, and will make debugging easier.

wmorland:
@jimmarino I'm happy for this to be merged. Are we good to go? (merge conflicts aside)

jimmarino:
@wmorland Should be good now pending approval. I'll take a look at your ensuing PR so we can get that in too.

jimmarino:
Looks as if the merge did not go well...Let me check

wmorland:
That was a little premature of me, I just looked at the delta

jimmarino:
OK fixed it. I accidentally took a dependency out during the merge

PR #629, opened 2018-12-12
Fix the docker image breaking from no JobHooks

wmorland:
The DTP docker image has been broken since we merged in the JobHooks code. This provides an empty Job Hooks implementation to quick fix the issue.


wmorland:
Yes #628 resolves it so this should be unnecessary.

PR #636, opened 2018-12-18
Set job state appropriately after finishing job

BrentChesny:
Do not remove job from jobstore upon completion, but set the job state to either COMPLETE or ERROR.


rtannenbaum:
+1 to setting a TTL

Issue #641, opened 2019-01-10
Implement Imgur Transfer Extension


seehamrun:
Imgur has an Oauth 2 API: https://apidocs.imgur.com/#intro

natlg:
Would be glad to take this issue! I'm working on the exporter for now

PR #645, opened 2019-01-31
Imgur exporter and auth extension

natlg:
Here is the exporter (issue #641), if this implementation is fine I'll add importer and tests in another PR.

Few notes:

Imgur doesn't require scopes but adding any value causes error, so I left it empty in the [ImgurOAuthConfig](https://github.com/natlg/data-transfer-project/blob/3fb7c7f93b4b3d7b408a78d25bfc2c5f05c278ac/extensions/auth/portability-auth-imgur/src/main/java/org.datatransferproject.auth.imgur/ImgurOAuthConfig.java#L49) and included checking to the [OAuth2DataGenerator](https://github.com/natlg/data-transfer-project/blob/3fb7c7f93b4b3d7b408a78d25bfc2c5f05c278ac/libraries/auth/src/main/java/org/datatransferproject/auth/OAuth2DataGenerator.java#L70)

Also seems public API doesn't allow to get non-album photos separately, only photos for each album or just all photos for the account (they don't include information about the album). So I save images from albums, request all account images and then pick those that are not in albums. Maybe there can be a better way.

During authorization API can return error "Imgur is temporarily over capacity ...", sometimes for the long time.


extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/ImgurTransferExtension.java
```diff
@@ -0,0 +1,96 @@
...
+import org.datatransferproject.spi.transfer.provider.Exporter;
+import org.datatransferproject.spi.transfer.provider.Importer;
+import org.datatransferproject.datatransfer.imgur.photos.ImgurPhotosExporter;
+import org.datatransferproject.datatransfer.imgur.photos.ImgurPhotosImporter;
+import org.datatransferproject.types.transfer.auth.AppCredentials;
+
+/** Extension for transferring Imgur data */
+public class ImgurTransferExtension implements TransferExtension {
+  private static final String SERVICE_ID = "Imgur";
+  private static final String BASE_URL = "https://api.imgur.com/3";
```
seehamrun: is BASE_URL used elsewhere other than passed to the exporter, if not lets move it into the Exporter

extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosExporter.java
```diff
@@ -0,0 +1,232 @@
...
+  private final String allPhotosUrlTemplate;
+  private final String albumsUrlTemplate;
+  private final String defaultAlbumId;
+
+  private final OkHttpClient client;
+  private final ObjectMapper objectMapper;
+  private Monitor monitor;
+
+  public ImgurPhotosExporter(Monitor monitor, String baseUrl) {
+    albumPhotosUrlTemplate = baseUrl + "/album/%s/images";
```
seehamrun: Since we dont expect the basUrl of the API to change, can we make all these final static and initialize them as constants above?

extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosExporter.java
```diff
@@ -0,0 +1,232 @@
...
+    albumsUrlTemplate = baseUrl + "/account/me/albums/%s";
+    allPhotosUrlTemplate = baseUrl + "/account/me/images/%s";
+    defaultAlbumId = "0";
+    this.client = new OkHttpClient.Builder().build();
```
seehamrun: I think you can pass in the extension context to the constructor and get the Http Transport and object mapper from there (or pass them in directly)

 For an example see: https://github.com/google/data-transfer-project/blob/master/extensions/data-transfer/portability-data-transfer-instagram/src/main/java/org/datatransferproject/transfer/instagram/InstagramTransferExtension.java#L69

extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosExporter.java
```diff
@@ -0,0 +1,232 @@
...
+   * Exports albums and photos. Gets albums first, then photos which are contained in albums and
+   * non-album photos
+   *
+   * @param jobId the job id
+   * @param authData authentication data for the operation
+   * @param exportInformation info about what data to export, see {@link ExportInformation} for more
+   */
+  @Override
+  public ExportResult<PhotosContainerResource> export(
+      UUID jobId, TokensAndUrlAuthData authData, Optional<ExportInformation> exportInformation)
```
seehamrun: We're not making use of the ExportInformation here since we're just getting all of the content in one go. 

Lets try and split it up, some ideas: 
- common pattern is you either retrieve the albums list or the photos list at each call based on the exportInformation. If its empty, it means we haven't gotten anything yet and we should export the list of albums, if there is a an album we should retrieve the data for it (See Flickr Exporter for an example of this)
- You mentioned the API doesn't allow to get non-album photos separately - when you retrieve the photos do they include which album they belong in? The other thing we can do is retrieve the list of albums, then all photos and add them to albums as we process each photo.  But again, so we dont request all of them at the same time, we include the pagination token in the ExportResult and when we enter this function, we pick up from the page that we left off at. 

Let me know if that makes sense, also adding chuy@ to take a look as well since its been a while since I've written or reviewed transfer extensions.

extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosImporter.java
```diff
@@ -0,0 +1,40 @@
+
```
seehamrun: delete this since we aren't dealing with import yet

natlg:
Hey Siham, thank you for the review! I'll return to it next week

Documentation/JettyApiSpec.yaml
```diff
@@ -244,12 +244,14 @@ components:
...
         - "TWITTER"
         - "GOOGLE"
+        - "IMGUR"
         importServices:
         - "FLICKR"
         - "SMUGMUG"
         - "Facebook"
         - "TWITTER"
         - "GOOGLE"
+        - "IMGUR"
```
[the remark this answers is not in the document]
natlg: Done

extensions/auth/portability-auth-imgur/src/main/java/org.datatransferproject.auth.imgur/ImgurOAuthConfig.java
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  // Imgur doesn't require scopes
+  @Override
+  public Map<String, Set<String>> getExportScopes() {
+    return ImmutableMap.of("PHOTOS", ImmutableSet.of(""));
+  }
+
+  @Override
+  public Map<String, Set<String>> getImportScopes() {
```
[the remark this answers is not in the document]
natlg: Done

extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/ImgurTransferExtension.java
```diff
@@ -0,0 +1,96 @@
...
+
+  private boolean initialized = false;
+
+  private static final ImmutableList<String> SUPPORTED_DATA_TYPES = ImmutableList.of("PHOTOS");
+
+  private ImgurPhotosImporter importer;
```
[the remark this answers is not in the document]
natlg: Done

extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/ImgurTransferExtension.java
natlg: It will be used in importer, so I left it here, but made `public static` for using in the exporter

extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosExporter.java
natlg: Done

extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosExporter.java
natlg: Done, but I used `OkHttpClient` and had to register it in the [WorkerMain](https://github.com/google/data-transfer-project/blob/2dc1f8ed03e9e52ffe93bdca7c81878b5041d714/portability-transfer/src/main/java/org/datatransferproject/transfer/WorkerMain.java#L81) 
It's used in MicrosoftPhotosExporter too. I can replace it to the `HttpTransport`, if it's preffered

extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosExporter.java
natlg: I refactored it and now exporter uses ExportInformation. Unfortunately retrieved photos don't include album (here is it's [data model](https://api.imgur.com/models/image) )
Exporter performs these steps:
- At first gets albums (with pagination) and saves their ids to the ExportInformation.  
- Retrieves photos for each album on the subsequent export calls (request doesn't support pagination), saves their ids to the Set.
- Exports non-album photos: retrieves all photos (with pagination) and excludes album photos saved to the Set.

extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosImporter.java

natlg:
I addressed comments, PTAL if anything else should be changed

extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosExporter.java
```diff
@@ -0,0 +1,284 @@
...
+  /**
+   * Exports albums.
+   *
+   * @param authData authentication information
+   * @param paginationData pagination information to use for subsequent calls
+   */
+  private ExportResult<PhotosContainerResource> requestAlbums(
+      TokensAndUrlAuthData authData, PaginationData paginationData) throws IOException {
+    ImmutableList.Builder<PhotoAlbum> albumBuilder = ImmutableList.builder();
+    List<IdOnlyContainerResource> albumIds = new ArrayList<>();
```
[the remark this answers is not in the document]
natlg: Added saving photos to the temp store, should it be done for albums too? I've found examples of only photo saving in SmugMugPhotosExporter and Google exporters

extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosExporter.java
```diff
@@ -0,0 +1,284 @@
...
+
+    int page = paginationData == null ? 0 : ((IntPaginationToken) paginationData).getStart();
+
+    String url = format(ALBUMS_URL_TEMPLATE, page);
+
+    List<Map<String, Object>> items = requestData(authData, url);
+
+    // Request result doesn't indicate if it's the last page
+    boolean hasMore = items != null && items.size() != 0;
```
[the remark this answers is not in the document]
natlg: Done

extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosExporter.java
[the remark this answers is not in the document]
seehamrun: This looks good to me! - Albums data can be stored if needed, this happens usually if we want to maintain what the mapping from source album -> destination albums will be, but this is probably only necessary in the importer.

holachuy:
Thank you natlg! would you mind syncing and resolving the conflicts? I'd like to try this locally and merge it in.

natlg:
>  would you mind syncing and resolving the conflicts? 

Done

holachuy:
I patched and built locally and was able to bring up the UI and initiate the flow. This is more than enough to checkin and we can further test a round trip once setting up the appropriate test accounts after checking in. Thanks!

PR #652, opened 2019-02-11
Publish auth

wmorland:
:libraries:auth was missing from Bintray


katrielalex:
BTW I grepped through to make sure there weren't any more missing from Bintray; I think this is the last of them. He says optimistically.

Issue #586, opened 2018-09-24
Add support for Social Posts


harlantwood:
Hi, great project!  I installed the demo server from docker and it asks you to select your data types from a dropdown list of:

- Calendar
- Contacts
- Mail
- Photos
- Tasks

The data type that is missing that I was hoping for is "posts" or similar -- eg FB posts, tweets, G+ posts, etc.

Does that exist, but it's not in the demo server?  If it doesn't exist, is it planned and is there an expected timeline?  

Excited about this project, thanks for your work!

harlantwood:
A bit of extra info -- possible data formats for such a featute include:
- https://schema.org/SocialMediaPosting
- https://www.w3.org/TR/activitystreams-core/

PR #653, opened 2019-02-11
Add a Blogger importer for Social-Data

bwillard:
This creates a blogger imported for social data, storing photos in Drive. (This is related to #586)

There are also some related changes around G+ and Mastodon exports.

### DIFF
--- a/distributions/demo-server/build.gradle
+++ b/distributions/demo-server/build.gradle
@@ -184,4 +184,15 @@ task installToolsAndDockerize {
     dependsOn ":client-rest:installTools", dockerize
 }
 
+configurations.all {
+    resolutionStrategy {
+        eachDependency { DependencyResolveDetails details ->
+            if (details.requested.group == 'com.google.guava') {
+                details.useVersion("${guavaVersion}")
+                details.because 'Some dependencies rely on lower version which does not support MoreCollections'
+
+            }
+        }
+    }
+}
 
--- a/extensions/auth/portability-auth-google/src/main/java/org/datatransferproject/auth/google/GoogleOAuthConfig.java
+++ b/extensions/auth/portability-auth-google/src/main/java/org/datatransferproject/auth/google/GoogleOAuthConfig.java
@@ -55,6 +55,7 @@ public class GoogleOAuthConfig implements OAuth2Config {
         .put("CONTACTS", ImmutableSet.of("https://www.googleapis.com/auth/contacts.readonly"))
         .put("MAIL", ImmutableSet.of("https://www.googleapis.com/auth/gmail.readonly"))
         .put("PHOTOS", ImmutableSet.of("https://www.googleapis.com/auth/photoslibrary.readonly"))
+        // For G+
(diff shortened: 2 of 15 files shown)


distributions/demo-server/build.gradle
```diff
@@ -184,4 +184,15 @@ task installToolsAndDockerize {
     dependsOn ":client-rest:installTools", dockerize
 }
 
+configurations.all {
```
holachuy: Is this related to this PR?

extensions/auth/portability-auth-google/src/main/java/org/datatransferproject/auth/google/GoogleOAuthConfig.java
```diff
@@ -57,6 +55,7 @@ public String getTokenUrl() {
         .put("CONTACTS", ImmutableSet.of("https://www.googleapis.com/auth/contacts.readonly"))
         .put("MAIL", ImmutableSet.of("https://www.googleapis.com/auth/gmail.readonly"))
         .put("PHOTOS", ImmutableSet.of("https://www.googleapis.com/auth/photoslibrary.readonly"))
+        .put("SOCIAL-POSTS", ImmutableSet.of("https://www.googleapis.com/auth/plus.login"))
```
holachuy: Does this work for both Blogger and Gplus? Or not used for both? Let's add a comment clarifying.

extensions/data-transfer/portability-data-transfer-google/build.gradle
```diff
@@ -37,6 +38,10 @@ dependencies {
     compile("com.googlecode.ez-vcard:ez-vcard:${ezVcardVersion}")
     compile("com.google.apis:google-api-services-plus:${googlePlusVersion}")
 
+    // Needed for OpenSocial ActivityStreams
+    compile group: 'joda-time', name: 'joda-time', version: '2.3'
```
holachuy: Do we want to make the version a constant in the global version file?

distributions/demo-server/build.gradle
bwillard: Only tangentially, you can't run DTP right now with out this change, so I included it in this PR

extensions/auth/portability-auth-google/src/main/java/org/datatransferproject/auth/google/GoogleOAuthConfig.java
bwillard: Just G+, added a comment.

extensions/data-transfer/portability-data-transfer-google/build.gradle
bwillard: No, Activity stream depends on this specific version of joda-time, So I don't think we want to generalize it.

bwillard:
PTAL?

extensions/data-transfer/portability-data-transfer-google/build.gradle
holachuy: Ok, let's add a comment

bwillard:
PTAL?

PR #656, opened 2019-02-18
Video datatype

CorneliusHagmeister:
This pull request introduces a video datatype in order to prepare for video exporters/importers. Due to the similarity between photos and videos (both MediaItems) the video data type is defined analog to what was already present for photos. Video-importer and -exporter for Google Photos and Facebook will follow in future pull requests.

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-spotify/src/main/java/org/datatransferproject/transfer/spotify/playlists/SpotifyPlaylistImporter.java
+++ b/extensions/data-transfer/portability-data-transfer-spotify/src/main/java/org/datatransferproject/transfer/spotify/playlists/SpotifyPlaylistImporter.java
@@ -65,10 +65,10 @@ public class SpotifyPlaylistImporter
   private void createPlaylist(MusicPlaylist playlist, String userId)
       throws IOException, SpotifyWebApiException {
     Playlist createPlaylistResult = spotifyApi
-        .createPlaylist(userId, playlist.headline())
+        .createPlaylist(userId, playlist.getHeadline())
         .collaborative(false)
         .public_(false)
-        .name("Imported - " + playlist.headline())
+        .name("Imported - " + playlist.getHeadline())
         .build()
         .execute();
     for (MusicRecording track : playlist.getTrack()) {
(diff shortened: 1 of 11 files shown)


portability-types-common/src/main/java/org/datatransferproject/types/common/models/videos/VideoModel.java
```diff
@@ -0,0 +1,102 @@
...
+ */
+
+package org.datatransferproject.types.common.models.videos;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.MoreObjects;
+import com.google.common.base.Objects;
+
+public class VideoModel {
```
holachuy: For the video content, can we use the schema.org model https://schema.org/VideoObject ?

CorneliusHagmeister: Hm, I am not sure what the way to go is here. When I implemented the model I had a look the schema.org model and decided to keep the implementation consistent with the already existing photo model. Nevertheless it might make sense for the future to go with the schema.org version.

holachuy: Yeah, the original photo model was based on a POJO approach. However, playlists and other more recent models have been able to incorporate some Schema.org models.  Once we incorporate video playlists, we're going to end up having to merge these two. 

Could we use the schema.org model just for the video object for this initial CL  since it's only a few fields we'd want to map for the moment?

CorneliusHagmeister: Adjusted the Model to be based on the schema.org model. We should probably think about the folder structure for the models, especially in regards of dependencies (e.g. in which folder should a MediaObject go?).

holachuy: LGTM. Free free to open an issue for the data model organization.

PR #659, opened 2019-02-21
Switch Google Photos import scope to appendonly

wmorland:
I've tested this and import still works fine for me. This limits the scope and changes the text a user sees on the OAuth screen.

screenshot 2019-02-05 at 17 04 50


holachuy:
Thanks for finding this.  I thought there was guidance for using the most limited scopes in our docs but it doesn't seem to be there. We'll want to include this when we update the docs https://github.com/google/data-transfer-project/blob/master/Documentation/Integration.md#auth-extension

PR #665, opened 2019-02-26
Imgur importer

natlg:
Added Imgur importer ( #641) and documentation.

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosImporter.java
+++ b/extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosImporter.java
@@ -161,6 +161,7 @@ public class ImgurPhotosImporter
 
     if (!Strings.isNullOrEmpty(albumId)) {
       TempPhotosData tempData = jobStore.findData(jobId, createCacheKey(), TempPhotosData.class);
+      // Get previously saved id of imported album
       String newAlbumId = tempData.lookupNewAlbumId(albumId);
       builder.add("album", newAlbumId);
     }
--- a/extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosImporter.java
+++ b/extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosImporter.java
@@ -17,6 +17,7 @@
 package org.datatransferproject.datatransfer.imgur.photos;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import com.google.common.io.ByteStreams;
 import okhttp3.*;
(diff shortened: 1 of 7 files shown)


extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosImporter.java
```diff
@@ -0,0 +1,194 @@
...
+    FormBody.Builder builder = new FormBody.Builder().add("title", album.getName());
+    if (!Strings.isNullOrEmpty(description)) {
+      builder.add("description", description);
+    }
+    RequestBody formBody = builder.build();
+    requestBuilder.post(formBody);
+
+    Response response = client.newCall(requestBuilder.build()).execute();
+    int code = response.code();
+    if (code >= 200 && code <= 299) {
```
seehamrun: Nit: can we wrap this check in a precondition instead of the if else block? 

Same with the check of body == null and Strings.isNullOrEmpty(newAlbumId)

extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosImporter.java
```diff
@@ -0,0 +1,194 @@
...
+
+    Request.Builder requestBuilder = new Request.Builder().url(UPLOAD_PHOTO_URL);
+    requestBuilder.header("Authorization", "Bearer " + authData.getAccessToken());
+
+    FormBody.Builder builder = new FormBody.Builder().add("image", imageData);
+
+    if (!Strings.isNullOrEmpty(albumId)) {
+      TempPhotosData tempData = jobStore.findData(jobId, createCacheKey(), TempPhotosData.class);
+      // Get previously saved id of imported album
+      String newAlbumId = tempData.lookupNewAlbumId(albumId);
```
seehamrun: Is it possible that this album Id doesnt exist?

extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosImporter.java

extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosImporter.java
natlg: Yes, it can be null for non-album photos, then it will be imported as non-album photo too

seehamrun: ah okay so setting album to the null value below in builder.add("album", newAlbumId) won't cause issues? or should we add an ifstatement around it?

natlg: Sorry, I thought you mentioned the old album id.
It's not likely that id of newly created album will be null if old album exists, as during import it [checks](https://github.com/google/data-transfer-project/blob/563e1d62e3325e75fee96add98eb0ea6bc3e363b/extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosImporter.java#L128) that new album is created. But I added [ifstatement](https://github.com/google/data-transfer-project/blob/563e1d62e3325e75fee96add98eb0ea6bc3e363b/extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosImporter.java#L164) in case of `newAlbumId` is null for any reason

seehamrun: Thanks!

seehamrun:
Could you update and push the branch - thanks!

natlg:
It's up to date with master now

natlg:
Please note some restrictions for image upload: maximum size of image is 10MB for API upload (though with their site it's 20MB) and upload limit of 50 images per hour

PR #669, opened 2019-03-07
Photos logging

olsona:
Adding some debug logging to Google Photos Exporter and Flickr Importer


portability-api-launcher/src/main/java/org/datatransferproject/api/launcher/Datum.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.datatransferproject.api.launcher;
+
+/**
+ * An interface representing a data point object, such as a timestamp or a timeseries element.
+ */
+public interface Datum {
```
rtannenbaum: is this still needed? i don't see it used

olsona: Thought I'd deleted that file, whoops!

PR #672, opened 2019-03-07
Fixing some problems with titles and descriptions

olsona:
It wasn't clear whether these fixes should happen on the importer or exporter side.  I went with importer side because that's where the titles of albums and photos are prepended with COPY_PREFIX (something I'm not sure we should do).

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-flickr/src/main/java/org/datatransferproject/datatransfer/flickr/photos/FlickrPhotosImporter.java
+++ b/extensions/data-transfer/portability-data-transfer-flickr/src/main/java/org/datatransferproject/datatransfer/flickr/photos/FlickrPhotosImporter.java
@@ -155,8 +155,12 @@ public class FlickrPhotosImporter implements Importer<AuthData, PhotosContainerR
       // with these (in case the album exists later).
       Preconditions.checkArgument(album != null, "Album not found: " + oldAlbumId);
 
-      Photoset photoset =
-          photosetsInterface.create(COPY_PREFIX + album.getName(), album.getDescription(), photoId);
+      // TODO: do we want to keep the COPY_PREFIX?  I feel like not
+      String albumName =
+          Strings.isNullOrEmpty(album.getName()) ? "" : COPY_PREFIX + album.getName();
+      String albumDescription = cleanString(album.getDescription());
+
+      Photoset photoset = photosetsInterface.create(albumName, albumDescription, photoId);
       monitor.debug(() -> String.format("%s: Flickr importer created album: %s", id, album));
 
       // Update the temp mapping to reflect that we've created the album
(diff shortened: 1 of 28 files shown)


extensions/data-transfer/portability-data-transfer-flickr/src/main/java/org/datatransferproject/datatransfer/flickr/photos/FlickrPhotosImporter.java
```diff
@@ -155,8 +155,12 @@ private void importSinglePhoto(UUID id, PhotoModel photo) throws FlickrException
       // with these (in case the album exists later).
       Preconditions.checkArgument(album != null, "Album not found: " + oldAlbumId);
 
-      Photoset photoset =
-          photosetsInterface.create(COPY_PREFIX + album.getName(), album.getDescription(), photoId);
+      // TODO: do we want to keep the COPY_PREFIX?  I feel like not
+      String albumName =
+          Strings.isNullOrEmpty(album.getName()) ? "" : COPY_PREFIX + album.getName();
```
holachuy: Does Flickr support importing an empty album name? or should we put in 'untitled'.  What happens if we get more than one empty album name?

extensions/data-transfer/portability-data-transfer-flickr/src/main/java/org/datatransferproject/datatransfer/flickr/photos/FlickrPhotosImporter.java
```diff
@@ -172,14 +176,20 @@ private void importSinglePhoto(UUID id, PhotoModel photo) throws FlickrException
 
   private String uploadPhoto(PhotoModel photo, UUID jobId) throws IOException, FlickrException {
     BufferedInputStream inStream = imageStreamProvider.get(photo.getFetchableUrl());
+
+    // TODO: do we want to keep COPY_PREFIX?  I think not
+    String photoTitle =
+        Strings.isNullOrEmpty(photo.getTitle()) ? "" : COPY_PREFIX + photo.getTitle();
```
holachuy: Same question regarding importing empty titles or if we should add untitled.

extensions/data-transfer/portability-data-transfer-flickr/src/main/java/org/datatransferproject/datatransfer/flickr/photos/FlickrPhotosImporter.java
```diff
@@ -195,6 +205,10 @@ private String createCacheKey() {
     return "tempPhotosData";
   }
 
+  private String cleanString(String string) {
```
holachuy: this can be static as it doesn't rely on class member state

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosExporter.java
```diff
@@ -312,7 +312,7 @@ private PhotoModel convertToPhotoModel(Optional<String> albumId, GoogleMediaItem
     Preconditions.checkArgument(mediaItem.getMediaMetadata().getPhoto() != null);
 
     return new PhotoModel(
-        "", // TODO: no title?
+        "", // TODO: no title?  Should we use the description instead?
```
holachuy: Is there a name?

extensions/data-transfer/portability-data-transfer-flickr/src/main/java/org/datatransferproject/datatransfer/flickr/photos/FlickrPhotosImporter.java
olsona: Flickr does support empty titles.

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosExporter.java
olsona: There is not.

extensions/data-transfer/portability-data-transfer-flickr/src/main/java/org/datatransferproject/datatransfer/flickr/photos/FlickrPhotosImporter.java
olsona: If there isn't a name in Google Photos, Google's API exports the album name as "Untitled".  There can be two albums with the same name in Flickr, and the photos are separated appropriately (no cross-contamination).

PR #673, opened 2019-03-07
update job as IN_PROGRESS once processing starts 

seehamrun:


portability-transfer/src/main/java/org/datatransferproject/transfer/JobProcessor.java
```diff
@@ -133,13 +135,22 @@ private AuthDataDecryptService getAuthDecryptService(String scheme) {
...
+
+  private void updateJobState(UUID jobId, State state) {
     PortabilityJob existingJob = store.findJob(jobId);
     PortabilityJob updatedJob = existingJob.toBuilder().setState(state).build();
 
     try {
       store.updateJob(jobId, updatedJob);
     } catch (IOException e) {
-      monitor.debug(() -> format("Could not mark job %s as finished.", jobId));
+      monitor.debug(() -> format("Could not mark job %s as %s", jobId, state));
```
rtannenbaum: can you log the exception here too?

portability-transfer/src/main/java/org/datatransferproject/transfer/JobProcessor.java
```diff
@@ -133,13 +135,22 @@ private AuthDataDecryptService getAuthDecryptService(String scheme) {
 
   private void markJobFinished(UUID jobId, boolean success) {
     State state = success ? State.COMPLETE : State.ERROR;
+    updateJobState(jobId, state);
+
```
rtannenbaum: nit: extra line?

portability-transfer/src/main/java/org/datatransferproject/transfer/JobProcessor.java
```diff
@@ -75,11 +75,13 @@ void processJob() {
     boolean success = false;
     UUID jobId = JobMetadata.getJobId();
     monitor.debug(() -> format("Begin processing jobId: %s", jobId));
+    markJobStarted(jobId);
     hooks.jobStarted(jobId);
 
     PortabilityJob job = store.findJob(jobId);
     JobAuthorization jobAuthorization = job.jobAuthorization();
     Preconditions.checkState(jobAuthorization.state() == JobAuthorization.State.CREDS_STORED);
```
rtannenbaum: i think CREDS_STORED should be a pre-condition of job-started.. WDYT of checking CREDS_STORED as a validator to the update in markJobStarted?

that will also do it as part of the same transaction so we may be able to save a transaction

there's an example at https://github.com/google/data-transfer-project/blob/228378c5d92cd5853e26080f4328161f6c154a97/portability-api/src/main/java/org/datatransferproject/api/action/transfer/ReserveWorkerAction.java#L64

portability-transfer/src/main/java/org/datatransferproject/transfer/JobProcessor.java
seehamrun: Done

portability-transfer/src/main/java/org/datatransferproject/transfer/JobProcessor.java
seehamrun: Done

portability-transfer/src/main/java/org/datatransferproject/transfer/JobProcessor.java
seehamrun: Good point - WDYT of this?

rtannenbaum: nice! thanks for getting the other place too

PR #675, opened 2019-03-11
Logging monitoring data, pt 1

olsona:
This adds another logging command to the Monitor interface, one specifically designed for logging monitoring data (timestamps, etc).  This way, each host can make their own Monitor implementation to send the monitoring data along to their monitoring framework of choice.  

The new command logs a Datum instance.  Right now there is only one implementing class, but it should cover a range of cases.  I'm not yet entirely clear on how we should log cumulative data - I suspect it will be a bit complicated.

This also illustrates a usage of the InstantaneousDatum class, logging timestamps for when jobs change status.


olsona:
I feel like pushing Datum and its extensions down into specific monitoring implementations will make it harder to monitor all the things we want - we'd lose a lot of flexibility, I think.  Perhaps I'm not seeing how, for example, you'd monitor timestamps of job status change if SinglePointLongDatum weren't available to JobProcessor.

jimmarino:
Hi,

For that case the approach could be to issue a job status change event (per Will's work). The event object would not need to include a timestamp because the monitor implementation could be responsible for generating it. A lot of metrics solutions have libraries for this type of stuff so delegating it down into the monitoring implementation would avoid an "extra hop" of going from Datum to the other metrics types. The monitor impl could also track things like moving averages, etc.

What do you think?

wmorland:
I think it does make sense to use EventCodes to identify the events that we want to benchmark metrics against. I would have thought a monitor implementation could then generate the necessary timestamp once it receives the EventCode and any Datums.

I think the Datum class does makes sense as a way to pass through specific metrics, such as a count of how many items have been transferred or a number of iterations, but maybe not for timing between two events which could be done by the monitor based on a starting EventCode and a finishing EventCode.

jimmarino:
> I think it does make sense to use EventCodes to identify the events that we want to benchmark metrics against. I would have thought a monitor implementation could then generate the necessary timestamp once it receives the EventCode and any Datums.
> 
> I think the Datum class does makes sense as a way to pass through specific metrics, such as a count of how many items have been transferred or a number of iterations, but maybe not for timing between two events which could be done by the monitor based on a starting EventCode and a finishing EventCode.

That sounds like a good approach to me.

olsona:
Sure, that sounds fine.  I'll revert this change.

PR #678, opened 2019-03-13
Add support for retrieving the current project version and Git hash

jimmarino:
This PR uses JGit during build to generate a launcher properties file which will be packaged in the API Launcher JAR (or runtime uber jar). The class Version reads the properties file and retrieves the project version and Git hash. The Runtime bootstrap can then access that information.

I updated the SingleVM bootstrapper to output the version and has to the monitor. Other bootstrappers could do the same.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -36,12 +36,16 @@
  *
  * Note that if you add a submodule and want it published with official releases, add the 'maven-publish' plugin to the submodule configuration.
  */
+import org.eclipse.jgit.api.Git
+import org.eclipse.jgit.lib.RepositoryBuilder
+
 buildscript {
     repositories {
         jcenter()
     }
     dependencies {
         classpath "io.spring.gradle:spring-bintray-plugin:0.11.1"
+        classpath 'org.eclipse.jgit:org.eclipse.jgit:5.2.1.201812262042-r'
     }
 }
 
@@ -52,6 +56,8 @@ plugins {
 ext.nodeType = hasProperty('nodeType') ? nodeType : 'local'
 ext.cloudType = hasProperty('cloudType') ? cloudType : 'local'
 println "Node installation: ${ext.nodeType}, Cloud type: ${ext.cloudType}";
+
+setGitVersion()
 
 def sourceProjects() {
     // define projects containing source code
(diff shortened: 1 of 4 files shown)


holachuy:
Also, I added this manually to the root build.gradle file when I generated version 0.2.1.  Not sure if we can get jgit to generate the property and fill it in.

root.appendNode('properties').appendNode('build-commit', 'b509b928244e03fcd0b54b93881d04f7799cc789')

jimmarino:
> Very cool. Do we want to move this to a higher level so it can be used in each package that's generated?

It's available from the launcher API module so anything that bootstraps a runtime (API server, worker) or provides a Monitor implementation will have access to the Version class. I'm not sure if anything else would need access to it.

jimmarino:
> Also, I added this manually to the root build.gradle file when I generated version 0.2.1. Not sure if we can get jgit to generate the property and fill it in.
> 
> root.appendNode('properties').appendNode('build-commit', 'b509b928244e03fcd0b54b93881d04f7799cc789')

Yeah, we should either be able to get rid of this or use the code snippet in the API launcher project script to retrieve that information programmatically. Where is that property written to?

holachuy:
The thought was to provide the build commit for packaging and import verification purposes.  

I just added the like to the pom.withXml clause in the build.gradle file so it's written to the generated pom file in the properties stanza, e.g.

https://bintray.com/data-transfer-project/data-transfer-project/portability-transfer/0.2.1#files/org%2Fdatatransferproject%2Fportability-transfer%2F0.2.1


  <properties>
    <build-commit>b509b928244e03fcd0b54b93881d04f7799cc789</build-commit>
  </properties>

jimmarino:
> The thought was to provide the build commit for packaging and import verification purposes.
> 
> I just added the like to the pom.withXml clause in the build.gradle file so it's written to the generated pom file in the properties stanza, e.g.
> 
> https://bintray.com/data-transfer-project/data-transfer-project/portability-transfer/0.2.1#files/org%2Fdatatransferproject%2Fportability-transfer%2F0.2.1
> 
> [b509b92](https://github.com/google/data-transfer-project/commit/b509b928244e03fcd0b54b93881d04f7799cc789)

Ah, so maybe we could add the hash to an extension property off of the Gradle project object and then any script in the build can access it? That way the POM file being written can use it as well as the launcher API to write out the properties file. Is that what you were thinking?

wmorland:
Thanks for this @jimmarino! It looks good. I'm happy to defer to @holachuy, the POM stuff sounds like it could be useful.

holachuy:
For the jar metadata use case, I don't think we need the git hash anywhere in the runtime, just the pom will do. If other folks need it on the runtime, we can add it to the launcher code but not sure how useful that is.

PR #680, opened 2019-03-19
Add a idempotent executor for import jobs

bwillard:
This will allow extensions to be implemented in such a way as to prevent duplicated object in the case of retries (within a job).

Also cleans up music vertical to ensure a id is provided for each item to work in the new system.


extensions/data-transfer/portability-data-transfer-deezer/src/main/java/org/datatransferproject/transfer/deezer/playlists/DeezerPlaylistExporter.java
```diff
@@ -91,9 +93,10 @@ private PlaylistContainerResource enumeratePlaylists(DeezerApi api)
   private MusicRecording convertTrack(DeezerApi api, long trackId) throws IOException {
     Track track = api.getTrack(trackId);
     return new MusicRecording(
+        track.getLink(),
```
holachuy: Do you want to add the music related changes to the PR description as a subpoint?

extensions/data-transfer/portability-data-transfer-flickr/src/main/java/org/datatransferproject/datatransfer/flickr/photos/FlickrPhotosImporter.java
```diff
@@ -142,38 +142,57 @@ private void importSinglePhoto(UUID id, PhotoModel photo) throws FlickrException
...
-      tempData.removeTempPhotoAlbum(oldAlbumId);
+      createAlbum(idempotentExecutor, jobId, oldAlbumId, photoId);
     } else {
       // We've already created the album this photo belongs in, simply add it to the new album
       photosetsInterface.addPhoto(newAlbumId, photoId);
     }
+  }
 
-    jobStore.update(id, createCacheKey(), tempData);
+  private void createAlbum(
```
holachuy: Not now but it might be nice to have this call a method creating the album without the idempotentExcecutor

portability-transfer/src/main/java/org/datatransferproject/transfer/InMemoryIdempotentImportExecutor.java
```diff
@@ -0,0 +1,65 @@
...
+import org.datatransferproject.spi.transfer.provider.IdempotentImportExecutor;
+
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.Callable;
+
+public class InMemoryIdempotentImportExecutor implements IdempotentImportExecutor {
+  final private Map<String, Serializable> knownValues = new HashMap<>();
```
holachuy: Quick comment on visibility purpose

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleJobStore.java
```diff
@@ -237,7 +238,7 @@ public UUID findFirst(JobAuthorization.State jobState) {
   }
 
   @Override
-  public <T extends DataModel> void create(UUID jobId, String key, T model) throws IOException {
```
rtannenbaum: i'd prefer to avoid changing the interface here so we know only DataModels are still being stored in JobStore.

If the executor interface requires serializable, can you convert to DataModel inside?

portability-spi-cloud/src/main/java/org/datatransferproject/spi/cloud/storage/JobStore.java
```diff
@@ -74,21 +75,21 @@
    */
   UUID findFirst(JobAuthorization.State jobState);
 
-  default <T extends DataModel> void create(UUID jobId, String key, T model) throws IOException {
+  default <T extends Serializable> void create(UUID jobId, String key, T model) throws IOException {
```
holachuy: Hmm, do we need to know the type is a DataModel vs arbitrary data?

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleJobStore.java
```diff
@@ -264,7 +265,7 @@ public UUID findFirst(JobAuthorization.State jobState) {
   }
 
   @Override
-  public <T extends DataModel> void update(UUID jobId, String key, T model) {
+  public <T extends Serializable> void update(UUID jobId, String key, T model) {
```
rtannenbaum: didn't look too closely so not sure it's possible, but do you think it makes sense to do any refactoring/visibility changes to ensure operations always go thru the executor when intended, and not thru jobstore directly?

this may make more sense if you make an idempotent exporter as well - which doesn't matter for the end result of the transfer but could preserve quota/resources from exporting the same thing twice

portability-spi-cloud/src/main/java/org/datatransferproject/spi/cloud/storage/JobStore.java
bwillard: I don't think so, it hasn't been a problem in my testing at all.

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleJobStore.java
bwillard: Why? Specifically why do we care what data is stored in job store? Limiting it to just DataModels prevents us from storing all kind of useful things like simple types and types from sdks.

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleJobStore.java
bwillard: Do you mean to lock down job store so integratos can't use it?

Flickr still has a fairly valid use case for it I think, so we couldn't lock it down from everyone.

bwillard:
PTAL

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleJobStore.java
holachuy: So I guess it's a difference between a very open API contract with undefined objects or more well-defined objects to store.

As far as tradeoffs: 
-If we define the contract narrowly using DataModel, are we doing unnecessary object conversions to fit the model?
- If we allow any Serializable object, are we open to misuse or less optimal use?

I don't mind at this point. My inclination is to leave it well-defined until we see a use case where it's a hindrance.

rtannenbaum: ah ok, sounds like there are arguments for opening it up, just wanted to flag that it was a change from previous behavior that I think was there for a reason. I remember there being some earlier discussions with Jim about this but do not recall the details / history of the design.

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleJobStore.java
rtannenbaum: yea like lock down jobstore so integrators are forced to go thru the executor/caching layer. would it make sense for flickr to use the executor directly?

bwillard: No their use case is on the export side so it wouldn't make sense to use the executor.

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleJobStore.java
bwillard: So for what it's worth it is a hindrance now.  I'll make the change back just so we can get this in and then evaluate the change to serializable on its own but examples where it is a problem:

1) Existing behavior with temp data objects being created and stored one per job is a huge anti-pattern that was created because you couldn't store simple data in the data store.

2) This current change wouldn't be able to use the JobStore easily without wrapping everything in a generic wrapper, which I guess I can do but that seems really goony.

bwillard:
Ok I switched JobStore back to needed a DataModel, want to take another look?

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleJobStore.java
rtannenbaum: oh did you see my other point about also doing an export executor?

the idea being that it could preserve quota/resources from exporting the same thing twice (though it doesn't affect the end result data like the importer does)

portability-spi-transfer/src/main/java/org/datatransferproject/spi/transfer/types/TempPhotosData.java
```diff
@@ -19,19 +19,17 @@
...
 
+import java.io.Serializable;
+import java.util.*;
+
 /*
  * TempPhotosData used to store album and photos information before they are ready to be uploaded.
  */
 @JsonTypeName("org.dataportability:TempPhotosData")
-public class TempPhotosData extends DataModel {
+public class TempPhotosData extends DataModel implements Serializable {
```
rtannenbaum: would it be simpler to make DataModel implement Serializable? Then I think the subclasses get it for free?

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleJobStore.java
bwillard: oh I didn't, ya we should do one of those as well, but I think that is best left to a different PR as this one is already huge.  Flickr also has a case where they need to use the jobstore in the importer as well, and the executor is not a good fit for the use case.

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleJobStore.java
```diff
@@ -36,6 +36,11 @@
 import com.google.common.collect.ImmutableMap;
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
+import org.datatransferproject.spi.cloud.storage.JobStore;
```
rtannenbaum: if it's not too much a pain can you revert these files that only changed from the old serializable stuff?

i think you can just git checkout master them as one way to easily revert

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleJobStore.java
rtannenbaum: ack - in that case I think it's worth thinking about other ways to make it so integrators don't accidentally bypass the caching & double-import, but i'm not sure the best way. i'm fine with it as is for now, though.

portability-spi-transfer/src/main/java/org/datatransferproject/spi/transfer/types/TempPhotosData.java
bwillard: I made PortableType implement serializable so that should cover it.

bwillard:
One more look?

PR #681, opened 2019-03-25
Audiomack start

olsona:
Getting my Audiomack start in - still no API keys, so we can't test


seehamrun:
Due to inactivity Im closing this PR out, but created #833 to link it to in case someone wants to pick this work back up

AravindAkuthota:
yes i would definitely happy to work on it

PR #682, opened 2019-03-27
Fix JettyMonitor

wmorland:
When a RuntimeException is thrown on the API, JettyMonitor currently logs it as info rather than severe. This changes the mapping to make it severe.

Mapping everything to info makes it much harder to detect issues with the api.

### DIFF
--- a/extensions/transport/portability-transport-jettyrest/src/main/java/org/datatransferproject/transport/jettyrest/http/JettyMonitor.java
+++ b/extensions/transport/portability-transport-jettyrest/src/main/java/org/datatransferproject/transport/jettyrest/http/JettyMonitor.java
@@ -24,7 +24,7 @@ public class JettyMonitor implements Logger {
   }
 
   public void warn(String msg, Throwable thrown) {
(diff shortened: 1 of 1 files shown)


extensions/transport/portability-transport-jettyrest/src/main/java/org/datatransferproject/transport/jettyrest/http/JettyMonitor.java
```diff
@@ -60,7 +60,7 @@ public void debug(Throwable thrown) {
   }
 
   public void debug(String msg, Throwable thrown) {
-    DELEGATE.info(() -> msg, thrown);
+    DELEGATE.debug(() -> msg, thrown);
```
wmorland: Maybe there is a reason for this being escalated to info from debug. If so I'm happy to add a comment explaining why.

jimmarino: I think it was a cut-and-paste error :-)

PR #685, opened 2019-04-02
Imgur Photo Exporter test

natlg:
Added  tests for ImgurPhotoExporter ( #641)
Mockito doesn't allow to mock static fields or methods, so I refactored Exporter and Importer constructors to accept API URL instead of getting it from static variable. It was needed for mocking web server responses.
Also fixed Exception in the Importer (getting cache for a null value)


extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosExporter.java
```diff
@@ -54,14 +53,12 @@
...
     implements Exporter<TokensAndUrlAuthData, PhotosContainerResource> {
-  private static final String BASE_URL = ImgurTransferExtension.BASE_URL;
   private static final String RESULTS_PER_PAGE = "10";
-  private static final String ALBUM_PHOTOS_URL_TEMPLATE = BASE_URL + "/album/%s/images";
-  private static final String ALBUMS_URL_TEMPLATE =
-      BASE_URL + "/account/me/albums/%s?perPage=" + RESULTS_PER_PAGE;
-  private static final String ALL_PHOTOS_URL_TEMPLATE =
-      BASE_URL + "/account/me/images/%s?perPage=" + RESULTS_PER_PAGE;
-  private static final String DEFAULT_ALBUM_ID = "defaultAlbumId";
+  private static String albumPhotosUrlTemplate;
```
holachuy: Can we remove static and add final to these. That way they, after they are assigned in the constructor, the won't be modified.

extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosImporter.java
```diff
@@ -52,20 +52,24 @@
   private final ObjectMapper objectMapper;
   private final TemporaryPerJobDataStore jobStore;
   private final Monitor monitor;
-  private static final String BASE_URL = ImgurTransferExtension.BASE_URL;
-  private static final String CREATE_ALBUM_URL = BASE_URL + "/album";
-  private static final String UPLOAD_PHOTO_URL = BASE_URL + "/image";
+
+  private static String createAlbumUrl;
```
holachuy: Likewise, if we can remove static and make both of these final.

extensions/data-transfer/portability-data-transfer-imgur/src/test/java/org/datatransferproject/transfer/imgur/photos/ImgurPhotoExporterTest.java
```diff
@@ -0,0 +1,272 @@
...
+    assertEquals("Album 1", album1.getName());
+
+    PhotoAlbum album2 =
+        resource
+            .getAlbums()
+            .stream()
+            .filter(album -> "albumId2".equals(album.getId()))
+            .findFirst()
+            .get();
+    assertEquals("Album 2", album2.getName());
```
holachuy: This is nice. Do you think we can condense the code from 105 and 108-124 if we use the Truth library?

I was thinking something along the lines of #3 here (https://google.github.io/truth/)  using assertThat() containsExactly() and inOrder() along with your filter for album id?

extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosExporter.java
natlg: Done

extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosImporter.java

extensions/data-transfer/portability-data-transfer-imgur/src/test/java/org/datatransferproject/transfer/imgur/photos/ImgurPhotoExporterTest.java
natlg: I updated some tests to use Truth library. All the code from lines 115-135 could be changed to 
```
    assertThat(resource.getAlbums())
        .containsExactly(
            new PhotoAlbum("albumId1", "Album 1", null),
            new PhotoAlbum("albumId2", "Album 2", "Description for Album 2")).inOrder();
```
But PhotoAlbum's [equals method](https://github.com/google/data-transfer-project/blob/1e31ec8a70b04ac6ceee017e6a51b805d57aa16a/portability-types-common/src/main/java/org/datatransferproject/types/common/models/photos/PhotoAlbum.java#L58) checks only if `id`s are equal. So need to check name and description separately.
PhotoModels are compared by all fields so I condensed their tests.

natlg:
Hi Chuy, thanks for the review! I addressed comments, PTAL

Issue #41, opened 2018-01-11
Add metrics


rtannenbaum:
Different developers/hosts may want different libraries/platforms for metrics

We should probably make this an interface and include a GCP implementation to start

rtannenbaum:
Prometheus is a widely used monitoring system that integrates with Kubernetes, whether it's on GCP, AWS or Azure

OpenCensus is a brand new open source metrics solution from Google that exports to Prometheus, Stackdriver (used by GCP), and others (https://opensource.googleblog.com/2018/01/opencensus.html).

Whichever metrics system we end up using, we should make sure is compatible with Kubernetes autoscale (https://docs.bitnami.com/kubernetes/how-to/configure-autoscaling-custom-metrics/) for custom metrics. We'll want a custom metric for "worker has a job" and autoscale our workers based on whether they have a job (see #88)

PR #687, opened 2019-04-02
Add a Stats interface and framework

bwillard:
Re: #41 add a stats/metrics framework to DTP that can be used for both internal events as well extended for integrators to add their own stats.

Not sold on names for the clases, so please let me know if you have better ones.


bwillard:
Thanks Jim, I switched to ServiceExtension based model, and that is much cleaner.

Instead of going the Guice route I tried it out with a per service DelegatingExtensionContext what do you think?

portability-api-launcher/src/main/java/org/datatransferproject/api/launcher/DelegatingExtensionContext.java
```diff
@@ -0,0 +1,73 @@
...
+    return baseExtensionContext.cloud();
+  }
+
+  @Override
+  public Constants.Environment environment() {
+    return baseExtensionContext.environment();
+  }
+
+  public <T> void registerOverrideService(Class<T> type, T service){
+    overriddenRegisteredClasses.put(type, service);
```
holachuy: Do we want to do a contains check so as to not override more than once? Or warn if we do?

portability-transfer/src/main/java/org/datatransferproject/transfer/WorkerModule.java
```diff
@@ -130,7 +141,13 @@ Exporter getExporter(ImmutableList<TransferExtension> transferExtensions) {
   Importer getImporter(ImmutableList<TransferExtension> transferExtensions) {
     TransferExtension extension =
         findTransferExtension(transferExtensions, JobMetadata.getImportService());
-    extension.initialize(context);
+    DelegatingExtensionContext serviceSpecificContext = new DelegatingExtensionContext(context);
+    serviceSpecificContext.registerOverrideService(
+        MetricRecorder.class,
+        new ServiceAwareMetricRecorder(
+            extension.getServiceId(),
+            context.getService(DtpInternalMetricRecorder.class)));
```
holachuy: For clear error checking, do we need to do a precondition that DtpInternalMetricRecorder exists ?

portability-transfer/src/main/java/org/datatransferproject/transfer/WorkerModule.java
```diff
@@ -121,7 +126,13 @@ AppCredentialStore getBucketStore() {
   Exporter getExporter(ImmutableList<TransferExtension> transferExtensions) {
     TransferExtension extension =
         findTransferExtension(transferExtensions, JobMetadata.getExportService());
-    extension.initialize(context);
+    DelegatingExtensionContext serviceSpecificContext = new DelegatingExtensionContext(context);
+    serviceSpecificContext.registerOverrideService(
+        MetricRecorder.class,
+        new ServiceAwareMetricRecorder(
+            extension.getServiceId(),
+            context.getService(DtpInternalMetricRecorder.class)));
```
holachuy: For clear error checking, do we need to do a precondition that DtpInternalMetricRecorder exists ?

portability-api-launcher/src/main/java/org/datatransferproject/api/launcher/DelegatingExtensionContext.java
bwillard: I added a warning, FWIW, the underlying classes allow registrations to be overwritten.

portability-transfer/src/main/java/org/datatransferproject/transfer/WorkerModule.java
bwillard: Good point, instead of adding it here I added it in ServiceAwareMeticRecorder so it covers both cases.

portability-transfer/src/main/java/org/datatransferproject/transfer/WorkerModule.java
bwillard: Good point, instead of adding it here I added it in ServiceAwareMeticRecorder so it covers both cases.

bwillard:
Chuy want to take another look?

PR #689, opened 2019-04-12
Make the DTP framework call the appropriate metric hooks

bwillard:
Removes importFinished and exportFinished as because we interleave the processing they don't make sense.

This addresses issue #41.

### DIFF
--- a/portability-transfer/src/main/java/org/datatransferproject/transfer/CallableExporter.java
+++ b/portability-transfer/src/main/java/org/datatransferproject/transfer/CallableExporter.java
@@ -16,14 +16,17 @@
 
 package org.datatransferproject.transfer;
 
+import com.google.common.base.Stopwatch;
 import com.google.inject.Provider;
-import java.util.Optional;
-import java.util.UUID;
-import java.util.concurrent.Callable;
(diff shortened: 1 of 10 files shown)


portability-transfer/src/main/java/org/datatransferproject/transfer/CallableExporter.java
```diff
@@ -34,19 +37,36 @@
...
 
   @Override
   public ExportResult call() throws Exception {
-    return exporterProvider.get()
-        .export(jobId, authData, exportInformation);
+    boolean success = false;
+    Stopwatch stopwatch = Stopwatch.createStarted();
+    try {
+      ExportResult result =  exporterProvider.get().export(jobId, authData, exportInformation);
+      success = result.getType() != ExportResult.ResultType.ERROR;
```
holachuy: For consistency with the importer, we should probably use ==  ResultType.OK for the success condition here.

bwillard: Ah, for export there are 3 conditions: END, CONTINUE, and ERROR, so I think we have to do != ERROR, or == with an or, so I think the != is cleaner.

PR #693, opened 2019-04-22
Create a service level config framework and per user rate limits

bwillard:


portability-spi-transfer/src/main/java/org/datatransferproject/spi/transfer/config/TransferConfig.java
```diff
@@ -0,0 +1,5 @@
+package org.datatransferproject.spi.transfer.config;
+
+public interface TransferConfig {
```
holachuy: java doc

portability-transfer/src/main/java/org/datatransferproject/transfer/WorkerModule.java
```diff
@@ -197,4 +206,22 @@ Monitor getMonitor() {
...
+
+  private TransferServiceConfig getTransferServiceConfig(TransferExtension ext) {
+    String configFileName = "config/" + ext.getServiceId().toLowerCase() + ".yaml";
+    InputStream inputStream = this.getClass().getClassLoader().getResourceAsStream(configFileName);
+    getMonitor().info(
+        () -> "Service %s has a config file: %s",
+        ext.getServiceId(),
+        (inputStream != null));
+    if (inputStream == null) {
+      return TransferServiceConfig.getDefaultInstance();
```
holachuy: Let's log: using default transfer config for serviceId ...

bwillard: This is logged about the if statement.

portability-spi-transfer/src/main/java/org/datatransferproject/spi/transfer/config/TransferConfig.java
bwillard: I think we can just delete this file.

bwillard:
Thanks, added https://github.com/google/data-transfer-project/issues/694

Issue #670, opened 2019-03-07
PortabilityJob State is never persisted in any jobstore


seehamrun:
ex: GoogleJobStore updatejob uses .toMap() to write the job to datastore, but that doesnt populate the Jobstate so its never persisted. http://go/gh/google/data-transfer-project/blob/master/extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleJobStore.java#L175

When we use .fromMap() the jobstate is also always set to NEW

http://go/gh/google/data-transfer-project/blob/master/portability-spi-cloud/src/main/java/org/datatransferproject/spi/cloud/types/PortabilityJob.java#L136

rtannenbaum:
https://github.com/google/data-transfer-project/issues/191 coming back to bite us. really easy to drop fields.

PR #697, opened 2019-05-01
Make DTP store error information from import tasks

bwillard:
This change does a couple things:
 - define a data model to store errors
 - Add the ability to store errors to the JobStore
 - Change IdempotentImportExecutor to have 2 methods: one that throws on error and one that swallows errors and stores the error; this allows single item errors not to affect other imports.
 - Changes the framework to store any errors that happened during import in the job store.


portability-types-transfer/src/main/java/org/datatransferproject/types/transfer/errors/ErrorDetail.java
```diff
@@ -0,0 +1,80 @@
...
+  @JsonProperty("id")
+  public abstract String id();
+
+  @JsonProperty("title")
+  public abstract String title();
+
+  @JsonProperty("exception")
+  public abstract Exception exception();
+
+  public Map<String, Object> toMap() {
```
seehamrun: This has been something that's bitten us in the past, if we update the object to include more details, but we forget to update the to/from map then we never persist those errors #670 #191 (and im sure others). Can we switch to json serialization?

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleJobStore.java
```diff
@@ -351,12 +371,19 @@ private Entity createUpdatedEntity(Key key, Map<String, Object> data)
...
-  private Key getKey(UUID jobId) {
-    return datastore.newKeyFactory().setKind(KIND).newKey(jobId.toString());
+  private Key getJobKey(UUID jobId) {
+    return datastore.newKeyFactory().setKind(JOB_KIND).newKey(jobId.toString());
+  }
+
+  private Key getErrorKey(UUID jobId, String errorId) {
+    return datastore.newKeyFactory()
+        .setKind(ERROR_KIND)
+        .addAncestor(PathElement.of(JOB_KIND, jobId.toString()))
```
holachuy: I know it's trivial but let's comment or link to the Datastore doc describing this relationship

portability-transfer/src/main/java/org/datatransferproject/transfer/InMemoryIdempotentImportExecutor.java
```diff
@@ -17,31 +17,41 @@
...
-  public <T extends Serializable> T execute(
-      String idempotentId,
-      String itemName,
-      Callable<T> callable) throws IOException {
+  public <T extends Serializable> T executeAndSwallowExceptions(
+      String idempotentId, String itemName, Callable<T> callable) throws IOException {
+    try {
+      return executeOrThrowException(idempotentId, itemName, callable);
+    } catch (Exception e) {
+      return null;
```
holachuy: So the executeOrThrowException throw IOException.  Do we want to catch that and ignore but also have a block for a general Exception in case anything else bubbles up, especially if there are future additions to this code.

portability-types-transfer/src/main/java/org/datatransferproject/types/transfer/errors/ErrorDetail.java
bwillard: Good point, what do you think about this approach?

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleJobStore.java
```diff
@@ -182,6 +189,19 @@ public void updateJob(UUID jobId, PortabilityJob job, JobUpdateValidator validat
     }
   }
 
+  @Override
+  public void addErrorsToJob(UUID jobId, Collection<ErrorDetail> errors) throws IOException {
+    if (errors == null) {
```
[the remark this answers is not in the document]
bwillard: I was mostly worried about NPE when looping, but I added an empty check to avoid doing the batch insert call with no entries.

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleJobStore.java
bwillard: Done.

portability-transfer/src/main/java/org/datatransferproject/transfer/InMemoryIdempotentImportExecutor.java
bwillard: Ya I changed it so any exceptions that are throw in the throwing catch block get bubbled up.  Hopefully that never happens but it will be much less confusing if it does happen now.

bwillard:
PTAL?

portability-transfer/src/main/java/org/datatransferproject/transfer/JobProcessor.java
```diff
@@ -147,7 +154,13 @@ private AuthDataDecryptService getAuthDecryptService(String scheme) {
...
 
-  private void markJobFinished(UUID jobId, boolean success) {
+  private void markJobFinished(UUID jobId, boolean success, Collection<ErrorDetail> errors) {
+    try {
+      store.addErrorsToJob(jobId, errors);
+    } catch (IOException | RuntimeException e) {
+      success = false;
+      monitor.severe(() -> "Problem adding errors to JobStore: %s", e);
+    }
     State state = success ? State.COMPLETE : State.ERROR;
```
seehamrun: Did we want to introduce a new state if the job was partially complete?

portability-types-transfer/src/main/java/org/datatransferproject/types/transfer/errors/ErrorDetail.java
seehamrun: Nice! I like it

PR #705, opened 2019-05-29
refresh tokens in the GooglePhotosInterface and GoogleVidoesInterface

seehamrun:


extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosInterface.java
```diff
@@ -175,15 +174,19 @@ BatchMediaItemResponse createPhoto(NewMediaItemUpload newMediaItemUpload) throws
...
     }
-    if (!updatedParams.containsKey(ACCESS_TOKEN_KEY)) {
-      updatedParams.put(ACCESS_TOKEN_KEY, credential.getAccessToken());
+
+    // getAccessToken will return null when the token needs to be refreshed
+    if (credential.getAccessToken() == null) {
+      credential.refreshToken();
     }
 
+    updatedParams.put(ACCESS_TOKEN_KEY, credential.getAccessToken());
```
bwillard: Do you want to check that it is non-null and throw if it is, just in case we were unable to refresh it I think that would give a clearer error message. (and in the other file)

seehamrun: Good point, done

PR #707, opened 2019-06-03
log guice startup errors through monitor

holachuy:
Log guice startup errors through monitor. This helps with debugging errors at startup.

Sorry, not sure why the import order changes all of the sudden.

### DIFF
--- a/portability-api/src/main/java/org/datatransferproject/api/ApiMain.java
+++ b/portability-api/src/main/java/org/datatransferproject/api/ApiMain.java
@@ -15,6 +15,10 @@
  */
 package org.datatransferproject.api;
 
+import static org.datatransferproject.config.extension.SettingsExtensionLoader.getSettingsExtension;
+import static org.datatransferproject.launcher.monitor.MonitorLoader.loadMonitor;
+import static org.datatransferproject.spi.cloud.extension.CloudExtensionLoader.getCloudExtension;
+
 import com.google.api.client.http.HttpTransport;
 import com.google.api.client.http.javanet.NetHttpTransport;
 import com.google.api.client.json.JsonFactory;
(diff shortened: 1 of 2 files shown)


portability-api/src/main/java/org/datatransferproject/api/ApiMain.java
```diff
@@ -167,17 +166,22 @@ public void initializeHttps(
...
+                  typeManager,
+                  cloudExtension.getJobStore(),
+                  keyGenerator,
+                  trustManagerFactory,
+                  keyManagerFactory,
+                  authServiceExtensions,
+                  tokenManager,
+                  extensionContext));
+    } catch (Exception e) {
+      monitor.info(() -> "Unable to initialize Guice in Apia", e);
```
bwillard: Why just info? Seems like this should be a warning.

bwillard: Apia?

bwillard: Also why not re-throw the exception as well?  Won't this lead to a NPE or some other hard to understand error latter?

portability-transfer/src/main/java/org/datatransferproject/transfer/WorkerMain.java
```diff
@@ -123,17 +122,22 @@ public void initialize() {
...
+                  extensionContext,
+                  cloudExtension,
+                  transferExtensions,
+                  publicKeySerializers,
+                  decryptServices,
+                  symmetricKeyGenerator,
+                  asymmetricKeyGenerator,
+                  jobHooks));
+    } catch (Exception e) {
+      monitor.info(() -> "Unable to initialize Guice in Apia", e);
```
bwillard: All the same comments here.

portability-api/src/main/java/org/datatransferproject/api/ApiMain.java
holachuy: Done. Used severe.

holachuy: Done. Fixed typo.

holachuy: Done. Had it but lost it in copy and paste.

portability-transfer/src/main/java/org/datatransferproject/transfer/WorkerMain.java
holachuy: Done. Thx!

PR #712, opened 2019-06-13
use byte array for key abstraction

holachuy:
Use byte array for key abstraction instead of java crypto types. This allows for other libraries to be used for encryption and decryption.

### DIFF
--- a/portability-transfer/src/main/java/org/datatransferproject/transfer/JobMetadata.java
+++ b/portability-transfer/src/main/java/org/datatransferproject/transfer/JobMetadata.java
@@ -45,13 +45,13 @@ public final class JobMetadata {
 
   static void init(
       UUID initJobId,
-      byte[] encodedPrivateKey,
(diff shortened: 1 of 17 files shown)


portability-transfer/src/main/java/org/datatransferproject/transfer/JobMetadata.java
```diff
@@ -32,29 +29,29 @@
...
       UUID initJobId,
-      PrivateKey initPrivateKey,
+      byte[] initEncodedPrivateKey,
       String initDataType,
       String initExportService,
       String initImportService) {
     Preconditions.checkState(!isInitialized(), "JobMetadata cannot be initialized twice");
     jobId = initJobId;
-    privateKey = initPrivateKey;
+    encodedPrivateKey = initEncodedPrivateKey;
```
holachuy: Updated static field assignment.

PR #715, opened 2019-06-13
Fix typo

gamwe6:


gamwe6:
I signed the CLA

how-does-dtp-work.md
```diff
@@ -7,7 +7,7 @@ DTP comprises three main components:
 Data Models are the canonical formats that establish a common understanding of how to transfer data.
 Adapters provide a method for converting each Provider’s proprietary data and authentication formats into a form that is usable by the system.
 Task Management Library provides the plumbing to power the system.
-Data Models
```
seehamrun: Actually I think the right fix for this is to make it a header

### Data Models 

And the above should be a list (adapters and task management have a section below on lines 21 and 31)

gamwe6: Thank you @seehamrun for the review

seehamrun:
Hiya, can you check that you've signed the google CLA so we can merge this?

seehamrun:
Due to inactivity I've copied these changes into #831

PR #716, opened 2019-06-16
Fix code blocks

gamwe6:
Fix some code blocks. Add language identifier to enable syntax highlighting


gamwe6:
What about terminal command code blocks? Is there a convention we should use?
- One backtick or three backticks?
- With language identifier (e.g. `shell`, `console`)?
- With or without command prompt?

seehamrun:
Hiya since this hasnt been udpated in a while, im closing and merging it as #830

PR #718, opened 2019-07-03
Insert missing uses of String.format

wmorland:
Fix calls to Monitor where format was not being applied to strings with obvious String format expressions in them (e.g. %s)


wmorland:
@bwillard Monitors are not expected to do the formatting themselves (or at least that was my assumption based on [`ConsoleMonitor.java`](https://github.com/google/data-transfer-project/blob/487c5ea1676433fecfe63a3028426a9c743e8f46/portability-api-launcher/src/main/java/org/datatransferproject/launcher/monitor/ConsoleMonitor.java) and how we have implemented our monitor). The additional parameters are not intended for string formatting.

wmorland:
I signed it!

PR #719, opened 2019-07-08
Mark a job as failed if any errors are returned from the copier

wmorland:
Because most exceptions are caught by the IdepotentExecutor it has been possible for a job to have a bunch of errors, to the point that no photos were transferred, but it could still be marked as complete. This change aims to mark a job as failed if there are any errors left when a copy finishes. (If an import is retried for an item that had previously failed then that error will already be removed on a successful retry)


wmorland:
Do we know why we no longer have the automatic build check on this pull request? Is there an issue with the CI service?

holachuy:
I'm wondering if it's a delayed turn down of the Github integration services -> apps migration. I just applied to turn on the Travis CI "app" for the repo.

PR #729, opened 2019-07-26
Refactor JobStore

wmorland:
This makes sure the JobStore interface only depends on Plain Old Java Objects and does not receive logic dynamically via its arguments. This means we can significantly optimise our JobStore implementation.

Once I removed the concept of passed in validation it meant a mix of specific state transitions and generic updates. To improve validation it seemed like a logical step to make every call on the JobStore a specific state change. This changes the philosophy of the JobStore and is fairly opinionated. I'm happy to discuss/explain this more as it is significant change in that respect (event though actual behaviour should not change).

### DIFF
--- a/portability-spi-cloud/src/main/java/org/datatransferproject/spi/cloud/storage/JobStoreWithValidator.java
+++ b/portability-spi-cloud/src/main/java/org/datatransferproject/spi/cloud/storage/JobStoreWithValidator.java
@@ -1,0 +1,87 @@
+package org.datatransferproject.spi.cloud.storage;
+
+import static org.datatransferproject.spi.cloud.types.JobAuthorization.State.CREDS_AVAILABLE;
+
+import com.google.common.base.Preconditions;
+import com.google.common.base.Strings;
+import java.io.IOException;
+import java.util.UUID;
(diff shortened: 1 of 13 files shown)


portability-spi-cloud/src/main/java/org/datatransferproject/spi/cloud/storage/JobStoreWithValidator.java
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  /**
+   * Verifies a {@code PortabilityJob} already exists for {@code jobId}, and updates the entry to
+   * {@code job}.
+   *
+   * @throws IOException if a job didn't already exist for {@code jobId} or there was a problem
+   * updating it
+   */
+  protected abstract void updateJob(UUID jobId, PortabilityJob job) throws IOException;
```
holachuy: Should this be in the base JobStore or JobStore with validator?

wmorland: I moved this to JobStoreWithValidator so there would be no generic update job method without validation on the JobStore. This then pushes every call to the JobStore to be a well defined state transition.

PR #737, opened 2019-08-28
Pass in a jobId to the IdempotentImportExecutor

wmorland:
Summary: By allowing different implementations of the IdempotentImportExecutor and allowing it to be aware of the JobId we can create a IdempotentImportExecutor which allows recovery of a transfer in the instance where a transfer worker completely crashes or is unexpectedly terminated. We still load the InMemoryIdempotentImportExecutor by default in the instance when another implementation is not explicitly provided (this avoids breaking existing deployments).

Test Plan: Ran a transfer using the docker container version.


wmorland:
Pushed a new commit which incorporated chuy's comments.

PR #742, opened 2019-09-06
Count imported items in JobStore

dragosmartac:
The changes in this pull request are meant to allow the counting of the successfully imported items; The idea is to use a map containing pairs of the form (dataName, dataCount), which is being passed from the current imported DataModel to the ImportResult. Finally, the elements of the map are merged into an accumulator map inside the JobStore.

This changes make the class ImportResult mutable.

Happy to take any feedback and suggestions.


portability-spi-transfer/src/main/java/org/datatransferproject/spi/transfer/provider/ImportResult.java
```diff
@@ -47,6 +48,17 @@ public ResultType getType() {
...
   }
 
+  /**
+   * Returns the number of imported items
+   */
+  public Map<String, Integer> getCounts() {
+    return counts;
+  }
+
+  public void setCounts(Map<String, Integer> counts) {
```
holachuy: To keep this immutable as it's returned to callers, we can either

1) Create a new object ImportResultWithCounts (with a better name) that wraps the ImportResult and adds the count data then we return ImportResultWithCounts 

or

2) We can make setCounts into a static factory method, e.g. withCounts or addCounts that returns a ImportResult which is a copy of the current one but with count data added

Both of these are to keep a clean immutable contract to callers.

Thoughts?

portability-types-common/src/main/java/org/datatransferproject/types/common/models/photos/PhotosContainerResource.java
```diff
@@ -46,6 +48,14 @@ public PhotosContainerResource(
     return photos;
   }
 
+  @Override
+  public Map<String, Integer> getCounts() {
+    return new ImmutableMap.Builder<String, Integer>()
+        .put("photosCount", photos.size())
```
holachuy: Should these be in constants somewhere?

dragosmartac: Added them as static constants inside PhotosContainerResource

portability-spi-transfer/src/main/java/org/datatransferproject/spi/transfer/provider/ImportResult.java
dragosmartac: I think both options make sense. I tried to implement the second one because I thought it might be slightly better to not add a new level of indirection.

The changes are:

void set counts ->(transformed to) ImportResult copyWithCounts -> which calls a new constructor to create a shallow copy of the current object but with the field counts referencing the mapping given as an argument.

PR #743, opened 2019-09-10
Fix default failure reason

cnnorris:


portability-spi-cloud/src/main/java/org/datatransferproject/spi/cloud/types/PortabilityJob.java
```diff
@@ -259,6 +263,7 @@ public boolean equals(Object o) {
     @JsonProperty("lastUpdateTimestamp")
     public abstract Builder setLastUpdateTimestamp(LocalDateTime lastUpdateTimestamp);
 
+    @JsonInclude(JsonInclude.Include.NON_NULL)
```
seehamrun: what does this NON_NULL include do?

cnnorris: it only gets included if it's specified: https://fasterxml.github.io/jackson-annotations/javadoc/2.0.0/com/fasterxml/jackson/annotation/JsonInclude.Include.html

cnnorris: I copied it from l256

seehamrun: nifty!

PR #749, opened 2019-09-13
Fix bug in UniformRetryStategy and improve logging for retries

wmorland:
I found that it was difficult to tell from the logs which strategy was used for a given error so this logging should make it easier to do so and check that the expected strategies are being used.


portability-types-transfer/src/main/java/org/datatransferproject/types/transfer/retry/RetryingCallable.java
```diff
@@ -71,9 +71,17 @@ public T call() throws RetryException {
         // TODO: do we want to reset anything (eg, number of retries) if we see a different
         // RetryStrategy?
         RetryStrategy strategy = retryStrategyLibrary.checkoutRetryStrategy(e);
+        monitor.info(
```
bwillard: This feels more like a debug log to me, but I don't feel strongly about it if you want to leave it as info.

wmorland: That makes sense, we already have the severe on line 69

PR #751, opened 2019-09-19
Create smugmug albums more flexibly

cnnorris:
@seehamrun @holachuy 

I think this importer was making the assumption that every album would be created before trying to import any of the photos, but evidently data.getAlbums() does not cover all of the photo.getAlbumId() calls

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-smugmug/src/main/java/org/datatransferproject/transfer/smugmug/photos/SmugMugPhotosImporter.java
+++ b/extensions/data-transfer/portability-data-transfer-smugmug/src/main/java/org/datatransferproject/transfer/smugmug/photos/SmugMugPhotosImporter.java
@@ -116,11 +116,14 @@ public class SmugMugPhotosImporter
       PhotoModel inputPhoto,
       SmugMugInterface smugMugInterface)
       throws IOException {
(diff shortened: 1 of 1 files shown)


cnnorris:
Actually wait, maybe there was something else going on. Don't merge this yet.

extensions/data-transfer/portability-data-transfer-smugmug/src/main/java/org/datatransferproject/transfer/smugmug/photos/SmugMugPhotosImporter.java
```diff
@@ -116,11 +116,14 @@ String importSinglePhoto(
...
       throws IOException {
-    String newAlbumUri = idempotentExecutor.getCachedValue(inputPhoto.getAlbumId());
-    checkState(
-        !Strings.isNullOrEmpty(newAlbumUri),
-        "Cached album URI for %s is null",
-        inputPhoto.getAlbumId());
+    try {
+        String newAlbumUri = idempotentExecutor.getCachedValue(inputPhoto.getAlbumId());
+    } catch (IllegalArgumentException) {
+        String newAlbumUri = idempotentExecutor.executeAndSwallowIOExceptions(
```
holachuy: Let's comment. Create the album on demand as all albums are not returned in the initial call.

PR #754, opened 2019-09-22
[Fix] Handle case when there are multiple primary names 

sukeesh:
This PR handles the case when there are multiple primary names. Existing code will consider only one primary name and ignores others (if there are any).

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/contacts/GoogleContactsExporter.java
+++ b/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/contacts/GoogleContactsExporter.java
@@ -110,7 +110,8 @@ public class GoogleContactsExporter
   }
 
   private static void convertToVCardNamesAndPopulate(VCard vCard, List<Name> personNames) {
-    // TODO(olsona): what if there's more than one primary name in a Google Contact?
+    // If there are multiple primary names, we consider the first one as primary
+    // and add remaining primary names in alternateStructuredNames
     StructuredName primaryStructuredName = null;
     LinkedList<StructuredName> alternateStructuredNames = new LinkedList<>();
     for (Name personName : personNames) {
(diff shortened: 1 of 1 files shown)


sukeesh:
@olsona Can you please take a look at this PR? 😅

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/contacts/GoogleContactsExporter.java
```diff
@@ -116,7 +116,7 @@ private static void convertToVCardNamesAndPopulate(VCard vCard, List<Name> perso
     for (Name personName : personNames) {
       StructuredName structuredName = convertToVCardNameSingle(personName);
       Boolean isNamePrimary = personName.getMetadata().getPrimary();
-      if (isNamePrimary != null && isNamePrimary) {
+      if (primaryStructuredName == null && isNamePrimary != null && isNamePrimary) {
```
seehamrun: This overall is an OK change, but I'm wondering if theres some other structure we can use to capture this? The main difference that's happening here iiuc, is whether we use the first or the the last primary name we see. 

@holachuy  for input

holachuy: Seems strange to have more than one primary name. I don't oppose either case.

sukeesh:
@holachuy @seehamrun Thanks for the review. Can you please merge this? 🙈

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/contacts/GoogleContactsExporter.java
seehamrun: Could you update the comment on line 113 to state how this case is handled now?

seehamrun: Could you update the comment on line 116 to state how we handle this case?

seehamrun: Oops, I meant to update the existing comment on like 113 since its not longer applicable

sukeesh: Thanks @seehamrun, Can you please check now?

Issue #739, opened 2019-08-30
Automatically check google-java-format on pull requests


wmorland:
We recommend google-java-format in the developer documentation for the project but we do not currently automatically check contributions for code style. We should add a Travis CI step to check the formatting on each pull request.

seehamrun:
similar to #151

PR #757, opened 2019-09-25
Travis check google-java-format on all PRs

sukeesh:
This PR introduces travis check on all PRs. Adding this as the first step in CI Check.

Fixes this https://github.com/google/data-transfer-project/issues/739

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,13 @@
 #
 matrix:
   include:
+    - name: "Google java format validation"
+      # Validate with Google Java Style
+      language: java
+      before_script:
+        - wget https://github.com/google/google-java-format/releases/download/google-java-format-1.7/google-java-format-1.7-all-deps.jar
+      script:
+        - find . -iname "*.java" | xargs java -jar google-java-format-1.7-all-deps.jar --dry-run --set-exit-if-changed
     - name: "Java (Gradle)"
       # Nothing to customize here. Travis knows to use Gradle (including running tests) based on
       # the presence of gradle build file
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ matrix:
       before_script:
         - wget https://github.com/google/google-java-format/releases/download/google-java-format-1.7/google-java-format-1.7-all-deps.jar
       script:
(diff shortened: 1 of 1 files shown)


sukeesh:
@seehamrun Can you please review this?

.travis.yml
```diff
@@ -15,6 +15,13 @@
 #
 matrix:
   include:
+    - name: "Google java format validation"
+      # Validate with Google Java Style
+      language: java
+      before_script:
+        - wget https://github.com/google/google-java-format/releases/download/google-java-format-1.7/google-java-format-1.7-all-deps.jar
+      script:
+        - find . -iname "*.java" | xargs java -jar google-java-format-1.7-all-deps.jar --dry-run --set-exit-if-changed
```
seehamrun: It looks like this is currently failing travis?

`The command "find . -iname "*.java" | xargs java -jar google-java-format-1.7-all-deps.jar --dry-run --set-exit-if-changed" exited with 123.`

Can you verify that the --dry-run doesn't fail the build? I imagine a lot of our files will probably fail the google-java-formatter.

sukeesh: @seehamrun I might have misunderstood the issue.
Let me clarify 🙇 So, we need to add a step to check for google java format when someone makes a PR, right? 

**What I did**
`--dry-run` would just show a list of files which are not following the format. `--set-exit-if-changed` along with dry run option would exit with non-zero status code so that this can be caught during CI Check.

> It looks like this is currently failing Travis?

That's expected, right? because almost every file in this repo is not following the format 😢

seehamrun: Yes but we dont want to block development by this. Lets do it in 3 steps
- turn on but only in dry-run and with a 0 status code so travis passes
- fix all of the files and show that dry-run returned no changes 
- update it so that a non-zero code is returned if there are violating files.

sukeesh: Understood!

wmorland: One intermediate step we could do is only fail on the files that have been added or modified. Something like: `git diff --name-only --diff-filter=AM HEAD...$TRAVIS_BRANCH` and then filtering to just .java files. That way we can slowly migrate to the correct formatting and the blame is slightly less affected.

PR #758, opened 2019-09-25
Restore smugmug to functionality of v0.3.5

cnnorris:
Undoing the changes I thought would fix the special character issue.

This leaves the smugmug integration in a state where it will fail to create albums with special characters in them, but work otherwise

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-smugmug/src/main/java/org/datatransferproject/transfer/smugmug/photos/SmugMugInterface.java
+++ b/extensions/data-transfer/portability-data-transfer-smugmug/src/main/java/org/datatransferproject/transfer/smugmug/photos/SmugMugInterface.java
@@ -31,8 +31,6 @@ import com.google.common.net.HttpHeaders;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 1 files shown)


extensions/data-transfer/portability-data-transfer-smugmug/src/main/java/org/datatransferproject/transfer/smugmug/photos/SmugMugInterface.java
```diff
@@ -110,11 +108,14 @@ SmugMugAlbumsResponse getAlbums(String url) throws IOException {
   SmugMugAlbumResponse createAlbum(String albumName) throws IOException {
     // Set up album
     Map<String, String> json = new HashMap<>();
+    String niceName = "Copy-of-" + albumName.replace(' ', '-');
```
seehamrun: Should we just change it so it strips non a-z chars?

cnnorris: I don't want to do that until we talk to the smugmug team. I'm worried about other alphabets, and name overlaps.

For now this is just a reversion, I'll take another shot at making it work in the future.

PR #761, opened 2019-09-30
Update documentation to install stable node version

cnnorris:
`nvm install node` was installing node 12, which is not compatible with our project. (Something to do with https://github.com/sass/node-sass/issues/2662)


Documentation/Developer.md
```diff
@@ -36,7 +36,7 @@ From the top level data-transfer-project directory:
 * Install Git: `sudo apt-get install git-all`
 * Install Gradle: `sudo apt-get install gradle`
  * Install NVM: `curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.0/install.sh | bash`
- * Install Node: `nvm install node`
+ * Install Node: `nvm install v10.16.3`
```
cnnorris: I'm actually not 100% sure this command does what it's supposed to, I haven't actually run it on linux yet

cnnorris: Confirmed that this works

PR #771, opened 2019-10-15
support imported item counts in google jobstore

seehamrun:


seehamrun:
The tests I wrote fail in travis but run locally. Holding off on review until I figure out whats going on

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleJobStore.java
```diff
@@ -338,6 +339,46 @@ public void create(UUID jobId, String key, InputStream stream) {
     googleTempFileStore.create(jobId, key, stream);
   }
 
+  @Override
+  public void addCounts(UUID jobId, Map<String, Integer> newCounts) throws IOException {
+    if (newCounts == null) {
```
holachuy: Do we just want to make this a precondition or are we expecting null?

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleJobStore.java
```diff
@@ -338,6 +339,46 @@ public void create(UUID jobId, String key, InputStream stream) {
...
+
+    Transaction transaction = datastore.newTransaction();
+
+    for (String dataType : newCounts.keySet()) {
+      Key key = getCountsKey(jobId, dataType);
+      Entity current = datastore.get(key);
+      Integer oldCount = 0;
+
+      if (current != null && current.getNames().contains(COUNTS_FIELD)) {
+        oldCount = Math.toIntExact(current.getLong(COUNTS_FIELD));
```
holachuy: Is this cause Datastore only allows longs? If so, let's add a comment.

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleJobStore.java
seehamrun: The localjobstore implementation expects null sometimes >.> 
And we default to null here: https://github.com/google/data-transfer-project/blob/master/portability-transfer/src/main/java/org/datatransferproject/transfer/PortabilityInMemoryDataCopier.java#L164

We could change that to be an empty map instead

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleJobStore.java
seehamrun: Done

seehamrun:
Tests finally pass on travis. But keeping the travis change to use the --info flag for future use

PR #773, opened 2019-10-23
Add gettoken to authdata

wmorland:
Most subclasses of AuthData have a way to get a token. We currently use introspection to get tokens for revocation but we have to use java reflection. This would allow us to get the token for any AuthData.

It's a bit of a fudge for the CookiesAndUrlAuthData so I'm prepared to make an abstract subclass of AuthData if that is preferred.

### DIFF
--- a/portability-types-transfer/src/main/java/org/datatransferproject/types/transfer/auth/AuthData.java
+++ b/portability-types-transfer/src/main/java/org/datatransferproject/types/transfer/auth/AuthData.java
@@ -18,4 +18,6 @@ package org.datatransferproject.types.transfer.auth;
 import org.datatransferproject.types.common.PortableType;
 
 /** Base type for authentication data. */
-public abstract class AuthData extends PortableType {}
+public abstract class AuthData extends PortableType {
+  public abstract String getToken();
+}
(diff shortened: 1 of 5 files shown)


portability-types-transfer/src/main/java/org/datatransferproject/types/transfer/auth/AuthData.java
```diff
@@ -18,4 +18,6 @@
 import org.datatransferproject.types.common.PortableType;
 
 /** Base type for authentication data. */
-public abstract class AuthData extends PortableType {}
+public abstract class AuthData extends PortableType {
+  public abstract String getToken();
```
holachuy: Can we add a bit of documentation?  Also, do we want to return null or throw an exception for subclasses that do not support this?

wmorland: Throwing an exception may be the more sensible thing to do for subclasses that explicitly do not support this. That means it won't fail silently but it can be caught and handled if wanted.

wmorland: Pushed a commit to throw an UnsupportedOperationException

PR #775, opened 2019-10-31
Implement a GoogleCloudIdempotentImportExecutor that is backed up by datastore

seehamrun:
- Created a new GoogleCloudUtils class to move some utility functions from the GoogleJobStore
- Added test

### DIFF
--- a/extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleCloudIdempotentImportExecutor.java
+++ b/extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleCloudIdempotentImportExecutor.java
@@ -1,19 +1,23 @@
 package org.datatransferproject.cloud.google;
 
+import static java.lang.String.format;
+
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
-import com.google.cloud.datastore.*;
+import com.google.cloud.datastore.Datastore;
+import com.google.cloud.datastore.DatastoreException;
+import com.google.cloud.datastore.Entity;
+import com.google.cloud.datastore.Key;
+import com.google.cloud.datastore.Query;
+import com.google.cloud.datastore.QueryResults;
 import com.google.cloud.datastore.StructuredQuery.CompositeFilter;
 import com.google.cloud.datastore.StructuredQuery.PropertyFilter;
+import com.google.cloud.datastore.Transaction;
 import com.google.common.base.Joiner;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
-import org.datatransferproject.api.launcher.Monitor;
-import org.datatransferproject.spi.transfer.idempotentexecutor.IdempotentImportExecutor;
-import org.datatransferproject.types.transfer.errors.ErrorDetail;
-
 import java.io.IOException;
 import java.io.Serializable;
 import java.util.Collection;
@@ -21,8 +25,9 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.Callable;
-
-import static java.lang.String.format;
+import org.datatransferproject.api.launcher.Monitor;
+import org.datatransferproject.spi.transfer.idempotentexecutor.IdempotentImportExecutor;
+import org.datatransferproject.types.transfer.errors.ErrorDetail;
 
 public class GoogleCloudIdempotentImportExecutor implements IdempotentImportExecutor {
 
@@ -36,12 +41,13 @@ public class GoogleCloudIdempotentImportExecutor implements IdempotentImportExec
 
   private final Datastore datastore;
   private final Monitor monitor;
+  private final ObjectMapper objectMapper;
 
+  // These are all variables corresponding to the job state. Only initialized when setJobId() is called
   private Map<String, Serializable> knownValues;
   private Map<String, ErrorDetail> errors;
   private UUID jobId;
   private String jobIdPrefix;
-  private final ObjectMapper objectMapper;
 
   public GoogleCloudIdempotentImportExecutor(Datastore datastore, Monitor monitor) {
     this.datastore = datastore;
@@ -147,6 +153,7 @@ public class GoogleCloudIdempotentImportExecutor implements IdempotentImportExec
 
   @Override
   public void setJobId(UUID jobId) {
+    Preconditions.checkNotNull(jobId);
     this.jobId = jobId;
     this.knownValues = getKnownValuesForJob(jobId);
     this.errors = getErrorDetailsForJob(jobId);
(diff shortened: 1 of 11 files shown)


extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleCloudIdempotentImportExecutor.java
```diff
@@ -0,0 +1,233 @@
...
+  private static final String JOB_ID_FIELD = "job_id";
+  private static final String IDEMPOTENT_ID_FIELD = "idempotent_id";
+  private static final String ERROR_FIELD = "error_details";
+
+  private final Datastore datastore;
+  private final Monitor monitor;
+
+  private Map<String, Serializable> knownValues;
+  private Map<String, ErrorDetail> errors;
+  private UUID jobId;
```
holachuy: Let's comment on when this is intialized

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleCloudIdempotentImportExecutor.java
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  @Override
+  public Collection<ErrorDetail> getErrors() {
+    return ImmutableList.copyOf(errors.values());
+  }
+
+  @Override
+  public void setJobId(UUID jobId) {
+    this.jobId = jobId;
```
holachuy: Should we check not null since these are lazy intialized member vars

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleCloudIdempotentImportExecutor.java
```diff
@@ -0,0 +1,233 @@
...
+    this.knownValues = getKnownValuesForJob(jobId);
```
holachuy: If most executions don't generate errors, this may be prematurely incurring a cost each time. What if we lazy init these for when they're needed?

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleCloudUtils.java
```diff
@@ -0,0 +1,46 @@
...
+import com.google.cloud.datastore.Blob;
+import com.google.cloud.datastore.Entity;
+import com.google.cloud.datastore.Key;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.ObjectOutputStream;
+import java.util.Map;
+
+public class GoogleCloudUtils {
+
```
holachuy: We should add a private no op constructor for best practices

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleCloudUtils.java
```diff
@@ -0,0 +1,46 @@
...
+  // Environment variable where GCP project ID is stored. The value is set in
+  // config/k8s/api-deployment.yaml.
+  private static final String GCP_PROJECT_ID_ENV_VAR = "GOOGLE_PROJECT_ID";
+
+  static Entity.Builder createEntityBuilder(Key key, Map<String, Object> data) throws IOException {
```
holachuy: Let's add quick documentation to this, it's useful in many places.

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleCloudIdempotentImportExecutor.java
seehamrun: Done

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleCloudIdempotentImportExecutor.java
seehamrun: Done

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleCloudIdempotentImportExecutor.java
seehamrun: Im not sure what you mean here, we would need to read these for every call to executeAndSwallowIOExceptions and executeOrThrowExceptions so that we are able to update the map correctly. 

Are you saying that we read from datastore for each of the idempotentIds as we get them? Wouldn't that result in a more filtered queries to datastore?

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleCloudUtils.java
seehamrun: Done

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleCloudUtils.java
seehamrun: Done

extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleCloudIdempotentImportExecutor.java
holachuy: Ah right, let's add a comment, that this lookup is once per executor.

seehamrun: Done

seehamrun:
I've added tests for this class PTAL

extensions/cloud/portability-cloud-google/src/test/java/org/datatransferproject/cloud/google/GoogleCloudIdempotentImportExecutorTest.java
```diff
@@ -0,0 +1,84 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+public class GoogleCloudIdempotentImportExecutorTest {
+
+  private static final String ITEM_NAME = "item1";
+  private static final Monitor monitor = Mockito.mock(Monitor.class);
+
+  private static final UUID JOB_ID = UUID.randomUUID();
```
holachuy: I wonder if we should initialize this is the setUp method in case we ever want to have more job id's. Or we can wait for the time that happens.

seehamrun: I added a test to make sure we dont load any items from a different JOB_ID, but I think we can still keep it as final static since this test is pretty straightforward. If we want to expand later we can

PR #777, opened 2019-11-05
Add http response content to exception in GooglePhotosInterface

wmorland:
Adds the content of an HTTPResponseException to the IOException thrown as it was difficult to debug the root cause without the content.

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosInterface.java
+++ b/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosInterface.java
@@ -162,20 +162,11 @@ public class GooglePhotosInterface {
     try {
       response = getRequest.execute();
     } catch (HttpResponseException e) {
-      // if the response is "unauthorized" and we've successfully refreshed the token, try the
-      // request again
-      if (e.getStatusCode() == 401 && credential.refreshToken()) {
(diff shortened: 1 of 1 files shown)


extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosInterface.java
```diff
@@ -229,6 +205,31 @@ BatchMediaItemResponse createPhoto(NewMediaItemUpload newMediaItemUpload) throws
...
+    if (e.getStatusCode() == 401) {
+      monitor.info(() -> "Attempting to refresh authorization token");
+      // if the credential refresh failed, let the error bubble up via the IOException that gets
+      // thrown
+      credential = credentialFactory.refreshCredential(credential);
+      monitor.info(() -> "Refreshed authorization token successfuly");
+
+      // if the second attempt throws an error, then something else is wrong, and we bubble up the
+      // response errors
+      return httpRequest.execute();
```
seehamrun: I think you need to recraft the request here (or update the access token that is in the request) because the credential has been refreshed.

wmorland: Good catch!

wmorland:
I now build the httpRequest via a supplier which will run after the credentials have been updated.

Issue #774, opened 2019-10-29
Microsoft Photos Importer doesn't refresh authorization token upon expiration


seehamrun:
https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow#refresh-the-access-token

https://github.com/google/data-transfer-project/blob/master/extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/datatransferproject/transfer/microsoft/photos/MicrosoftPhotosImporter.java#L187-L189

Though this might get tricky since according to the docs the server may issue a new refresh token. In that case we will have to update the credentials in the jobstore

cnnorris:
Fixed with #778

PR #778, opened 2019-11-05
Add token refreshing to ms photos importer

cnnorris:
@seehamrun

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/datatransferproject/transfer/microsoft/photos/MicrosoftPhotosImporter.java
+++ b/extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/datatransferproject/transfer/microsoft/photos/MicrosoftPhotosImporter.java
@@ -16,6 +16,7 @@
 package org.datatransferproject.transfer.microsoft.photos;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.api.client.auth.oauth2.Credential;
 import com.google.common.base.Strings;
 import okhttp3.MediaType;
 import okhttp3.OkHttpClient;
@@ -62,7 +63,7 @@ public class MicrosoftPhotosImporter implements Importer<TokensAndUrlAuthData, P
   private final TemporaryPerJobDataStore jobStore;
   private final Monitor monitor;
   private final MicrosoftCredentialFactory credentialFactory;
-  private final Credential credential;
+  private Credential credential;
 
   private final String createFolderUrl;
   private final String uploadPhotoUrlTemplate;
@@ -156,7 +157,6 @@ public class MicrosoftPhotosImporter implements Importer<TokensAndUrlAuthData, P
   private String importSinglePhoto(
       PhotoModel photo,
       UUID jobId,
-      Credential credential,
       IdempotentImportExecutor idempotentImportExecutor)
       throws IOException {
     InputStream inputStream = null;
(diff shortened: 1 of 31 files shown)


extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/datatransferproject/transfer/microsoft/photos/MicrosoftPhotosImporter.java
```diff
@@ -187,6 +192,17 @@ private String importSinglePhoto(
         if (code < 200 || code > 299) {
           throw new IOException("Got error code: " + code + " message " + response.message());
         }
+        if (code == 401){
+            // If there was an unauthorized error, then try refreshing the creds
```
holachuy: Should we add a TODO to add retry?

cnnorris: What do you mean?

extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/datatransferproject/transfer/microsoft/photos/MicrosoftPhotosImporter.java
```diff
@@ -187,6 +192,17 @@ private String importSinglePhoto(
...
+            credentialFactory.refreshCredential(credentialFactory.createCredential(authData));
+            monitor.info(() -> "Refreshed authorization token successfuly");
+
+            Response newResponse = client.newCall(requestBuilder.build()).execute();
```
seehamrun: Is the new token attached here? Or do we have to recreate the request?

seehamrun: FYI Will did this with a Supplier in #777 which is pretty neat

seehamrun:
#774

extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/datatransferproject/transfer/microsoft/photos/MicrosoftPhotosImporter.java
cnnorris: I believe it's attached: https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow#successful-response-2

Should I do something with it? just add it to the AuthData field?

seehamrun: Oh sorry I meant, is it attched to the requestBuilder? I think it may have the older invalid credential and needs to be rebuilt. 

The older cred is added on line 181:  `requestBuilder.header("Authorization", "Bearer " + authData.getAccessToken()); ` I think we just need to replace the header.

extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/datatransferproject/transfer/microsoft/photos/MicrosoftPhotosImporter.java
```diff
@@ -187,6 +195,18 @@ private String importSinglePhoto(
         if (code < 200 || code > 299) {
           throw new IOException("Got error code: " + code + " message " + response.message());
         }
+        if (code == 401){
+            // If there was an unauthorized error, then try refreshing the creds
```
seehamrun: Did you mean to switch the conditions?

cnnorris: That would be a good idea wouldn't it.

PR #780, opened 2019-11-07
Return correct photo ID in ms photo importer

cnnorris:
@seehamrun

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/datatransferproject/transfer/microsoft/photos/MicrosoftPhotosImporter.java
+++ b/extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/datatransferproject/transfer/microsoft/photos/MicrosoftPhotosImporter.java
@@ -190,10 +190,9 @@ public class MicrosoftPhotosImporter implements Importer<TokensAndUrlAuthData, P
       // PUT the stream
       requestBuilder.put(body);
 
-      ResponseBody responseBody;
       try (Response response = client.newCall(requestBuilder.build()).execute()) {
         int code = response.code();
-        responseBody = response.body();
+        ResponseBody responseBody = response.body();
         if (code == 401){
             // If there was an unauthorized error, then try refreshing the creds
             credentialFactory.refreshCredential(credential);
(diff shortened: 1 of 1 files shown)


extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/datatransferproject/transfer/microsoft/photos/MicrosoftPhotosImporter.java
```diff
@@ -200,12 +201,21 @@ private String importSinglePhoto(
...
         if (code < 200 || code > 299) {
           throw new IOException("Got error code: " + code + " message " + response.message());
         }
-        // TODO return photo ID
-        return "fakeId";
+
+        // Extract photo ID from response body
+        if (body == null) {
+          throw new IOException("Got null body");
+        }
```
seehamrun: Can we make this a precondition check instead?

extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/datatransferproject/transfer/microsoft/photos/MicrosoftPhotosImporter.java
```diff
@@ -200,12 +201,21 @@ private String importSinglePhoto(
             requestBuilder.header("Authorization", "Bearer " + credential.getAccessToken());
             Response newResponse = client.newCall(requestBuilder.build()).execute();
             code = newResponse.code();
+            body = newResponse.body();
```
seehamrun: responseBody?

cnnorris:
Re: album Id, we do that on lines 142-149, that's where I got  this code from

extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/datatransferproject/transfer/microsoft/photos/MicrosoftPhotosImporter.java
cnnorris: good catch

Issue #783, opened 2019-11-11
Can't access API after following "Running Locally" instructions?


gobengo:
I followed these instructions for [Running Locally](https://github.com/google/data-transfer-project/blob/master/Documentation/RunningLocally.md).

(i.e. I ran `docker run --rm -p 3000:443 -p 5005:5005 -p 8080:8080 --env-file distributions/demo-server/env.secrets --name dtp-demo datatransferproject/demo` after creating env.secrets from env.secrets.template)

The app boots fine, I think.

In the ["Interact with the Application" section of that page](https://github.com/google/data-transfer-project/blob/master/Documentation/RunningLocally.md#interact-with-the-application), it says:
> The API is accessible via https://localhost:8080/_/listDataTypes.

However, I can't access that URL:
```
bengo@LAPTOP-9VBP4TSS:~$ curl https://localhost:8080/_/listDataTypes -k
{ error: '404'}
```

From [JettyApiSpec.yaml](https://github.com/google/data-transfer-project/blob/master/Documentation/JettyApiSpec.yaml#L17) and some source code, I can see that maybe `/_/listDataTypes` is an old URL and now it should be `/datatypes`, but that 404s for me too:
```
bengo@LAPTOP-9VBP4TSS:~$ curl https://localhost:8080/datatypes -k
{ error: '404'}
bengo@LAPTOP-9VBP4TSS:~$ curl https://localhost:8080/_/datatypes -k
{ error: '404'}
```

Am I missing something obvious?
If not, what sort of curl should work to request the data types API?

If I can get it working, I will make a PR to update the "Running Locally" document for what works for me.

Thank you.

gobengo:
I found out that the following works:

```
bengo@LAPTOP-9VBP4TSS:~$ curl -k https://localhost:8080/api/datatypes
{"dataTypes":["PLAYLISTS","PHOTOS","BLOBS","CALENDAR","CONTACTS","MAIL","SOCIAL-POSTS","TASKS","VIDEOS"]}
```

You can also request the same path through port 3000 (containerPort 443), because nginx will proxy_pass it to the api server.
```
bengo@LAPTOP-9VBP4TSS:~$ curl -k https://localhost:3000/api/datatypes
{"dataTypes":["PLAYLISTS","PHOTOS","BLOBS","CALENDAR","CONTACTS","MAIL","SOCIAL-POSTS","TASKS","VIDEOS"]}
```

[...]

PR #784, opened 2019-11-11
Update RunningLocally.md to use latest /datatypes URL

gobengo:
Addresses https://github.com/google/data-transfer-project/issues/783


mikeadams1:
@googlebot can you verify that I signed it?

Issue #786, opened 2019-11-12
WORKER_JOB_ERRORED when transfering Contacts Google -> Microsoft


gobengo:
I configured oauth credentials for Google and Microsoft using the demo docker container, and tried to make a contacts transfer using the demo UI. The Oauth flows succeded, and there was no error in the demo UI, but I see the following in the logs right away.

Any idea why?

```
DEBUG Polling for a job in state CREDS_AVAILABLE[0m
DEBUG Found job 6b4a4d1f-e27c-418e-9974-25fa68c52ea3[0m
DEBUG JWEKeyGenerator generate[0m
DEBUG JWEKeyGenerator generated WorkerKeyPair[0m
DEBUG JobPollingService: tryToClaimJob: jobId: PortabilityJob{state=NEW, exportService=Google, importService=Microsoft, transferDataType=CONTACTS, exportInformation=null, createdTimestamp, lastUpdateTimestamp, jobAuthorization=JobAuthorization{state=CREDS_AVAILABLE, encryptedInitialExportAuthData=null, encryptedInitialImportAuthData=null, encryptedAuthData=null, encryptionScheme=jwe, sessionSecretKey=LfI8dQcnl0sCm9doLjwHbg==, authPublicKey=null, instanceId=null}, failureReason=null}[0m
DEBUG Stored updated job: tryToClaimJob: jobId: PortabilityJob{state=NEW, exportService=Google, importService=Microsoft, transferDataType=CONTACTS, exportInformation=null, createdTimestamp, lastUpdateTimestamp, jobAuthorization=JobAuthorization{state=CREDS_AVAILABLE, encryptedInitialExportAuthData=null, encryptedInitialImportAuthData=null, encryptedAuthData=null, encryptionScheme=jwe, sessionSecretKey=LfI8dQcnl0sCm9doLjwHbg==, authPublicKey=null, instanceId=null}, failureReason=null}[0m
DEBUG Stored updated job: tryToClaimJob: JobMetadata initialized: 6b4a4d1f-e27c-418e-9974-25fa68c52ea3[0m
DEBUG Updated job 6b4a4d1f-e27c-418e-9974-25fa68c52ea3 to CREDS_ENCRYPTION_KEY_GENERATED, publicKey length: 162[0m
DEBUG JobMetadata.isInitialized(): true[0m
DEBUGollUntilJobIsReady[0m
[...]

PR #787, opened 2019-11-14
Fix AuthData serialization issues.

wmorland:
Do not serialize the getToken methods on AuthData classes where it is a proxy for another field. Also adds in a test that would have picked up this bug.

Should fix issue https://github.com/google/data-transfer-project/issues/786

### DIFF
--- a/portability-types-transfer/src/test/java/org/datatransferproject/types/transfer/auth/AuthDataSerializationTest.java
+++ b/portability-types-transfer/src/test/java/org/datatransferproject/types/transfer/auth/AuthDataSerializationTest.java
@@ -1,0 +1,87 @@
+package org.datatransferproject.types.transfer.auth;
+
+import static org.junit.Assert.*;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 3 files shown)


portability-types-transfer/src/test/java/org/datatransferproject/types/transfer/auth/AuthDataSerializationTest.java
```diff
@@ -0,0 +1,90 @@
...
+    final String url = "https://www.example.com/auth";
+    final CookiesAndUrlAuthData authData = new CookiesAndUrlAuthData(cookies, url);
+
+    final String s = objectMapper.writeValueAsString(authData);
+    final AuthData readValue = objectMapper.readValue(s, AuthData.class);
+    assertTrue(
+        "The read AuthData should be an instance of CookiesAndUrlAuthData",
+        readValue instanceof CookiesAndUrlAuthData);
+    final CookiesAndUrlAuthData readAuthData = (CookiesAndUrlAuthData) readValue;
+    assertEquals("Expect cookies to be the same", cookies, readAuthData.getCookies());
```
holachuy: Eventually we may wan to add equals methods on these objects but this is a great start for now.

PR #792, opened 2019-11-27
Fix GoogleVideosImporter

wmorland:
This changes GoogleVideosImporter to use IdempotentImportExecutor for the full operation. Also corrects createVideo to map the response to BatchMediaItemResponse. Also google-java-format-1.7.

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/videos/GoogleVideosImporter.java
+++ b/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/videos/GoogleVideosImporter.java
@@ -35,9 +35,13 @@ import com.google.api.client.auth.oauth2.Credential;
 import com.google.api.client.json.JsonFactory;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
+import java.io.InputStream;
+import java.util.Collections;
+import java.util.UUID;
 import org.datatransferproject.api.launcher.Monitor;
 import org.datatransferproject.datatransfer.google.common.GoogleCredentialFactory;
 import org.datatransferproject.datatransfer.google.mediaModels.NewMediaItem;
+import org.datatransferproject.datatransfer.google.mediaModels.NewMediaItemResult;
 import org.datatransferproject.datatransfer.google.mediaModels.NewMediaItemUpload;
 import org.datatransferproject.spi.transfer.idempotentexecutor.IdempotentImportExecutor;
 import org.datatransferproject.spi.transfer.provider.ImportResult;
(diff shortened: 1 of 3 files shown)


extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/videos/GoogleVideosImporter.java
```diff
@@ -127,27 +122,26 @@ void importSingleVideo(
...
     NewMediaItemUpload uploadItem =
-            new NewMediaItemUpload(null, Collections.singletonList(newMediaItem));
+        new NewMediaItemUpload(null, Collections.singletonList(newMediaItem));
 
-    executor.executeAndSwallowIOExceptions(
-            inputVideo.getDataId(),
-            inputVideo.getName(),
-            () -> getOrCreateVideosInterface(authData).createVideo(uploadItem));
+    final NewMediaItemResult result =
+        getOrCreateVideosInterface(authData).createVideo(uploadItem).getResults()[0];
```
holachuy: Do we want to ensure the result array is not empty?

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/videos/GoogleVideosImporter.java
```diff
@@ -127,27 +122,26 @@ void importSingleVideo(
...
   private synchronized GoogleVideosInterface getOrCreateVideosInterface(
-          TokensAndUrlAuthData authData) {
+      TokensAndUrlAuthData authData) {
     return videosInterface == null ? makeVideosInterface(authData) : videosInterface;
   }
 
   private synchronized GoogleVideosInterface makeVideosInterface(TokensAndUrlAuthData authData) {
     Credential credential = credentialFactory.createCredential(authData);
-    GoogleVideosInterface videosInterface = new GoogleVideosInterface(credential, this.jsonFactory);
+    videosInterface = new GoogleVideosInterface(credential, this.jsonFactory);
```
holachuy: Should we just inline with the return statement below?

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/videos/GoogleVideosImporter.java
wmorland: The GooglePhotosImporter makes the same assumption and we do check for success within the VideosInterface. Might make a separate fix which handles both the Videos and the Photos Importer.

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/videos/GoogleVideosImporter.java
wmorland: This is setting the field on the class rather than a local variable so that it will be reused.

wmorland:
Yes, I've tested this and the IdempotentImportExecutor catches the exceptions from `uploadVideoContent` for example which it wasn't before.

PR #793, opened 2019-12-05
Use google photoslibary for importing video (+ upgrade Mockito)

wmorland:
We were seeing issues with the GoogleVideosImporter and we have resolved them by switching to using the official Google Photos Library. I believe this is better handling the resumable uploads for the video content and it resolves our issues. To support mocking a final class I also updated the version of Mockito to the latest version and resolved any deprecations.


extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/videos/GoogleVideosImporter.java
```diff
@@ -94,54 +99,99 @@ public ImportResult importItem(
...
 
-    inputStream = this.videoStreamProvider.get(inputVideo.getContentUrl().toString());
-
     String filename;
     if (Strings.isNullOrEmpty(inputVideo.getName())) {
       filename = "untitled";
     } else {
       filename = COPY_PREFIX + inputVideo.getName();
     }
+    File tmp = File.createTempFile(filename, ".mp4");
```
bwillard: Hmm, this just uses a local file, I don't think we make any guarantees/assumptions of how much disk there is on the worker, it seems like you should temp store here right?

wmorland: Yes, that certainly seemed like it would be preferable but UploadMediaItemRequest only supports reading the data from a RandomAccessFile which limits our options. https://google.github.io/java-photoslibrary/1.4.0/com/google/photos/library/v1/upload/UploadMediaItemRequest.Builder.html I don't know if this might be suitable for a later improvement to the Google Photos Library.

bwillard: It seems like the thing we might have to do then is add a method to our JobStore that produces a RandomAccessFile then, that way it can get stored in the persistent storage provided by the cloud extension.

wmorland: Ok, I've created a default implementation of getTempFileFromInputStream on TemporaryPerJobDataStore.

PR #798, opened 2019-12-18
Log the code that is returned on smugmug failures

seehamrun:


extensions/data-transfer/portability-data-transfer-smugmug/src/main/java/org/datatransferproject/transfer/smugmug/photos/SmugMugInterface.java
```diff
@@ -247,7 +246,8 @@ public InputStream getImageAsStream(String urlStr) {
     Response response = request.send();
     if (response.getCode() < 200 || response.getCode() >= 300) {
       throw new IOException(
-          String.format("Error occurred in request for %s : %s", fullUrl, response.getMessage()));
+          String.format("Error occurred in request for %s, code: %s, message: %s", fullUrl,
```
cnnorris: Could we include the request too?

seehamrun: Done - logged it if its a bad request error code

PR #801, opened 2019-12-20
Fix Demo Server environment.apiBaseUrl

### LINKED ISSUE #800: Demo server unable to list data types
While running the demo server, after clicking "Next: Choose Data", we get a 404 error calling the url: https://localhost:3000//api/datatypes.

Screen Shot 2019-12-19 at 11 05 13 PM

Upon investigation, #791 changes the environment variable for apiBaseUrl by attempting to get document.baseURI, which has a trailing slash in all browser implementations I tried (Firefox, Chrome, Safari).

The fix is to clean up the url with a regex that removes trailing slashes.

gpriess:
Fixes #800 

With this change, API calls will now be formed by a base URL that is cleaned up to remove any trailing slashes.

### DIFF
--- a/client-rest/src/environments/environment.ts
+++ b/client-rest/src/environments/environment.ts
@@ -5,5 +5,5 @@
 export const environment = {
     production: false,
     encryptionScheme: "jwe", // supported values are "jwe" and "cleartext" which must correspond to the schemes supported by the server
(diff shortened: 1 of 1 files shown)


wmorland:
I've tested this change and it looks to fix the issue. I'll update the branch.

PR #803, opened 2019-12-20
ignore flickr errors that are caused by the photo already being in the set

seehamrun:


extensions/data-transfer/portability-data-transfer-flickr/src/main/java/org/datatransferproject/datatransfer/flickr/photos/FlickrPhotosImporter.java
```diff
@@ -124,6 +123,8 @@ public ImportResult importItem(
         } catch (FlickrException e) {
           if (e.getMessage().contains("Upload limit reached")) {
             throw new DestinationMemoryFullException("Flickr destination memory reached", e);
+          } else if (e.getMessage().contains("Photo already in set")) {
```
holachuy: Eventually let's put these in constants

seehamrun: Ack - i added a comment to explain why this isn't a "Real" error

PR #816, opened 2020-01-08
Use album id and photo id as the idempotent ids in MicrosoftPhotosImporter

seehamrun:
We were using the hashcode before which caused the IDs to be different when jobs were restarted. Both the albumid and photoid are used because we need a way to store photos in multiple albums, this is already being done in the flickr and google importers: 

https://github.com/google/data-transfer-project/blob/master/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosImporter.java#L102

https://github.com/google/data-transfer-project/blob/master/extensions/data-transfer/portability-data-transfer-flickr/src/main/java/org/datatransferproject/datatransfer/flickr/photos/FlickrPhotosImporter.java#L154


seehamrun:
I'm not sure what you mean - we allow for the multiple album support by also using the album id in the idempotent id. This will mean that we will upload a photo more than once if it is in more than 1 album. 

We shouldnt ever overwrite any photos in albums (if they have the same title) because each one has a different photo id from the Google Photos Exporter and if that does happen, Microsoft will just rename the file to be something like "original (1)"

pablomd314:
Ok, nevermind, I was confused about the issue.

PR #820, opened 2020-01-08
fixing forbidden characters for onedrive imports

pablomd314:
fixes #818

PR #823, opened 2020-01-15
Transmogrification config fix

pablomd314:
fixes class inheritance issues for config from #820


seehamrun:
Feel free to create an issue to address this later. But a better model would probably be to make each of the fields a json property and supply them via a configuration file instead. 

That will prevent issues in the future if we need to expand this config or add other options

PR #827, opened 2020-01-23
[MicrosoftImporter]: also add body to log when error occurs uploading a single photo...

seehamrun:
I missed this in my last PR


seehamrun:
Travis is still broken due to chrome stable - so going to merge without it

PR #837, opened 2020-01-30
Handle null values in album name when cleaning name up

pablomd314:


portability-types-common/src/main/java/org/datatransferproject/types/common/models/photos/PhotoAlbum.java
```diff
@@ -86,6 +86,9 @@ public int hashCode() {
   }
 
   public void cleanName(String forbiddenCharacters, char replacementCharacter, int maxLength) {
+    if (name == null) {
+      return;
+    }
```
seehamrun: Lets add a comment that these should be handled in the executor and in each import service (maybe a todo to make it a part of the transmorgifier?)

PR #838, opened 2020-02-03
Make MicrosoftPhotosImporter use upload sessions

pablomd314:
According to this, for files larger than 4MB, it is required that we use createUploadSession of the OneDrive API.

https://docs.microsoft.com/en-us/onedrive/developer/rest-api/api/driveitem_createuploadsession?view=odsp-graph-online#create-an-upload-session

This code completes that requirement, and now we can upload files that are larger than 4MB to OneDrive 😲

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/datatransferproject/transfer/microsoft/photos/MicrosoftPhotosImporter.java
+++ b/extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/datatransferproject/transfer/microsoft/photos/MicrosoftPhotosImporter.java
@@ -245,7 +245,7 @@ public class MicrosoftPhotosImporter implements Importer<TokensAndUrlAuthData, P
         "Got error code: " + code + " message: " + response.message() + " body: " + response
         .body().string());
     } else if (code != 200) {
-      monitor.info(() -> String.format("Got an unexpected non-200, non-error response code: %s, %s", code, responseBody.string()));
+      monitor.info(() -> String.format("Got an unexpected non-200, non-error response code"));
     }
     Preconditions.checkState(responseBody != null, "Got Null Body when creating photo upload session %s", photo);
     Map<String, Object> responseData = objectMapper.readValue(responseBody.bytes(), Map.class);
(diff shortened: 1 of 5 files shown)


extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/datatransferproject/transfer/microsoft/photos/MicrosoftPhotosImporter.java
```diff
@@ -142,143 +149,233 @@ private String createOneDriveFolder(PhotoAlbum album) throws IOException {
...
+      if (chunkCode == 200) {
+        monitor.info(() -> String.format("Uploaded chunk %s-%s successfuly", chunk.getStart(), chunk.getEnd()));
       }
     }
+    // get complete file response
+    Preconditions.checkState(chunkCode == 201 || chunkCode == 200, "Got bad response code when finishing uploadSession: %d", chunkCode);
+    //
+    ResponseBody chunkResponseBody = chunkResponse.body();
+    Map<String, Object> chunkResponseData = objectMapper.readValue(chunkResponseBody.bytes(), Map.class);
+    return (String) chunkResponseData.get("id");
```
cnnorris: For readability purposes can you generalize the concrete blocks of this function so that it's not 100 lines long?

e.g. `List<Chunks> = getpages(photo)`, ` for chunk in chunks: upload_file(chunk)`, that sorta stuff.

extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/datatransferproject/transfer/microsoft/photos/MicrosoftPhotosImporter.java
```diff
@@ -142,143 +149,233 @@ private String createOneDriveFolder(PhotoAlbum album) throws IOException {
...
-    @SuppressWarnings("NullableProblems")
-    public void writeTo(BufferedSink sink) throws IOException {
-      Source source = null;
-      try {
-        source = Okio.source(stream);
-        sink.writeAll(source);
-      } finally {
-        Util.closeQuietly(source);
-      }
+  private static class Chunk {
```
cnnorris: This should probably be defined in a common class.

extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/datatransferproject/transfer/microsoft/photos/MicrosoftPhotosImporter.java
```diff
@@ -142,143 +149,233 @@ private String createOneDriveFolder(PhotoAlbum album) throws IOException {
...
-        } catch (IOException e1) {
-          monitor.info(() -> "Couldn't close input stream");
-        }
+    }
+    if (offset != 0) {
+      chunksToSend.add(new Chunk(data, offset, chunksRead * CHUNK_SIZE));
+      totalFileSize += offset;
+      chunksRead++;
+    }
+    final int _debugTotalFileSize = totalFileSize;
```
cnnorris: do you need these? it seems like you only use them once without modifying them before that

extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/datatransferproject/transfer/microsoft/photos/MicrosoftPhotosImporter.java
```diff
@@ -176,86 +179,51 @@ private String createOneDriveFolder(PhotoAlbum album) throws IOException {
...
-        String photoId = (String) responseData.get("id");
-        checkState(
-            !Strings.isNullOrEmpty(photoId), "Expected id value to be present in %s", responseData);
-        return photoId;
+    Integer chunkCode = null;
+    Response chunkResponse = null;
+    for (DataChunk chunk : chunksToSend) {
+      chunkResponse = uploadChunk(chunk, photoUploadUrl, totalFileSize, photo.getMediaType());
+      chunkCode = chunkResponse.code();
+      if (chunkCode < 200 || chunkCode > 299) {
```
cnnorris: I might move this exception detection stuff into the uploadChunk method, around where you're doing the 401 catch.

PR #839, opened 2020-02-04
Count bytes transferred in photos and videos transfers

wmorland:
This includes a google-java-format-1.7 commit so to just see the meaningful changes you can isolate to https://github.com/google/data-transfer-project/commit/56ef12f459ccfc8acfac5186655b82905de531e5


wmorland:
Travis CI failure seems to be a timeout on fetching a dependency.

PR #843, opened 2020-02-21
Create an abstract CopyExceptionWithFailureReason

wmorland:
So exceptions can store the failure reason on the exception and we have no MemoryFull specific logic.

### DIFF
--- a/portability-spi-transfer/src/main/java/org/datatransferproject/spi/transfer/types/CopyExceptionWithFailureReason.java
+++ b/portability-spi-transfer/src/main/java/org/datatransferproject/spi/transfer/types/CopyExceptionWithFailureReason.java
@@ -1,0 +1,13 @@
+package org.datatransferproject.spi.transfer.types;
+
+import javax.annotation.Nonnull;
+
+public abstract class CopyExceptionWithFailureReason extends CopyException {
+
+  public CopyExceptionWithFailureReason(String message, Throwable cause) {
+    super(message, cause);
+  }
+
+  @Nonnull
+  public abstract String getFailureReason();
+}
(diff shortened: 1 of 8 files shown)


portability-spi-transfer/src/main/java/org/datatransferproject/spi/transfer/types/CopyExceptionWithFailureReason.java
```diff
@@ -0,0 +1,13 @@
+package org.datatransferproject.spi.transfer.types;
+
+import javax.annotation.Nonnull;
+
+public abstract class CopyExceptionWithFailureReason extends CopyException {
```
holachuy: Would you be able to add some javadoc here

portability-spi-transfer/src/main/java/org/datatransferproject/spi/transfer/types/DestinationMemoryFullException.java
```diff
@@ -13,11 +13,19 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.datatransferproject.transfer;
-import org.datatransferproject.transfer.CopyException;
+package org.datatransferproject.spi.transfer.types;
+
+import javax.annotation.Nonnull;
+
+public class DestinationMemoryFullException extends CopyExceptionWithFailureReason {
```
holachuy: java doc here as well

portability-spi-transfer/src/main/java/org/datatransferproject/spi/transfer/types/DestinationMemoryFullException.java
```diff
@@ -13,11 +13,19 @@
...
 
-public class DestinationMemoryFullException extends CopyException {
   public DestinationMemoryFullException(String message, Throwable cause) {
     super(message, cause);
   }
+
+  @Nonnull
+  @Override
+  public String getFailureReason() {
+    return "DESTINATION_FULL";
```
holachuy: Should we put this in a constants file somewhere?

PR #846, opened 2020-03-05
fix NPE in MicrosoftPhotosImporter

pablomd314:


extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/datatransferproject/transfer/microsoft/photos/MicrosoftPhotosImporter.java
```diff
@@ -209,8 +209,11 @@ private String importSinglePhoto(
     Response chunkResponse = null;
     for (DataChunk chunk : chunksToSend) {
       chunkResponse = uploadChunk(chunk, photoUploadUrl, totalFileSize, photo.getMediaType());
+      chunkCode = chunkResponse.code();
     }
     // get complete file response
+    Preconditions.checkNotNull(chunkResponse, "chunkResponse is null");
+    Preconditions.checkNotNull(chunkCode, "chunkCode is null");
```
seehamrun: Can we move these checks to inside the for loop - if an individual chunk fails, thats a bad state right?

PR #848, opened 2020-03-09
Fixed null ptr caused by missing string formatting argument

pablomd314:


extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/datatransferproject/transfer/microsoft/photos/MicrosoftPhotosImporter.java
```diff
@@ -210,6 +210,11 @@ private String importSinglePhoto(
     for (DataChunk chunk : chunksToSend) {
       chunkResponse = uploadChunk(chunk, photoUploadUrl, totalFileSize, photo.getMediaType());
     }
+    if (chunkResponse.code() != 200 && chunkResponse.code() != 201) { 
+      // Once we upload the last chunk, we should have either 200 or 201. 
+      // This should change to a precondition check after we debug some more.
+      monitor.debug(() -> "Received a bad code on completion of uploading chunks", chunkResponse.code());
```
seehamrun: If this happens, should we continue the function call or return? My initial thought is to just make this a precondition off the bat, whats stopping us from doing that?

pablomd314: We don't really know under what conditions we won't get 200 or 201, it's possible that a healthy upload could result in a code other than 200 or 201. I want to see if we ever get a code other than 200 or 201 before committing to a precondition.

PR #850, opened 2020-03-10
cleanup unused vars and functions in smugmug extension

seehamrun:


seehamrun:
Not sure if theres a way to retrigger travis, but this looks like a transient failure

PR #861, opened 2020-04-01
Fix flickr gif test

seehamrun:


seehamrun:
Travis failure looks like a transient failure in the AngularJS module.

PR #871, opened 2020-05-04
close input streams

pablomd314:


extensions/data-transfer/portability-data-transfer-flickr/src/main/java/org/datatransferproject/datatransfer/flickr/photos/FlickrPhotosImporter.java
```diff
@@ -230,6 +229,7 @@ private String uploadPhoto(PhotoModel photo, UUID jobId) throws IOException, Fli
             .setDescription(photoDescription);
     perUserRateLimiter.acquire();
     String uploadResult = uploader.upload(inStream, uploadMetaData);
+    inStream.close();
```
holachuy, line 232: Is uploader.upload a blocking call that ensures the stream is no longer used? 

We could also use a try with resources block if the call is AutoCloseable.

pablomd314: it can be async or sync, but we call setAsync(false) on the metadata, so this is a synchronous call that only returns on success.

PR #873, opened 2020-05-04
no microsoft empty album names

pablomd314:


extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/datatransferproject/transfer/microsoft/photos/MicrosoftPhotosImporter.java
```diff
@@ -140,7 +140,9 @@ public ImportResult importItem(
   private String createOneDriveFolder(PhotoAlbum album) throws IOException {
 
     Map<String, Object> rawFolder = new LinkedHashMap<>();
-    rawFolder.put("name", album.getName());
+    // clean up album name for microsoft specifically
+    String albumName = album.getName() == null || album.getName().length() == 0 ? "Untitled" : album.getName();
```
seehamrun, line 144: Length for the single space string will be 1 right? Can we trim and use Strings.isNullOrEmpty?

pablomd314: no, it gets trimmed in PhotoAlbum to remove trailing and leading spaces. changed to isnullorempty

extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/datatransferproject/transfer/microsoft/photos/MicrosoftPhotosImporter.java
```diff
@@ -140,7 +140,9 @@ public ImportResult importItem(
   private String createOneDriveFolder(PhotoAlbum album) throws IOException {
 
     Map<String, Object> rawFolder = new LinkedHashMap<>();
-    rawFolder.put("name", album.getName());
+    // clean up album name for microsoft specifically
+    String albumName = Strings.isNullOrEmpty(album.getName()) ? "Untitled" : album.getName();
```
seehamrun, line 144: I think you need to trim it first to get rid of the whitespace. this would return false for a string with a single space.

pablomd314: no, it gets trimmed in PhotoAlbum to remove trailing and leading spaces.

seehamrun: Ah forgot about that, lets add it in the above comment.

PR #874, opened 2020-05-05
Add new exception to capture user permission denial

zhang123cnn:
Summary:
Currently, when user import photos to google photos, the user needs to
select the google account with access to photos.

Otherwise, Google API would reply with a 403 error which cause a Java
IOException.

This commit is to capture this 403 error in a new exception class based
on CopyExceptionWithFailureReason. So it would allow displaying better
error message to the front end users.

Test Plan:
Import photos to a google account with no access to photos. Worker
should throw a PermissionDeniedException rather than an IOException

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/videos/GoogleVideosImporter.java
+++ b/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/videos/GoogleVideosImporter.java
@@ -61,6 +61,7 @@ import org.datatransferproject.spi.transfer.idempotentexecutor.IdempotentImportE
 import org.datatransferproject.spi.transfer.provider.ImportResult;
 import org.datatransferproject.spi.transfer.provider.Importer;
 import org.datatransferproject.spi.transfer.types.DestinationMemoryFullException;
+import org.datatransferproject.spi.transfer.types.UploadErrorException;
 import org.datatransferproject.transfer.ImageStreamProvider;
 import org.datatransferproject.types.common.models.videos.VideoObject;
 import org.datatransferproject.types.common.models.videos.VideosContainerResource;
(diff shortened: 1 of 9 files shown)


extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/videos/GoogleVideosImporter.java
```diff
@@ -165,6 +166,14 @@ VideoResult importSingleVideo(VideoObject inputVideo, PhotosLibrarySettings sett
...
       if (uploadResponse.getError().isPresent() || !uploadResponse.getUploadToken().isPresent()) {
         Error error = uploadResponse.getError().orElse(null);
+        if (error != null
+            && error
+                .getCause()
+                .getMessage()
+                .contains("The upload url is either finalized or rejected by the server")) {
+          throw new UploadErrorException(
+              "Upload was terminated because of error", error.getCause());
+        }
```
seehamrun, line 176: Q: If this happens once does it mean it will happen for all other videos as well? I'm wondering why we call out this error rather than all the other ones? Wouldn't it be better to mark this error as Fatal in the retry strategy and continue with the rest of the import?

zhang123cnn: We will mark this error as Fatal in the retry strategy, because you are right if this happens once it would fail all other videos too. 
We use a separate exception derived from CopyExceptionWithFailureReason because within, we would be able to capture this specific error and show a better error message to our user on UI.

seehamrun: Sorry for the misunderstanding: I meant if this failure is expected to fail all the other video uploads too, then I think this is fine, because the whole job is expected to fail on this error. 

But if it only fails for a specific video (and other videos can still be uploaded), I think it makes sense to just have special handling for it in the retry strategy, that way it won't get retried and other iterations of the copy can still take place. 

Let me know if that makes sense, or if Im misunderstanding something, I am not as familiar with the error codes returned from the video upload sessions

wmorland: Yes, this is related to https://github.com/google/java-photoslibrary/issues/31 There is more in the HttpResponse but we just get that the Upload Status was fatal.

seehamrun: Ok - so if all other videos are also expected to fail, this makes sense as is since the whole job fails! I just wasn't sure if this error was specific to the video that was being uploaded. 

Thanks for the explanation!

PR #879, opened 2020-05-08
Hash and equals

pablomd314:
if we redefine equals we should redefine hash to prevent issues when dealing with things like Sets and other hash using hashers


seehamrun:
For posterity, AutoValue implements hash and equals properly, which is why we're removing the equals() override in PortabilityJob

PR #880, opened 2020-05-11
Insecure crypto

pablomd314:
fixes 
1) non-static encryption methods
2) ambiguous AES encryption method


libraries/security/src/main/java/org/datatransferproject/security/CryptoTransformations.java
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.datatransferproject.security;
+
+/** Provides compile-time constants for secure crypto transformations. For use 
+in creation of Encrypter/Decrypter */
+public final class CryptoTransformations {
```
holachuy, line 20: How about we use an enum to be typsafe then we can use it in the case conditions?

public enum CryptoTransformation {
    AES_CBC_NOPADDING("AES/CBC/NoPadding"),
    RSA_ECB_PKCS1("RSA/ECB/PKCS1Padding"),

    public final String algorithm;
 
    private CryptoTransformation(String algorithm) {
        this.algorithm = algorithm;
    }
}

libraries/security/src/main/java/org/datatransferproject/security/EncrypterImpl.java
```diff
@@ -47,21 +49,47 @@
...
   public String encrypt(String data) {
     try {
-      Cipher cipher = Cipher.getInstance(transformation);
-      cipher.init(Cipher.ENCRYPT_MODE, key);
-      byte[] salt = new byte[8];
+      Cipher cipher;
+      switch (transformation) {
+      case CryptoTransformations.AES_CBC_NOPADDING:
+        cipher = Cipher.getInstance(CryptoTransformations.AES_CBC_NOPADDING);
+        cipher.init(Cipher.ENCRYPT_MODE, key, generateIv(cipher));
```
holachuy, line 56: Per offline converation, let's add high level outline to javadoc.

PR #881, opened 2020-05-12
Add finishedJob metric when job is cancelled

zhang123cnn:
Summary:
Currently when a job is started and then cancelled by user, dtp worker
would only print the startedJob metrics line in the log. This makes the
log line incomplete and confuses engineers with unfinished job.

The commit here is to add the finishedJob metrics in job cancellation.

Test Plan: Cancel a job and see the new log line appearing there.


zhang123cnn:
Could anyone take a look at this PR? Thanks :)

PR #882, opened 2020-05-13
Fix token refresh issue for google photo importer

zhang123cnn:
Summary:
Currently google photo importer create a new google photo interface
every time it sends a http request.

If the original access token is expired, the new access token acquired
from refresh token would not be persisted across http request.

This commit is to fix the inefficiency here that we don't create a new
google photo interface every time we send a new request.

Test Plan:
Send request after access token expires then the refresh action should
only happen once for the next few requests.

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosImporter.java
+++ b/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosImporter.java
@@ -228,7 +228,10 @@ public class GooglePhotosImporter
 
   private synchronized GooglePhotosInterface getOrCreatePhotosInterface(
       TokensAndUrlAuthData authData) {
-    return photosInterface == null ? makePhotosInterface(authData) : photosInterface;
+    if (photosInterface == null) {
+      photosInterface = makePhotosInterface(authData);
+    }
+    return photosInterface;
   }
 
   private synchronized GooglePhotosInterface makePhotosInterface(TokensAndUrlAuthData authData) {
--- a/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosImporter.java
+++ b/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosImporter.java
@@ -23,6 +23,8 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.HttpURLConnection;
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.UUID;
 import org.datatransferproject.api.launcher.Monitor;
 import org.datatransferproject.datatransfer.google.common.GoogleCredentialFactory;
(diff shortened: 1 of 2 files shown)


extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosImporter.java
```diff
@@ -228,7 +228,10 @@ private String getPhotoDescription(PhotoModel inputPhoto) {
 
   private synchronized GooglePhotosInterface getOrCreatePhotosInterface(
       TokensAndUrlAuthData authData) {
-    return photosInterface == null ? makePhotosInterface(authData) : photosInterface;
+    if (photosInterface == null) {
+      photosInterface = makePhotosInterface(authData);
+    }
+    return photosInterface;
```
seehamrun, line 234: So I think we only have the existing setup for testing (so that an interface doesnt get created for test purposes).

The reason we didn't do this originally is to future proof ourselves when the time came to make the worker multithreaded. Since the import/export function takes in a set of AuthData, it is expected that this can be used more than once and different auth data should be using a different client interface. 

For now, instead of storing this as a member variable, I think it is better to make it a map of jobid --> photos interfaces instead of having the single member variable.

zhang123cnn: Thanks for your comment @seehamrun 

IMHO, we can just run more workers to scale up the processing units. What would be the advantage for adopting a multithreading approach here?

I can understand if we want to do multithreading to shorten the processing time of one job. But in that case, we also don't need to have multiple photo interface per job because the auth data for one job would be the same. 

Happy to discuss

seehamrun: I agree multithreading for one job would be OK because the auth data is the same for that job. 

The concern here is for future work if we wanted to have a worker perform and process more than one job at a time. Right now we are scaling up by running more workers, but that has its downsides and at some point we will exhaust resources and not be able to scale that way. By making the export/import functionality resilient to that it allows us to go down that path. 

A safer but bigger change would be to make each importer/exporter specific to a job and provide the authentication data in the constructor. However this would require a larger design change in the startup logic and its been on our backlog for a while. With the design right now, we are assuming that these exporters/importers can be reused with different authdatas between each call.

I think the alternative of maintaining a map here is a good middle ground and can be refactored later when we do the above and prevents a potential multithreading bug that would occur if overlooked

zhang123cnn: > but that has its downsides and at some point we will exhaust resources and not be able to scale that way

Could you elaborate what resources we are worried to exhaust here? By having more workers, I can scale up horizontally by just adding more machines. I am curious about what potential downsides I am missing here.

seehamrun: Discussed offline, since the importer/exporter interface here expects that auth data will be passed in per request with potential to run more jobs in parallel, we should maintain a map instead.

In the future, we should make the importer/exporter interface job based and initialize the auth data at construction time so instances can not be reused across requests.

PR #892, opened 2020-06-05
Fix token issue when importing video to google

zhang123cnn:
Summary:
Currently once google auth token expires after 1 hour, all video imports
would be rejected by Google due to invalid token.

This commit is to ask the video importer always refresh token before it
sends the first http request. So the right expiry time would be filled in for the job. 

Test Plan: Test by using expired token to transfer video and the refresh logic would kick in.


extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/videos/GoogleVideosImporter.java
```diff
@@ -52,9 +52,9 @@
 import java.io.InputStream;
 import java.io.RandomAccessFile;
 import java.net.HttpURLConnection;
-import java.util.Collections;
-import java.util.List;
-import java.util.UUID;
+import java.util.*;
```
seehamrun, line 55: Nit: I think its standard style to avoid * imports

PR #894, opened 2020-06-09
Add per service per data type logging to RetryingCallable

zhang123cnn:
Summary:
We would like to better keep track of retrying activities across
different data types and services.

Adding extra logging to record service and data type info when retrying.

Test Plan: Do a transfer which requires retry and check the log line
example log line -
```
DEBUG 2020-06-09T15:27:18.422 Attempt 1 failed, using retry strategy: ExponentialBackoffStrategy{maxAttempts=5, initialIntervalMillis=1000, multiplier=1.5}, service: Google, dataType: PHOTOS
DEBUG 2020-06-09T15:27:18.424 Strategy has 999 remainingIntervalMillis after 1 elapsedMillis
DEBUG 2020-06-09T15:27:19.425 Make 2 attempt of retry to service Google on data type PHOTOS
```

### DIFF
--- a/portability-types-transfer/src/main/java/org/datatransferproject/types/transfer/retry/RetryingCallable.java
+++ b/portability-types-transfer/src/main/java/org/datatransferproject/types/transfer/retry/RetryingCallable.java
@@ -35,6 +35,8 @@ public class RetryingCallable<T> implements Callable<T> {
   private final RetryStrategyLibrary retryStrategyLibrary;
   private final Clock clock;
   private final Monitor monitor;
+  private final String dataType;
+  private final String service;
 
   private volatile int attempts;
   private volatile Exception mostRecentException;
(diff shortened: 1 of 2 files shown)


portability-types-transfer/src/main/java/org/datatransferproject/types/transfer/retry/RetryingCallable.java
```diff
@@ -85,6 +92,11 @@ public T call() throws RetryException {
           if (nextAttemptIntervalMillis > 0L) {
             try {
               Thread.sleep(nextAttemptIntervalMillis);
+              monitor.debug(
+                  () ->
```
seehamrun, line 96: Does it make sense to move this log line up to the beginning of the try block? (where the attempt actually happens)? Also what do you think of this format? "Attempt %d started, service:%s, dataType: %s" to be consistent with the log above.

zhang123cnn: If we move this log line to the beginning, then we would make the logging even for the first try. And first try is not a retry, so that's why I have put it here.
Having said that, no strong feeling either way. I can make that change if you think that is better.

seehamrun: I think thats fine, the log would just look like "Attempt 1 started..." - just a thought since it would avoid using that `attempt + 1` in the format string

PR #900, opened 2020-07-01
Fix wrong importer retry due to getErrors on executor

zhang123cnn:
Summary:
Currently in CallableImporter, the code would call getErrors on
idempotentExecutor to determine if there was any error when importing.

However, this getErrors would contain ALL errors happened in the
previous import actions, not errors only happened for the current
import. This would mean one import error in the early stage would cause
retry for all subsequent import actions.

This commit is to add new methods to track only recent errors on
executor and only throw out exceptions when those recent errors are
created.

Test Plan:
Raise one IOException when importing and check if it would impact the
subseqent import retries.

### DIFF
--- a/portability-transfer/src/main/java/org/datatransferproject/transfer/CallableImporter.java
+++ b/portability-transfer/src/main/java/org/datatransferproject/transfer/CallableImporter.java
@@ -63,9 +63,11 @@ public class CallableImporter implements Callable<ImportResult> {
     boolean success = false;
     Stopwatch stopwatch = Stopwatch.createStarted();
     try {
+      idempotentImportExecutor.resetRecentErrors();
       ImportResult result = importerProvider.get()
           .importItem(jobId, idempotentImportExecutor, authData, data);
-      Collection<ErrorDetail> errors = idempotentImportExecutor.getErrors();
+
+      Collection<ErrorDetail> errors = idempotentImportExecutor.getRecentErrors();
       success = result.getType() == ImportResult.ResultType.OK && errors.isEmpty();
 
       if (!success) {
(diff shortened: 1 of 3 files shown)


portability-transfer/src/main/java/org/datatransferproject/transfer/CallableImporter.java
```diff
@@ -63,9 +63,11 @@ public ImportResult call() throws Exception {
     boolean success = false;
     Stopwatch stopwatch = Stopwatch.createStarted();
     try {
+      idempotentImportExecutor.resetRecentErrors();
       ImportResult result = importerProvider.get()
           .importItem(jobId, idempotentImportExecutor, authData, data);
-      Collection<ErrorDetail> errors = idempotentImportExecutor.getErrors();
+
+      Collection<ErrorDetail> errors = idempotentImportExecutor.getRecentErrors();
```
seehamrun, line 70: Just to clarify here, the recent errors will be per "import page" correct? What do you think of renaming the interface method to be explicit about that? ex: getLastImportPageErrors or getImportAttemptErrors?

I wonder if a better design paradigm would be to return the errors in the ImportResult instead of relying on resetting the executor? That way the executor maintains its own overall state of errors and the result would contain the page errors. Not sure if thats feasible with the current implementation, but maybe we should file a github issue to pick it up and investigate it later?

zhang123cnn: > Just to clarify here, the recent errors will be per "import page" correct?
Yes, but I think theoretically getRecentErrors is not related with page or anything. It will contain errors until .resetRecentErrors is called. In that sense, I am reluctant to rename it to getLastImportPageErrors or getImportAttemptErrors because it may give the wrong impression that we do not need to call resetRecentErrors to clear it up.

> I wonder if a better design paradigm would be to return the errors in the ImportResult instead of relying on resetting the executor?
Yes, this will be another good design. And I did start with that to fix the issue. But unfortunately we have many importers and they don't share the same understanding of what a OK import result really means. And changing all of them to be consistent is too much effort as of now. So that's why I decided to try out this approach. Raising a github issue sounds like a good idea and will do that shortly.

seehamrun: That makes sense - thanks for the explanation

PR #903, opened 2020-07-10
Handle NullPointerException when unwrapping Google Photos response

mukul-rathi:
Summary:
Previously we got a null pointer exception when importing a single photo into Google Photos.  This commit adds null checks and instead throws an IOException, which can be used by our workers to retry a photo import.

Test Plan: Ran a test transfer in Docker. Add a test to confirm that an IOException is raised, not a NullPointerException.

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosImporter.java
+++ b/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosImporter.java
@@ -28,8 +28,11 @@ import java.util.Map;
 import java.util.UUID;
 import org.datatransferproject.api.launcher.Monitor;
 import org.datatransferproject.datatransfer.google.common.GoogleCredentialFactory;
+import org.datatransferproject.datatransfer.google.mediaModels.BatchMediaItemResponse;
 import org.datatransferproject.datatransfer.google.mediaModels.GoogleAlbum;
+import org.datatransferproject.datatransfer.google.mediaModels.GoogleMediaItem;
 import org.datatransferproject.datatransfer.google.mediaModels.NewMediaItem;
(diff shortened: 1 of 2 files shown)


extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosImporter.java
```diff
@@ -202,13 +205,20 @@ PhotoResult importSinglePhoto(
...
+      }
+      NewMediaItemResult[] mediaItemResults = photoCreationResponse.getResults();
+      if (mediaItemResults == null || mediaItemResults.length != 1) {
+        throw new IOException("No media item results returned");
+      }
+      GoogleMediaItem mediaItem = mediaItemResults[0].getMediaItem();
+      if (mediaItem == null) {
+        throw new IOException("No media item returned");
+      }
+      return new PhotoResult(mediaItem.getId(), bytes);
```
seehamrun, line 221: Would it be better here to use a preconditon checkNotNull?

wmorland: @seehamrun This touches on a question I have about the behaviour of the idempotentImportExecutor. The executor only swallows IOExceptions, was there a particular reason for this choice? My instinct is that this sort of issue with the API response is the sort of thing we would want the executor to handle but using precondition checkNotNull would not handle this.

seehamrun: Sorry for the late response, that's a good point. I dont think we had any specific reason for this. We wanted to distinguish between framwork errors and api errors somehow and this seemed like the easiest way to do so. 

However, we can revisit this as well. Do you want to file a github issue to follow up on it?

PR #910, opened 2020-07-20
Add photos batching

wmorland:
Summary: This uses the same technique we used for videos to batch photos

Test Plan: Ran a transfer on my machine using the docker container


extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosImporter.java
```diff
@@ -123,15 +130,36 @@ public ImportResult importItem(
...
+          googleAlbumId = null;
+        } else {
+          // Note this will throw if creating the album failed, which is what we want
+          // because that will also mark this photo as being failed.
+          googleAlbumId = idempotentImportExecutor.getCachedValue(originalAlbumId);
+        }
+
+        // We partition into groups of 49 as 50 is the maximum number of items that can be created
+        // in
+        // one call. (We use 49 to avoid potential off by one errors)
```
seehamrun, line 154: nit: looks like this comment got split up unusually

PR #912, opened 2020-07-22
Read ExportInformation as a string and deserialize it in the JobProcessor

seehamrun:
Previously we were expecting that the builder be able to pull out the the ExportInformation from the map, however the cloud extension thats used writes this out as a json string

Interested in thoughts here, we already deserialize objects in the Processor (ex: the AuthData pair). The other thing we can do is deserialize in the Cloud extension, but we dont have a good convention for what is and isnt deserialized (or decrypted) there.

### DIFF
--- a/portability-api/src/main/java/org/datatransferproject/api/action/transfer/CreateTransferJobAction.java
+++ b/portability-api/src/main/java/org/datatransferproject/api/action/transfer/CreateTransferJobAction.java
@@ -1,11 +1,20 @@
 package org.datatransferproject.api.action.transfer;
 
+import static java.lang.String.format;
+import static org.datatransferproject.api.action.ActionUtils.encodeJobId;
+import static org.datatransferproject.spi.api.auth.AuthServiceProviderRegistry.AuthMode.EXPORT;
+import static org.datatransferproject.spi.api.auth.AuthServiceProviderRegistry.AuthMode.IMPORT;
+
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import com.google.common.io.BaseEncoding;
 import com.google.inject.Inject;
+import java.io.IOException;
+import java.util.Optional;
+import java.util.UUID;
+import javax.crypto.SecretKey;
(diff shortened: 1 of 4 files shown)


seehamrun:
Im going to go with a different approach for this

seehamrun:
OK so I was going to try and write out the ExportInformation as a serialized object in our datastore, we have support for deserializing byte strings here: https://github.com/google/data-transfer-project/blob/master/extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleJobStore.java#L105-L111

However, it turns out using this approach doesnt exactly work for ExportInformation because the subtypes are not serializable. We are trying to write out an ExportInformation with a PhotosContainerResource (which contains a PhotoAlbum). The subclass PhotoAlbum is not serializable: https://github.com/google/data-transfer-project/blob/d03e62c864b306e30f3ed4d3d0be39ead34a0906/portability-types-common/src/main/java/org/datatransferproject/types/common/models/photos/PhotoAlbum.java

seehamrun:
Ok it looks like this is the case for all the other underlying container resources as well. Im going to reopen this PR once I fix the build issues and we can address the consistency fixes needed to make the underlying types serializable later on (and we can then revert this PR)

portability-api/src/main/java/org/datatransferproject/api/action/transfer/CreateTransferJobAction.java
```diff
@@ -79,9 +80,14 @@ public TransferJob handle(CreateTransferJob request) {
...
-    PortabilityJob job =
-        createJob(encodedSessionKey, dataType, exportService, importService, exportInformation, encryptionScheme);
-
+    PortabilityJob job;
+    try {
+      job =
+          createJob(encodedSessionKey, dataType, exportService, importService, exportInformation,
+              encryptionScheme);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
```
holachuy, line 89: Should some additional information be logged, e.g. the request?

seehamrun: The request could contain PII so I dont want to unnecessarily log it. And in this case the exception should be enough (it will contain the reason the deserialization failed).

PR #914, opened 2020-07-22
Enable exporting of specific resources in the GooglePhotosExporter and FlickrPhotosExporter

seehamrun:
This is a quick first pass at enabling exporting specific resources in these exporters. 

Previously they only accepted IdOnlyContainerResources in the export information. This assumes that if the ExportInformation contains a PhotosContainerResource that the export process should start there instead of listing out all the user media items

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-flickr/src/main/java/org/datatransferproject/datatransfer/flickr/photos/FlickrPhotosExporter.java
+++ b/extensions/data-transfer/portability-data-transfer-flickr/src/main/java/org/datatransferproject/datatransfer/flickr/photos/FlickrPhotosExporter.java
@@ -123,6 +123,14 @@ public class FlickrPhotosExporter implements Exporter<AuthData, PhotosContainerR
 
     RequestContext.getRequestContext().setAuth(auth);
 
+    if (exportInformation.isPresent()
+        && exportInformation.get().getContainerResource() instanceof PhotosContainerResource) {
+      // if ExportInformation is a photos container, this is a request to only export the contents
+      // in that container instead of the whole user library
+      return exportPhotosContainer(
+          (PhotosContainerResource) exportInformation.get().getContainerResource());
+    }
+
     PaginationData paginationData =
         exportInformation.isPresent() ? exportInformation.get().getPaginationData() : null;
     IdOnlyContainerResource resource =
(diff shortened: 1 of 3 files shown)


extensions/data-transfer/portability-data-transfer-flickr/src/main/java/org/datatransferproject/datatransfer/flickr/photos/FlickrPhotosExporter.java
```diff
@@ -123,6 +123,14 @@ static String toMimeType(String flickrFormat) {
 
     RequestContext.getRequestContext().setAuth(auth);
 
+    if (exportInformation.isPresent()
```
holachuy, line 126: Let's add a comment here to describe this logic

extensions/data-transfer/portability-data-transfer-flickr/src/main/java/org/datatransferproject/datatransfer/flickr/photos/FlickrPhotosExporter.java
```diff
@@ -136,6 +144,37 @@ static String toMimeType(String flickrFormat) {
...
+        // Adding subresources tells the framework to recall export to get all the photos
+        subResources.add(new IdOnlyContainerResource(photoset.getId()));
+      }
+
+      for (PhotoModel photo : container.getPhotos()) {
+        Photo p = photosInterface.getInfo(photo.getDataId(), null);
+        photosBuilder.add(toCommonPhoto(p, null));
+      }
+    } catch (FlickrException e) {
+      return new ExportResult<>(e);
```
holachuy, line 168: Any additional logging we want to try here?

extensions/data-transfer/portability-data-transfer-flickr/src/main/java/org/datatransferproject/datatransfer/flickr/photos/FlickrPhotosExporter.java
```diff
@@ -136,6 +144,37 @@ static String toMimeType(String flickrFormat) {
...
+    }
+
+    PhotosContainerResource photosContainerResource =
+        new PhotosContainerResource(albumBuilder.build(), photosBuilder.build());
+    ContinuationData continuationData = new ContinuationData(null);
+    subResources.forEach(resource -> continuationData.addContainerResource(resource));
+    return new ExportResult<>(ResultType.CONTINUE, photosContainerResource, continuationData);
```
holachuy, line 175: What would happen if there are no albums and no photos and each of these is empty?

extensions/data-transfer/portability-data-transfer-flickr/src/main/java/org/datatransferproject/datatransfer/flickr/photos/FlickrPhotosExporter.java
seehamrun: I added a comment, but it was below the if statmenet, moved it to the top

extensions/data-transfer/portability-data-transfer-flickr/src/main/java/org/datatransferproject/datatransfer/flickr/photos/FlickrPhotosExporter.java
seehamrun: I followed the same thing we were doing with the individual getAlbums() and getPhotos() calls. The exception would be logged up a level (in the executor)

extensions/data-transfer/portability-data-transfer-flickr/src/main/java/org/datatransferproject/datatransfer/flickr/photos/FlickrPhotosExporter.java
seehamrun: Then there would be nothing to import. The job will be counted as "success". 

I guess what we do here depends on what the expectation is if there is no content to export. If they specified an invalid photo/album id the lookup fails and an error gets thrown. 

If we want the job to fail, we can add a precondition check, but again not sure if thats what we want to do.

seehamrun:
@wmorland Added you in case you were creating the ExportInformation differently.

Also note: I chose to look up the album based on the ID (though theoratically, we could take in the id, title, description and other metadata from the request) just in case between selection and the job processing the album details change or get deleted (or if theres a misconfiguration and the album doenst exist)

Issue #916, opened 2020-07-27
Move remaining models out of portability-types-transfer and into portability-types-common


seehamrun:
Theres documentation and content about the Blobby model in portability-types-transfer whereas the rest of the models live in portability-types-common. 

https://github.com/google/data-transfer-project/tree/abb554b604329dcacaa62a8b8c5615f6aa47e0bd/portability-types-transfer/src/main/java/org/datatransferproject/types/transfer/models

https://github.com/google/data-transfer-project/tree/abb554b604329dcacaa62a8b8c5615f6aa47e0bd/portability-types-common/src/main/java/org/datatransferproject/types/common/models

jzacsh:
reopening (looks like the documentation was missed - see #1056 - but there might be more string matches like this too, so I'm just reopening to do another check)

lisad:
Hopefully JZacsh did the needed work to address the documentation that was missed in the various commits that reference this.  In any case, closing for now because keeping the issue open but static is not very useful.

PR #919, opened 2020-07-31
Update subtypes of ContainerResource

mukul-rathi:
I noticed that VideoContainerResource was not a subtype of ContainerResource.

This diff corrects the ContainerResource file to add the other subtypes of Container resource, the only exception being BlobbyDataContainerResource - this is because it resides in another module and doing so would introduce a cyclic build dependency. (https://github.com/google/data-transfer-project/issues/916 already flags this).


mukul-rathi:
@seehamrun If you're happy with this PR, could you merge this and issue a new release v0.3.45 to bintray? We're currently blocked internally at Facebook by this minor fix, so would appreciate it, thanks!

PR #932, opened 2020-09-08
Update social posts data model

mukul-rathi:
As discussed in the maintainers’ sync last week, we chose to move away from IBM’s Activity Streams and use a DTP data model defined within this repo.  The `SocialActivityModel` data model is based off the W3C  Activity Streams standard https://www.w3.org/TR/activitystreams-core/ and based off the current usage of Activity Streams in the existing adaptors. Moving the data representation in-house means we can stick closely to the standard, but we have the flexibility in future to add custom properties. Moreover, the third-party JAR currently used in the open-source is a version “0.0.1-SNAPSHOT” and has not been maintained since 2014.

As part of this addition, we have also updated the existing adaptors to work with the new `SocialActivityModel` representation of social posts. 

We’ve tested the changes with the Google Blogger Importer.

A more detailed description can be found in each of the commits’ summaries.


seehamrun (commented):
Sorry for the late reply, this LGTM but adding chuy/brian to take another look at the model in case they have any historical context

wmorland (approved):
Given @seehamrun was happy and we haven't anything from the other reviewers I'm happy to approve this.

Issue #702, opened 2019-05-15
Consistent Logging Formats


seehamrun:
In some places, we pass in the format string and variables (like in the [LoggingDtpInternalMetricRecorder.java](https://github.com/google/data-transfer-project/blob/master/portability-api-launcher/src/main/java/org/datatransferproject/launcher/metrics/LoggingDtpInternalMetricRecorder.java)). In others we pass a message and an exception (like in the [JobProcessor](https://github.com/google/data-transfer-project/blob/master/portability-transfer/src/main/java/org/datatransferproject/transfer/JobProcessor.java#L83))

Which leads one or the other to being handled incorrectly in the metric implementation (since the method has a String, List<Object> signature). 

Example, with #700, logs from the JobProcess now print out the whole stacktrace but logs from the metric recorder look like this:

> Metric: finishedJob, data type: %s, from: %s, to: %s, success: %s, duration: %s PHOTOS GOOGLE FLICKR false PT0S

lisad:
We should document in the project what the consistent logging format should be - then close this issue - devs are always able to clean up logging messages when they are working on one in a format different than the standard, but a large-scale logging message rewrite is not practical .  Marking this as a Documentation issue.

Nishtha65:
Hi, I would like to work on this issue. Please assign it to me.

PR #951, opened 2021-01-07
Imrpove severe logging with exception in seperate parameter

zhang123cnn:
Summary:
Move the exception severe logging to the seperate parameter. So it can
be better displayed and parsed by logging framework.

Test Plan: Built/tested/loglined

Reviewers: #dtp

Subscribers:

Tasks: T81987523

Tags:

Blame Revision:

Differential Revision: https://phabricator.intern.facebook.com/D25824059

### DIFF
--- a/portability-bootstrap-vm/src/main/java/org/datatransferproject/bootstrap/vm/SingleVMMain.java
+++ b/portability-bootstrap-vm/src/main/java/org/datatransferproject/bootstrap/vm/SingleVMMain.java
@@ -42,7 +42,7 @@ public class SingleVMMain {
     this.monitor = loadMonitor();
     this.errorCallback =
         (e) -> {
-          monitor.severe(() -> format("Exiting abnormally, exception:%n%n" + e));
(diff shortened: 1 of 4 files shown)


zhang123cnn:
@seehamrun Hey Siham, do you mind taking a look at this PR? It should be a simple change to improve the logging. Let me know if you have any concerns.

seehamrun (approved):
Not related, but I think we've brought up issues before with the inconsistent monitoring formats since the Monitor takes a String List as an argument which leads to some confusion depending on the implementation. Details in #702

seehamrun:
Also not sure if you saw, but it looks like this opens up 1 issue: 
```
This pull request introduces 1 alert when merging 46d78ce into d5c4479 - view on LGTM.com

new alerts:

1 for Unused format argument
```

zhang123cnn:
@seehamrun Yes, I am aware of that issue. It was triggered for the first revision of this PR. I have fixed it now so if you check latest LGTM page, it is all good now

PR #958, opened 2021-02-02
Updating Job Watcher service

abhgarg-fb:
Changes done:
Updating JobWatching service to check the job state as Preempted or errored.
This is one of the edge case which is to check whether the current job is errored or preempted, and stop the worker job all together. 

Earlier, the JobWatching service was exiting the system in case job is cancelled.

### DIFF
--- a/portability-api-launcher/src/main/java/org/datatransferproject/launcher/monitor/events/EventCode.java
+++ b/portability-api-launcher/src/main/java/org/datatransferproject/launcher/monitor/events/EventCode.java
@@ -15,6 +15,9 @@ public class EventCode {
   public static final EventCode WORKER_JOB_FINISHED = new EventCode("WORKER_JOB_FINISHED");
   public static final EventCode WORKER_JOB_STARTED = new EventCode("WORKER_JOB_STARTED");
 
+  public static final EventCode WATCHING_SERVICE_JOB_ERRORED = new EventCode("WATCHING_SERVICE_JOB_ERRORED");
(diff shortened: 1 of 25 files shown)


portability-api-launcher/src/main/java/org/datatransferproject/launcher/monitor/events/EventCode.java
```diff
@@ -15,6 +15,9 @@
   public static final EventCode WORKER_JOB_FINISHED = new EventCode("WORKER_JOB_FINISHED");
   public static final EventCode WORKER_JOB_STARTED = new EventCode("WORKER_JOB_STARTED");
 
+  public static final EventCode WATCHING_SERVICE_JOB_ERRORED = new EventCode("WATCHING_SERVICE_JOB_ERRORED");
```
seehamrun, line 18: Why do we want to specify the WATCHING_SERVICE here? Instead of using WORKER_JOB_PREEMPTED or WORKER_JOB_ERRORED?

abhgarg-fb: these are added to distinguish events being logged by WORKER_JOB threads and WATCHING_SERVICE threads. 
Existing event, WORKER_JOB_CANCELLED is left untouched so that existing metrics are left unmatched.

seehamrun (approved):

seehamrun (approved):

abhgarg-fb:
@googlebot I fixed it.

abhgarg-fb:
Github branch got corrupted as update branch from the PR has a bug in github. Hence, cleaning up the older fork and created a new one. 
Now, A new PR is raised to handle all this change.

PR #961, opened 2021-02-02
Switch to using a padded AES encryption transformation

seehamrun:
This should fix #934 

Note: ideally this would be tracked with a unit test, I'll work on adding that with this PR

PR #964, opened 2021-02-16
Include encrypted salt as intended

philipsharp:
EncrypterImpl is encrypting a salt to avoid needing to store the IV, but currently throws it away instead of returning it with the ciphertext. DecrypterImpl then throws away the first decrypted block, intending to throw away the salt, but instead it's part of the plaintext.

This includes the first encrypted block in the return as intended.

This was discovered while testing #961 but the fix is separate.


philipsharp:
@googlebot I fixed it.

smugphilip:
@googlebot I fixed it.

smugphilip:
@googlebot I consent.

seehamrun (approved):

PR #965, opened 2021-02-19
Add a user timezone field to PortabilityJob

KseniiaPelykh:
When a user creates a transfer job we do not consider their timezone on the backend worker and instead all dates simply match to the timezone of the server. To fix this we added a field to the job which exporters and importers can use to render dates in the user's time zone.

- Added the user timezone field to PortabilityJob. 
- Used the timezone field in KoofrPhotosImporter. This means that the dates in the titles get formatted to match the user's timezone preference (if the timezone has not been set we default to the system timezone).

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-koofr/src/main/java/org/datatransferproject/transfer/koofr/photos/KoofrPhotosImporter.java
+++ b/extensions/data-transfer/portability-data-transfer-koofr/src/main/java/org/datatransferproject/transfer/koofr/photos/KoofrPhotosImporter.java
@@ -21,6 +21,7 @@ import java.io.IOException;
 import java.net.HttpURLConnection;
 import java.text.SimpleDateFormat;
 import java.util.Date;
+import java.util.TimeZone;
 import java.util.UUID;
 import org.apache.commons.imaging.Imaging;
 import org.apache.commons.imaging.common.ImageMetadata;
@@ -29,7 +30,7 @@ import org.apache.commons.imaging.formats.tiff.TiffImageMetadata;
 import org.apache.commons.imaging.formats.tiff.constants.ExifTagConstants;
 import org.apache.commons.io.IOUtils;
 import org.datatransferproject.api.launcher.Monitor;
-import org.datatransferproject.spi.cloud.storage.TemporaryPerJobDataStore;
+import org.datatransferproject.spi.cloud.storage.JobStore;
 import org.datatransferproject.spi.transfer.idempotentexecutor.IdempotentImportExecutor;
 import org.datatransferproject.spi.transfer.provider.ImportResult;
 import org.datatransferproject.spi.transfer.provider.Importer;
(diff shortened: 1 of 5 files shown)


extensions/data-transfer/portability-data-transfer-koofr/src/main/java/org/datatransferproject/transfer/koofr/photos/KoofrPhotosImporter.java
```diff
@@ -74,13 +75,17 @@ public ImportResult importItem(
...
-
     monitor.debug(
         () ->
             String.format(
                 "%s: Importing %s albums and %s photos before transmogrification",
                 jobId, resource.getAlbums().size(), resource.getPhotos().size()));
 
+   TimeZone userTimeZone = jobStore.findJob(jobId).userTimeZone();
+    if (null != userTimeZone) {
+      titleDateFormat.setTimeZone(userTimeZone);
```
bancek, line 86: If `KoofrPhotosImporter` is used in a context where multiple jobs can be processed with the same instance concurrently this will cause a data race between multiple jobs.

`titleDateFormat` should be cloned for each `importItem` call and `setTimeZone` can be called for the cloned object.

bancek (changes requested):

bancek:
Code should be formatted using `google-java-format`.

extensions/data-transfer/portability-data-transfer-koofr/src/main/java/org/datatransferproject/transfer/koofr/photos/KoofrPhotosImporter.java
seehamrun: +1 you can probably keep a cache from job id --> timezone to avoid looking up the job each time import is called but it should be separate per job.

KseniiaPelykh: Thank you for your suggestions! I will try to fix it.

extensions/data-transfer/portability-data-transfer-koofr/src/main/java/org/datatransferproject/transfer/koofr/photos/KoofrPhotosImporter.java
```diff
@@ -225,4 +223,20 @@ private Date getDateCreated(PhotoModel photo, byte[] bytes) {
       return null;
     }
   }
+
+  private synchronized SimpleDateFormat getOrCreateTitleDateFormat(UUID jobId) {
+    if (titleDateFormats.containsKey(jobId)) {
+      return titleDateFormats.get(jobId);
+    }
+
+    SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd HH.mm.ss ");
```
bancek, line 232: NITPICK `"yyyy-MM-dd HH.mm.ss "` could be moved to a class constant.

bancek (dismissed):
Looks good.

extensions/data-transfer/portability-data-transfer-koofr/src/main/java/org/datatransferproject/transfer/koofr/photos/KoofrPhotosImporter.java
KseniiaPelykh: Done!

bancek (approved):

wmorland (approved):
Approving from the FB end in addition to the existing approval.

PR #969, opened 2021-03-25
Switch to using Files.createTempFile

### LINKED ISSUE #968: Make temp file location configurable in Jobstore
We use `File.createTempFile` in the PerJobDatastore because a Random Access file is needed for some apis. This is currently only the GoogeVideoImporter and the BackblazeImporters: https://github.com/google/data-transfer-project/search?q=getTempFileFromInputStream

However `File.createTempFile` creates files in the the system temporary directory with file permissions -rw-r--r-- by default.

We'll need to figure out how to hold the file in memory somehow and still be comptiable with the APIs or simply make the location configurable.

seehamrun:
Fixes #968 

Files.createTempFile does not set world readable permissions https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#createTempFile(java.nio.file.Path,%20java.lang.String,%20java.lang.String,%20java.nio.file.attribute.FileAttribute...)


wmorland (approved):
Looks good to me

PR #970, opened 2021-03-25
Explicitly use system default time in console monitor

seehamrun:
We were implicitly using the system default time, while it is still a problem to use the default time we should make it explicit.


wmorland (approved):
Makes sense to me 👍

PR #971, opened 2021-03-25
delete jar from portability-types-common subdir

seehamrun:
Im not sure when this was added, but we no longer use the activity streams jar anyway, so should be deleted


wmorland (approved):
ActivityStreams is odd in that it doesn't seem to be reliably published which is probably why the Jar was added. That's part of the reason we moved away from using it directly and given that move this looks good to me.

seehamrun:
Ah I see yea thats probably why we added the jar then, but since its not needed anymore we shouldnt keep it around. (if we need it in the future, we can always revert this commit)

PR #972, opened 2021-03-25
Add language support

KseniiaPelykh:
Added user's locale field to the job which importers can use to create titles in the user's language.

- Created a base MultilingualDictionary class that uses Resource Bundle to manage multilingual strings.
- Used MultilingualDictionary in GooglePhotosImporter for internationalizing copy prefix.
- Added copy prefix translations into some of the most common languages.  

Didn't add language support to `GoogleVideosImporter` as it requires more time to refactor deprecated `setFileName` and use `NewMediaItem` that supports Unicode.


extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosImporter.java
```diff
@@ -172,7 +175,9 @@ String importSingleAlbum(UUID jobId, TokensAndUrlAuthData authData, PhotoAlbum i
       throws IOException, InvalidTokenException, PermissionDeniedException {
     // Set up album
     GoogleAlbum googleAlbum = new GoogleAlbum();
-    String title = COPY_PREFIX + inputAlbum.getName();
+    String copyOf = getOrCreateStringDictionary(jobId).get(GoogleMultilingualString.CopyOf);
+    String title = MessageFormat.format(copyOf, Strings.nullToEmpty(inputAlbum.getName()));
```
KseniiaPelykh, line 179: `MessageFormat.format` creates a new MessageFormat to do the formatting, so it should be thread safe

seehamrun:
What makes this specific to the GoogleTransfer extension? I'm wondering if we can pull this out to maybe the common directory or something for use by all the other extensions?

seehamrun:
> > What makes this specific to the GoogleTransfer extension? I'm wondering if we can pull this out to maybe the common directory or something for use by all the other extensions?
> 
> We also had this doubt. For now I found "Copy of" without ":", "-" only in GoogleVideosImporter, GooglePhotosImporter and SmugMugInterface, so I thought it could be a good idea to leave every service possibility to define its own translated strings. And in case it is needed we can always refactor and move it to the common directory. But if you think that the probability that the most services will share the same strings is high, I am happy to move them to the common directory in this PR.

I think most services will use the same strings, the differences now are probably oversights on our side when writing the code initially. If a service finds that it needs a different string, it can always be added in separately right?

KseniiaPelykh:
> I think most services will use the same strings, the differences now are probably oversights on our side when writing the code initially. If a service finds that it needs a different string, it can always be added in separately right?

If want to keep everything in common, the service that needs to use a slightly different string will need to find a key to identify the string without creating collision, that can be complicated in case it differs only because of something like "-", ":" or letter case . Or override the resource bundle that adds complexity in case it happens often. But I trust your opinion on this one.

KseniiaPelykh:
> > > What makes this specific to the GoogleTransfer extension? I'm wondering if we can pull this out to maybe the common directory or something for use by all the other extensions?
> > 
> > 
> > We also had this doubt. For now I found "Copy of" without ":", "-" only in GoogleVideosImporter, GooglePhotosImporter and SmugMugInterface, so I thought it could be a good idea to leave every service possibility to define its own translated strings. And in case it is needed we can always refactor and move it to the common directory. But if you think that the probability that the most services will share the same strings is high, I am happy to move them to the common directory in this PR.
> 
> I think most services will use the same strings, the differences now are probably oversights on our side when writing the code initially. If a service finds that it needs a different string, it can always be added in separately right?

I moved everything to the common directory 👍

seehamrun (dismissed):
Sorry for the late review! this looks great! Thanks for putting it together

wmorland (approved):
Re-approving

PR #983, opened 2021-05-06
Update demo-server Dockerfile

philipsharp:
This re-orders the Dockerfile steps for the `demo-server` distribution to run the less-frequently-changing commands first. The DTP-specific files (the demo client and the API/Transfer jar) are copied into the image last. This allows Docker to use the image cache more efficiently and speeds up rebuilding the Docker image when only the DTP code has changed.


smugphilip:
@googlebot I consent.

seehamrun (approved):

PR #987, opened 2021-05-13
Fix casting error for JobHooks array

philipsharp:
When trying to implement a `JobHooksExtension` the following error is thrown:
```
java.lang.ClassCastException: class [Ljava.lang.Object; cannot be cast to class [Lorg.datatransferproject.spi.transfer.hooks.JobHooks;
```

This fixes the casting error by using the typed form of `toArray()`.


smugphilip:
@googlebot I consent.

seehamrun (approved):

PR #988, opened 2021-05-21
Update FlickrPhotosImporter to use temp store as needed

philipsharp:
If the exporter being used for the job put the photo into the temp store, the Flickr importer wouldn't use it.


smugphilip:
@googlebot I consent.

seehamrun (approved):
nice catch!

PR #989, opened 2021-05-26
Update OAuth1Config for additional parameters

philipsharp:
Updates OAuth1Config and the implementing classes to support additional parameters.

The OAuth1 spec allows parameters on the token request and authorization URLs. The old configs only supported a single "scope" parameter. This allows multiple parameters, based on the combination or URL, data type, and mode.

Adds an additional parameter for SmugMug to allow reading and writing of private albums as necessary.

### DIFF
--- a/extensions/auth/portability-auth-flickr/src/main/java/org/datatransferproject/auth/flickr/FlickrOAuthConfig.java
+++ b/extensions/auth/portability-auth-flickr/src/main/java/org/datatransferproject/auth/flickr/FlickrOAuthConfig.java
@@ -16,17 +16,21 @@
 
 package org.datatransferproject.auth.flickr;
 
-import com.google.api.client.auth.oauth.OAuthHmacSigner;
-import com.google.api.client.auth.oauth.OAuthSigner;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 import org.datatransferproject.auth.OAuth1Config;
+import org.datatransferproject.spi.api.auth.AuthServiceProviderRegistry.AuthMode;
 
 /**
  * Class that supplies Flickr-specific OAuth1 info
  * See https://www.flickr.com/services/api/auth.oauth.html
  */
 public class FlickrOAuthConfig implements OAuth1Config {
+
+  private static final String PERMS = "perms";
 
   @Override
   public String getServiceName() {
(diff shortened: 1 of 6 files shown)


extensions/auth/portability-auth-flickr/src/main/java/org/datatransferproject/auth/flickr/FlickrOAuthConfig.java
```diff
@@ -49,17 +53,28 @@ public String getAccessTokenUrl() {
...
+  public Map<String, String> getAdditionalUrlParameters(
+      String dataType, AuthMode mode, OAuth1Step step) {
+    if (dataType.equals("PHOTOS")) {
+      if (step == OAuth1Step.AUTHORIZATION) {
+        if (mode == AuthMode.EXPORT) {
+          return ImmutableMap.of(PERMS, "read");
+        } else {
+          return ImmutableMap.of(PERMS, "write");
+        }
+      }
```
seehamrun, line 74: nit: I think it would be cleaner if you combined the step and mode conditions in the same if statement instead of nesting them 

OR the other thing you could do is create a static multimap, then you could do something like urlparams.get(datatype).get(mode).get(step) here -- this might be overkill for these two configs since they only support 1 datatype each but just a thought.

extensions/auth/portability-auth-smugmug/src/main/java/org/datatransferproject/auth/smugmug/SmugMugOAuthConfig.java
```diff
@@ -49,17 +54,28 @@ public String getAccessTokenUrl() {
...
+      String dataType, AuthMode mode, OAuth1Step step) {
+    if (dataType.equals("PHOTOS")) {
+      if (step == OAuth1Step.AUTHORIZATION) {
+        if (mode == AuthMode.EXPORT) {
+          return ImmutableMap.of(ACCESS, "Full", PERMISSIONS, "Read");
+        } else {
+          return ImmutableMap.of(ACCESS, "Full", PERMISSIONS, "Add");
+        }
+      }
+    }
```
seehamrun, line 76: Same as the Flickr comment above

seehamrun (changes requested):

extensions/auth/portability-auth-flickr/src/main/java/org/datatransferproject/auth/flickr/FlickrOAuthConfig.java
philipsharp: I started with a static map and then decided it was overkill. Lots of extra extra code and repeated data. I'll simplify the steps.

extensions/auth/portability-auth-smugmug/src/main/java/org/datatransferproject/auth/smugmug/SmugMugOAuthConfig.java
```diff
@@ -16,18 +16,23 @@
 
 package org.datatransferproject.auth.smugmug;
```
sparuvu, line 17: nit: Can we extract the duplicate string literals into private static fields?

sparuvu (approved):

seehamrun (approved):

seehamrun:
Looks good to me! Thanks for updating these

extensions/auth/portability-auth-smugmug/src/main/java/org/datatransferproject/auth/smugmug/SmugMugOAuthConfig.java
philipsharp: I made most of them static. I think only duplicate literal I didn't make static is the "Full" permission in the SmugMug config. I'm going to leave that one for now, since it matches the pattern used elsewhere (keys may be static, but values are not) and I think that makes it easier to read.

PR #981, opened 2021-04-29
Handle empty album names in the Smugmug Importer.

seehamrun:
During test imports we would see this error caused by the album name being null. This handles the empty photo album name the same way that the Flickr Importer does, at the Importer level instead of at the interface level. This would allow us then to use some of the language support that was added in #972 more easily 

```
RetryingCallable caught an exception
java.io.IOException: Problem with importer, forcing a retry, first error: java.lang.NullPointerException
	at org.datatransferproject.transfer.smugmug.photos.SmugMugInterface.cleanName(SmugMugInterface.java:251)
	at org.datatransferproject.transfer.smugmug.photos.SmugMugInterface.createAlbum(SmugMugInterface.java:94)
	at org.datatransferproject.transfer.smugmug.photos.SmugMugPhotosImporter.importSingleAlbum(SmugMugPhotosImporter.java:109)
	at org.datatransferproject.transfer.smugmug.photos.SmugMugPhotosImporter.lambda$importItem$0(SmugMugPhotosImporter.java:91)
	at com.google.dataliberation.transfer.worker.extensions.cloud.T2XIdempotentImportExecutor.executeOrThrowException(T2XIdempotentImportExecutor.java:105)
	at com.google.dataliberation.transfer.worker.extensions.cloud.T2XIdempotentImportExecutor.executeAndSwallowIOExceptions(T2XIdempotentImportExecutor.java:84)
	at org.datatransferproject.transfer.smugmug.photos.SmugMugPhotosImporter.importItem(SmugMugPhotosImporter.java:88)
```

PR #991, opened 2021-06-15
Remove copy of prefix for all photos importers

seehamrun:
This also fixes a bug that was introduced in #981 where the copy of prefix gets added twice for the title of the album here: https://github.com/google/data-transfer-project/blob/9614d543897a5b6235344476922457d57e666914/extensions/data-transfer/portability-data-transfer-smugmug/src/main/java/org/datatransferproject/transfer/smugmug/photos/SmugMugInterface.java#L98


holachuy (approved):

smugphilip (dismissed):

philipsharp (dismissed):

seehamrun:
Updated some of the tests, PTAL

holachuy (approved):

PR #992, opened 2021-06-15
Allow exporting from SmugMug private albums

philipsharp:
PR #989 updated the SmugMug OAuth tokens to include "Full" permissions, allowing the SmugMug exporter and importer to operate on non-public resources (including private and passworded albums).

This removes the restriction that disallowed exporting private albums. Export URLs are already being signed with OAuth (despite the comment), and these will now work.


smugphilip:
@googlebot I consent.

seehamrun (approved):

PR #998, opened 2021-08-13
Rename `VideoObject` to `VideoModel` to follow the naming convention we use for most of the other data models

seehamrun:
Ex, we have `PhotosModel` , `TaskModel`, `SocialActivityModel` and several others.


kateyeo (approved):
Looks good

PR #999, opened 2021-08-13
Create MediaContainerResource to be used by the new MediaExporters and MediaImporters

seehamrun:
This combines the photo and video resources, I've reused the PhotoModel and the VideoObject (renaming that to VideoModel in #998) but created a new "MediaAlbum" to replace the Photo/VideoAlbum and also copied over all the transmogrification code that was included in the photoscontainerresource and included videos in that calculation


portability-types-common/src/main/java/org/datatransferproject/types/common/models/media/MediaContainerResource.java
```diff
@@ -0,0 +1,149 @@
...
+          config.getPhotoTitleReplacementCharacter(),
+          config.getPhotoTitleMaxLength());
+    }
+
+    for (MediaAlbum album : albums) {
+      album.cleanName(
+          config.getPhotoTitleForbiddenCharacters(),
+          config.getPhotoTitleReplacementCharacter(),
+          config.getPhotoTitleMaxLength());
+    }
```
sparuvu, line 102: seems like we are using the photos-related config for videos and albums as well. Is that intentional?

seehamrun: Good catch, done!

portability-types-common/src/main/java/org/datatransferproject/types/common/models/videos/VideoModel.java
```diff
@@ -85,6 +86,28 @@ public boolean equals(Object o) {
...
+  // remove all forbidden characters
+  public void cleanName(String forbiddenCharacters, char replacementCharacter, int maxLength) {
+    String name = getName().chars()
+        .mapToObj(c -> (char) c)
+        .map(c -> forbiddenCharacters.contains(Character.toString(c)) ? replacementCharacter : c)
+        .map(Object::toString)
+        .collect(Collectors.joining("")).trim();
+    if (maxLength <= 0) {
+      return;
+    }
```
sparuvu, line 105: nit: 

this could have some performance impact. 
We are unnecessarily processing the name when the maxLength is <= 0. Why not push this condition to be the first stmt of the method. Otherwise, for thousands of files, the name would be processed.

sparuvu (approved):
Looks good (pending https://github.com/google/data-transfer-project/pull/999/files#r694111918)

portability-types-common/src/main/java/org/datatransferproject/types/common/models/videos/VideoModel.java
seehamrun: Hmm I copy pasted this from the photo model code, I think its not quite right, we still need to remove forbidden characters if max length is <=0, but in this code we actually never set the name after we remove the forbidden characters -- ill send a commit to fix this in both classes

seehamrun: Ah I see, this is not a problem for the PhotosModel because we modify the title directly in the class instead of through the setName method.

portability-types-common/src/main/java/org/datatransferproject/types/common/models/TransmogrificationConfig.java
```diff
@@ -1,49 +1,65 @@
 package org.datatransferproject.types.common.models;
 
 /**
-    This class is meant to be extended with custom logic for each service.
-    These defaults represent no-chnage transmogrifications.  
-*/
+ * This class is meant to be extended with custom logic for each service. These defaults represent
+ * no-chnage transmogrifications.
```
kateyeo, line 5: (nit) spelling
```suggestion
 * no-change transmogrifications.
```

kateyeo (dismissed):
Looks good!

kateyeo (approved):

PR #1003, opened 2021-08-24
Create GoogleMediaExporter

seehamrun:
This is essentially a copy of the `GooglePhotosExporter`: https://github.com/google/data-transfer-project/blob/889f1fa92fc54046b42a9448b53d9de0df84912c/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosExporter.java with some very slight changes to support videos

Major Updates: 
- The logic for albumless photos already makes no distinction between photos & videos so that is unchanged
- The logic for the exportMedia function also remained the same, but the function that converts the GoogleMediaItem to a MediaResource previously skipped videos, so this has been updated to handle and include the video in the Container resource 
- Moved out the `convertToPhotoModel` and `convertToVideoModel` functions from the `GoogleVideoExporter` and `GooglePhotosExporter` into the media model itself. 
- If a ContainerResource is provided in the initial export call, the only thing that will be exported are the contents in that container resource. The GooglePhotosExporter had support for a PhotosContainerResource to be passed in, I kept that to help support the migration to the new resource on our end and added additional support for a MediaContainerResource to be passed in. 
- Updated the visibility for some of the methods in the `GooglePhotosInterface`, the interface as a whole should probably be moved out to the mediaModels directory and that should be renamed to be explicit `GooglePhotosApi` or something but that is a problem for a different day.

Note: this is not yet tested and is likely not ready for use, the GoogleMediaImporter will follow in the upcoming weeks.


kateyeo (approved):

seehamrun:
> we should close this in favor of #1004 right?

#1004 is for the importer not the exporter, though this PR is extremely out of date.

jzacsh:
this is obosleted by #1218

PR #1004, opened 2021-08-25
Create GoogleMediaImporter

seehamrun:
This is based off of PR #1006 which pulled out some of the reusable functionality from the GooglePhotosImporter. The GoogleMediaImporter is a copy of that file, but with 2 additional functions: `importVidoes` and `importVideoBatch`

Once PR #1006 is submitted, the diff should be cleaner

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/media/GoogleMediaImporter.java
+++ b/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/media/GoogleMediaImporter.java
@@ -1,0 +1,504 @@
+/*
+ * Copyright 2018 The Data Transfer Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.datatransferproject.datatransfer.google.media;
+
+import static java.lang.String.format;
+
+import com.google.api.client.auth.oauth2.Credential;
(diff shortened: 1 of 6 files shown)


seehamrun:
Ok i have rebased this branch off of PR #1004 so the history got removed, but it is still a copy of the GooglePhotosImporter w/  2 functions added `importVidoes`, `importVideoBatch`

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/MediaResult.java
```diff
@@ -2,11 +2,11 @@
 
 import java.io.Serializable;
 
-class PhotoResult implements Serializable {
+public class MediaResult implements Serializable {
```
jzacsh, line 5: nit: should this file land in the google/media/ folder instead?

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/media/GoogleMediaImporter.java
```diff
@@ -0,0 +1,504 @@
...
+    // Uploads album metadata
+    if (data.getAlbums() != null && data.getAlbums().size() > 0) {
+      for (MediaAlbum album : data.getAlbums()) {
+        idempotentImportExecutor.executeAndSwallowIOExceptions(
+            album.getId(), album.getName(), () -> importSingleAlbum(jobId, authData, album));
+      }
+    }
+
+    long bytes = importPhotos(data.getPhotos(), idempotentImportExecutor, jobId, authData);
+    bytes += importVideos(data.getVideos(), idempotentImportExecutor, jobId, authData);
```
alexeyqu-fb, line 134: We assume that `MediaContainerResource` has either photos or videos and process them at the same time. 
Is it possible that a single `MediaContainerResource` has both?

alexeyqu-fb (commented):
The proposed `GoogleMediaImporter` has pairs of identical methods like `importPhotoBatch`, `importVideoBatch` etc. 
I am not sure I get why do we want to merge it into a common class -- is there a future where photo and video transfers would be merged?

I would rather steer towards having generic `PhotoImporter` and `VideoImporter` that define the API for every other adapter to inherit from, define the common stuff there (counts/bytes reporting, common ways to report errors etc), and then overriding the exact destination-specific import instructions (which API to call, batching, etc) on the `DestinationPhotoImporter` level.

PR #1005, opened 2021-08-31
Skip not found videos in KoofrVideosImporter

calmarj:
If video is deleted in CDN, when skip it in Koofr Video Importer.


wmorland (approved):
Looks good to me

PR #1006, opened 2021-09-09
Some refactoring of the GooglePhotosImporter

seehamrun:
I'm preparing to create the GoogleMediaImporter and noticed a few things that are Photos specific that can be pulled out and reused in the Video context.

- I've moved the logic that imports photos into a separate function 
- I changed the getPhotoDescription method to be a little more generic 
- I split out the logic for creating an input stream from a URL into a different function 
- I also pulled out the code that processed the BatchMedia response into a different function for use w/ videos.

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosImporter.java
+++ b/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosImporter.java
@@ -208,7 +208,6 @@ public class GooglePhotosImporter
     final HashMap<String, PhotoModel> uploadTokenToDataId = new HashMap<>();
     final HashMap<String, Long> uploadTokenToLength = new HashMap<>();
 
-
     // TODO: resumable uploads https://developers.google.com/photos/library/guides/resumable-uploads
     //  Resumable uploads would allow the upload of larger media that don't fit in memory.  To do
     //  this however, seems to require knowledge of the total file size.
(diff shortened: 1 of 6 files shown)


extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosImporter.java
```diff
@@ -202,42 +209,33 @@ long importPhotoBatch(
...
-          inputStream = conn.getInputStream();
-        }
+        Pair<InputStream, Long> inputStreamBytesPair =
+            getInputStreamForUrl(jobId, photo.getFetchableUrl(), photo.isInTempStore());
 
         String uploadToken =
-            getOrCreatePhotosInterface(jobId, authData).uploadPhotoContent(inputStream);
-        mediaItems.add(new NewMediaItem(getPhotoDescription(jobId, photo), uploadToken));
+            getOrCreatePhotosInterface(jobId, authData)
+                .uploadPhotoContent(inputStreamBytesPair.getFirst());
```
wmorland, line 222: Because the input stream gets exhausted in upload I don't think it should be too much of an issue but it might be nice to make sure we close the input stream afterwards either with a finally statement or with a try-with-resources statement. I think this can help with garbage collection

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosImporter.java
```diff
@@ -311,21 +294,55 @@ long importPhotoBatch(
...
+      NewMediaItemResult mediaItem,
+      String idempotentId,
+      IdempotentImportExecutor executor,
+      String title,
+      long bytes)
+      throws Exception {
+    Status status = mediaItem.getStatus();
+    if (status.getCode() == Code.OK_VALUE) {
+      executor.executeAndSwallowIOExceptions(
+          idempotentId, title, () -> new MediaResult(mediaItem.getMediaItem().getId(), bytes));
```
wmorland, line 307: My understanding is that if the result is cached already then we will never hit this point as the photo will already have been filtered out. If this weren't the case and we were to run this for a photo that had already been transferred then this change might not be push safe as we would try to do deserialize a PhotoResult as a MediaResult and Java would complain. I wanted to flag this as we have seen errors before when we changed the return type of a function in the idempotentExecutor.

wmorland (approved):
Looks good to me but I have highlighted in the comments the risk of changing and idempotentExecutor return type for push safety.

kateyeo (approved):

PR #1013, opened 2021-10-13
Add publication instructions

wmorland:
Instructions for publishing to Maven Central


seehamrun:
I noticed that with this new process, the build commit version isnt included in the pom file: https://repo1.maven.org/maven2/org/datatransferproject/portability-data-transfer-google/0.3.58/portability-data-transfer-google-0.3.58.pom 

I forgot how this was generated before but could we make sure to add it in ? or somehow tag which version corresponds to which commit?

wmorland:
@seehamrun I've now addressed your comments. Thanks for the feedback.

seehamrun (approved):

PR #1020, opened 2021-12-01
Throw generic copy exceptions to mark transfers as failed when copy iteration fails in importer

pziebal:
Currently when copy iteration fails in the importer and when the exception thrown is not assignable from CopyExceptionWithFailureReason we mark transfers as successful. Throwing generic copy exception will make sure that transfers are marked as failed.


wmorland (approved):
LGTM. Definitely bad that some Exceptions just aren't being handled at all and this should mean they all bubble up to https://github.com/google/data-transfer-project/blob/995efca635bffc58b6209d040e8c96cd88816880/portability-transfer/src/main/java/org/datatransferproject/transfer/JobProcessor.java#L154

PR #1023, opened 2021-12-13
Photobucket integration

vasylshche:
This code is already in use [by FB](https://www.facebook.com/dtp/) and PB. Contains:
- photos importer
- videos importer
- auth extension


vasylshche:
Pls rescan this PR, as CLA issue should be resolved as well

seehamrun:
This looks good to me in general. Are there plans to also add an exporter? We typically shy away from importer only adapters  in case someone is trying this project out locally or otherwise and wants to get their data exported from the third party service

vasylshche:
> Are there plans to also add an exporter? 
@seehamrun  Most probably not in the nearest future, for several reasons. One of them - for now we are working on the new FE app and as a result rewriting out graphql read layer, so if we do any steps forward writing of the exporter - most probably we will have to significantly update it eventually.

jzacsh:
Hi it looks like this PR's commits got a bit messed up, but the gist of the work appears to be this diff: https://github.com/dtinit/data-transfer-project/pull/1023/files/d3612492bc382c9b9aa388d08e20c34a3327f9c7..ac6ce3194e0764f2c09292fb9d52b58129104604

I'm going to close for now, but if folks start working on photobucket in the future, I'm certain this patchset will be helpful!

PR #1032, opened 2022-01-25
Add gradle tasks for bumping version

KseniiaPelykh:
Added gradle tasks for bumping project versions. This can be an initial step to avoid changing versions manually in the settings file. Ideally, we should manage project version during release.


seehamrun (approved):
Thanks for updating this! how does this task get triggered?

PR #1033, opened 2022-01-25
changed idempotent id for photo import to photo model id

KseniiaPelykh:


extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosImporter.java
```diff
@@ -332,10 +332,6 @@ private long processMediaResult(
         conn.getInputStream(), conn.getContentLengthLong() != -1 ? conn.getContentLengthLong() : 0);
   }
 
-  String getIdempotentId(PhotoModel photo) {
```
KseniiaPelykh, line 335: @seehamrun before opening PR, wanted to check if there is a specific reason we calculate idempotent id for photo as concatenation of album and photoId.

seehamrun: Ah this has some history. 
This was to handle the case where a photo belongs to more than 1 album. Currently the export flow is first we export all of the albums and then we export individual items within those albums. So in the case where we have 

```
album1: photo1, photo2
album2: photo1, photo3
```

we would first upload the images and attach them to album1, but  then when we get to album2, photo1 wouldn't get added to album2 if we only keyed by photo id. So to fix this we added the album id with it as well. This does result in uploading the photo twice but at the time it was an acceptable solution. 

I guess we could change the behavior [here](https://github.com/google/data-transfer-project/blob/00bdc76730e29c18b1d43469fd73f6347cc13cfa/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosImporter.java#L220) to first check for the result of the albumid_photoid in the executor first (in the case that we're retrying a batch for the album) and then check for the photoid and add that to the album without uploading again and after the addition to the album happens also write a result for albumid_photoid.

KseniiaPelykh: Thank you @seehamrun very much for the explanation! In Facebook exporters we assume that the same photo cannot belong to multiple albums - so I was missing this use case. I think it's a good idea to avoid same photo upload, but our necessity is to uniquely track the photo status in the idempotent executor among all photo importers. I thought about using idempotent id for this purpose. I see 2 possibilities to manage this situation:
- We can unify all photo importers to calculate photo Id as concatenation of album id and photo id.
- We can add an additional parameter to the executor to track the import service, and based on this value, we know how to associate idempotent id to actual photo Id. 

Would like to have your opinion on this one - please share in case you have other ideas.

seehamrun: My understanding was that most of the importers already do this? Is that not the case? Can you link the ones that don't... I think the behavior here depends on the behavior of the import service. EX: some services might not support the image getting uploaded once but associated to 2 albums and its not possible to avoid uploading the image twice. 

I'm not sure I understand what you mean by the second bullet point, could you elaborate? I'd prefer that any service specific work get done in the service specific importer/exporter and we shouldn't add any logic to the executor to pull out information. Because we write an ImportResult in the executor, could we not just add the photo id there? (I need to look at this again to refresh my memory but for example for errors, we store the image title and the album information and don't need to parse the ID to get that).

KseniiaPelykh: Hey @seehamrun, my issue is related to our definition of `idempotentId` for photos, not to the upload process. The list of importers that doesn't concatenate album id and photo id for calculating `idempotentId` during photo import:
1. [Backblaze](https://github.com/Neil-Chao/DTP/blob/9f952bb99f462fed039f4b2978ebc906fbfe2590/extensions/data-transfer/portability-data-transfer-backblaze/src/main/java/org/datatransferproject/datatransfer/backblaze/photos/BackblazePhotosImporter.java#L80)
2. [Daybook](https://github.com/google/data-transfer-project/blob/f9afe5c9e387812f9597f1c4d1276b534dfe97e0/extensions/data-transfer/portability-data-transfer-daybook/src/main/java/org/datatransferproject/transfer/daybook/photos/DaybookPhotosImporter.java#L86)
3. [Twitter](https://github.com/google/data-transfer-project/blob/889f1fa92fc54046b42a9448b53d9de0df84912c/extensions/data-transfer/portability-data-transfer-twitter/src/main/java/org/datatransferproject/transfer/twitter/TwitterPhotosImporter.java#L65)

Regarding the second bullet: we want to update the item (photo in this case) transfer state in real time in the IdempotentImportExecutor, it's easier to accomplish in case we have a unique way of calculating `idempotentId` among data types.

PR #1034, opened 2022-01-31
Switch BackblazeTransferClientFactory from using 1 single client for every job to keeping a map of clients based on JobId

seehamrun:
this prevents the race condition where we need to use a different set of auth data for a different job but use the old client factory


wmorland (approved):
Looks good to me. Thanks!

PR #1040, opened 2022-02-09
Handle Backblaze FileNotFoundException for videos

wmorland:
If a URL hit FNF exception then the data has been deleted and we should skip rather than retrying or failing. This is based on how both Google and Koofr handle this exception in their adapters.

As we introduce detailed transfers we may move to marking these items specifically as skipped which can then be surfaced to users

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-backblaze/src/main/java/org/datatransferproject/datatransfer/backblaze/videos/BackblazeVideosImporter.java
+++ b/extensions/data-transfer/portability-data-transfer-backblaze/src/main/java/org/datatransferproject/datatransfer/backblaze/videos/BackblazeVideosImporter.java
@@ -16,6 +16,7 @@
 
 package org.datatransferproject.datatransfer.backblaze.videos;
 
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.UUID;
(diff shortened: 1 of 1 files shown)


extensions/data-transfer/portability-data-transfer-backblaze/src/main/java/org/datatransferproject/datatransfer/backblaze/videos/BackblazeVideosImporter.java
```diff
@@ -78,11 +79,16 @@ public ImportResult importItem(
...
+        imageStreamProvider.getConnection(video.getContentUrl().toString()).getInputStream()) {
 
-    return b2Client.uploadFile(
-        String.format("%s/%s.mp4", VIDEO_TRANSFER_MAIN_FOLDER, video.getDataId()),
-        jobStore.getTempFileFromInputStream(videoFileStream, video.getDataId(), ".mp4"));
+      return b2Client.uploadFile(
+          String.format("%s/%s.mp4", VIDEO_TRANSFER_MAIN_FOLDER, video.getDataId()),
+          jobStore.getTempFileFromInputStream(videoFileStream, video.getDataId(), ".mp4"));
+    } catch (FileNotFoundException e) {
+      monitor.severe(
```
seehamrun, line 89: (just to refresh my memory), how does this behave with the idempotent executor? Would we want to transform this execption into an IOException so it will get swallowed and recorded? (in what other ways would this get marked as a "failure" in the job record?

seehamrun (approved):
LGTM just a clarifying question

extensions/data-transfer/portability-data-transfer-backblaze/src/main/java/org/datatransferproject/datatransfer/backblaze/videos/BackblazeVideosImporter.java
wmorland: So for this the video has actually already been deleted so we don't mark this as a failure. With our work to add more details to transfers we plan to add a new state where we can mark an item specifically as skipped and will apply that to this use case but for now this is a very rare edge case where an item has just been deleted so we skip without marking the job as failed.

PR #1049, opened 2022-03-11
Add hook for Job Cancellations

saurabgu:
This change adds capability to add hooks for Job workers to do some clean up at their end when Jobs are intentionally/abruptly cancelled or fail. 
The new addition will allow JobStore implementations to provide their choice of handling for workers when Jobs are abandoned.


wmorland (approved):
All checks pass and this is a safe change because of the no-op default implementation. LGTM @seehamrun just flagging this again

PR #1057, opened 2022-04-04
Fix a wrong package name of blob models

cimfalab:
I've seen the compilation problem of blob models so fixed to "types.common":
* Its package is "types.transfer" but located under "types.common" directory.
* 85ca4b2 commit maybe means to move the models from "types.transfer" to "types.common" package.


seehamrun:
Thanks for catching this. Its not clear why this wasnt caught with the checks on the original PR #1002... but thats okay. I'll wait for these checks to finsh running and then merge

seehamrun (approved):

PR #1063, opened 2022-04-20
Add unit tests for Daybook importers

xokker:
Summary:
1. Added 2 basic non-exhaustive test suites for both Daybook importers.
2. Moved direct calls to `JobMetadata` upstream to make the importers testable.
3. Removed unused paratermers here and there.
4. Some other minor refactorings such as better generics.

Test Plan: `./gradlew ':extensions:data-transfer:portability-data-transfer-daybook:test'`

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-daybook/build.gradle
+++ b/extensions/data-transfer/portability-data-transfer-daybook/build.gradle
@@ -23,10 +23,12 @@ dependencies {
     compile project(':portability-spi-cloud')
     compile project(':portability-spi-transfer')
     compile project(':portability-transfer')
-    
+
     compile("com.google.api-client:google-api-client:${googleApiClient}")
     compile "com.squareup.okhttp3:okhttp:${okHttpVersion}"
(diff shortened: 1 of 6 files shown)


xokker:
Looks like my force-push broke the code analysis, I'll create a separate PR.

jzacsh:
> Looks like my force-push broke the code analysis, I'll create a separate PR.

Just for my edification, I thought github prevents this kind of problem? (We have `Permit force pushes for all users with push access.` setting unticked). Are you sure it was a force-push issue?

I'm asking because I want to know how not to make such mistakes myself :laughing:  (sometimes I feel safe force pushing to my own private repo if it's pretty new, so it'd be good to know how/if this can bite me).

xokker:
> Are you sure it was a force-push issue?

I am not :D 

Come to think of it, it probably wasn't related to the force push. I created https://github.com/google/data-transfer-project/pull/1064 right after this one and I remember it failing with the same signals as this PR. If I were to bet, I'd say there was an intermittent issue with our build system.

PR #1072, opened 2022-04-30
Remove usage of old Pair in GooglePhotosImporter

seehamrun:
Pair is not supported by Guava and we were using an older gdata pair, switch to using supported apache pair instead

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosImporter.java
+++ b/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosImporter.java
@@ -19,6 +19,7 @@ import static java.lang.String.format;
 
 import com.google.api.client.auth.oauth2.Credential;
 import com.google.api.client.json.JsonFactory;
+import com.google.auto.value.AutoValue;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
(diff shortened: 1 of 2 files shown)


kateyeo (approved):

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosImporter.java
```diff
@@ -218,11 +218,11 @@ long importPhotoBatch(
...
             getInputStreamForUrl(jobId, photo.getFetchableUrl(), photo.isInTempStore());
 
-        try (InputStream s = inputStreamBytesPair.getFirst()) {
+        try (InputStream s = inputStreamBytesPair.getLeft()) {
           String uploadToken = getOrCreatePhotosInterface(jobId, authData).uploadPhotoContent(s);
           mediaItems.add(new NewMediaItem(cleanDescription(photo.getDescription()), uploadToken));
           uploadTokenToDataId.put(uploadToken, photo);
-          uploadTokenToLength.put(uploadToken, inputStreamBytesPair.getSecond());
+          uploadTokenToLength.put(uploadToken, inputStreamBytesPair.getRight());
         }
```
xokker, line 226: Is there a reason we don't use `InputStreamWrapper` directly instead of working with pairs?

```
InputStreamWrapper streamWrapper = jobStore.getStream(jobId, fetchableUrl);
try (InputStream s = streamWrapper.getStream()) {
  ....
  uploadTokenToLength.put(uploadToken, inputStreamBytesPair.getBytes());
}
```

It would also allow us to get rid of `inputStreamBytesPair` method.

xokker (approved):
The change looks good to me, but I believe it would be even better to remove the use of pairs completely.

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosImporter.java
xokker: Please disregard my review, I totally missed the fact `getInputStreamForUrl` does more than just creating a stream wrapper. Sorry

PR #1080, opened 2022-05-31
Fix equals bug in photos resource

xokker:
Changes:
* `!equals` instead of `!=` when comparing string album IDs;
* simpler `PhotoAlbum::split`;
* basic tests for `PhotoAlbum`;
* formatting of every file I touched.

Looks like we use this `!=` comparison to find the album for the given group of photos. It was introduced in https://github.com/google/data-transfer-project/pull/805. I *suspect* the comparison bug didn't really show itself because the album ids for photos are probably reused everywhere (therefore have the same pointer).

I'll make the container resource easier to understand in the following PRs.


wmorland (approved):
LGTM

PR #1097, opened 2022-06-22
Make exception mandatory on error

xokker:
1. Every error must have an exception. If a client does not have one it has to create one.
2. Added some unit tests.

Test plan: `ItemImportResultTest`.


wmorland (approved):
Lgtm. Just the check and tests.

PR #1098, opened 2022-06-23
Switch use from `JacksonFactory` to `GsonFactory`.

Capstan:
This is mirroring a global change we're making in the monorepo to reduce concerns from security.


Capstan:
Gradle is strangely reporting a seemingly-unrelated Guava error when I build with these changes:

```
> Task :portability-transfer:compileJava FAILED
/usr/local/google/home/nherring/github/data-transfer-project/portability-transfer/src/main/java/org/datatransferproject/transfer/WorkerModule.java:18: error: cannot find symbol
import static com.google.common.collect.MoreCollectors.onlyElement;
                                       ^
  symbol:   class MoreCollectors
  location: package com.google.common.collect
/usr/local/google/home/nherring/github/data-transfer-project/portability-transfer/src/main/java/org/datatransferproject/transfer/WorkerModule.java:18: error: static import only from classes and interfaces
import static com.google.common.collect.MoreCollectors.onlyElement;
^
2 errors
```

I cannot tell if my depends are perturbing something in the Guava version resolution machine.

jzacsh (commented):
Left a comment but otherwise LGTM. I may hold this off just to do some internal testing.

Capstan:
`./gradlew test` from the root is what I was running.

Capstan:
And the `demo-server` `build.gradle` blanket overrides the guava version for all artifacts, which doesn't work for the versioning scheme for `listenablefuture` or `failureaccess`.

distributions/demo-server/build.gradle
```diff
@@ -197,7 +197,7 @@ task installToolsAndDockerize {
 configurations.all {
     resolutionStrategy {
         eachDependency { DependencyResolveDetails details ->
-            if (details.requested.group == 'com.google.guava') {
+            if (details.requested.group == 'com.google.guava' && (details.requested.name != 'failureaccess' && details.requested.name != 'listenablefuture')) {
                 details.useVersion("${guavaVersion}")
                 details.because 'Some dependencies rely on lower version which does not support MoreCollections'
```
Capstan, line 202: I wonder if having this `resolutionStrategy` in `portability-transfer/build.gradle` would also have worked around the original problem?

This whole thing seems quite fragile.

portability-transfer/src/main/java/org/datatransferproject/transfer/WorkerMain.java
```diff
@@ -77,7 +77,7 @@ public void initialize() {
     // TODO this should be moved into a service extension
     extensionContext.registerService(HttpTransport.class, new NetHttpTransport());
     extensionContext.registerService(OkHttpClient.class, new OkHttpClient.Builder().build());
-    extensionContext.registerService(JsonFactory.class, new JacksonFactory());
+    extensionContext.registerService(JsonFactory.class, GsonFactory.getDefaultInstance());
```
seehamrun, line 80: I have a slight concern with updating this in the extension context because it gets pulled in arbitrarily in other classes.. ex: it gets piped through to the GooglePhotosInterface through this line: https://github.com/google/data-transfer-project/blob/889f1fa92fc54046b42a9448b53d9de0df84912c/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/GoogleTransferExtension.java#L76,L109 

I'm not sure how this would behave if another service is using Jackson somewhere else in their code but pulling in this Gson factory instead. 

I would guess there shouldnt be too many instances of this but i have not done the analysis

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosInterface.java
```diff
@@ -134,7 +134,7 @@ MediaItemSearchResponse listMediaItems(Optional<String> albumId, Optional<String
     if (pageToken.isPresent()) {
       params.put(TOKEN_KEY, pageToken.get());
     }
-    HttpContent content = new JsonHttpContent(new JacksonFactory(), params);
+    HttpContent content = new JsonHttpContent(GsonFactory.getDefaultInstance(), params);
```
seehamrun, line 137: I wonder if we can just use `this.jsonFactory` instead of creating a new instance here and below on L171?

Capstan: That does seem like a better approach overall.

Capstan: Done.

distributions/demo-server/build.gradle
seehamrun: Hiya! sorry for the delay on this review, I was discussing this with zacsh@ yesterday and there are a few things that arent clear with this... 

1) where/how did you determine that you also needed `failureaccess` and `listenablefuture` here? Its not obvious from the logs 
2) If we delete this block entirely, would everything just work? 
3) Im looking through to see if there are other places where we explicitly make a new JacksonFactory, in theory there shouldn't be since each class should use the one that's registered in the extensioncontext in WorkerMain.java, but we've seen that the GooglePhotosInterface broke that :) so we should double check the project for other instances of this.

Capstan: If you remove them, you end up with failures to find those targets at the overridden version when you use the latest guava (which includes them and includes versioned references to them, and those versions get edited, IIRC).

portability-transfer/src/main/java/org/datatransferproject/transfer/WorkerMain.java
Capstan: I updated GooglePhotosInterface to use the provided `JsonFactory` so it will at least be consistent with whatever is done here.

distributions/demo-server/build.gradle
Capstan: For 2; I don't know. I have definitely seen weird compilation failures related to be should-be-findable-but-aren't methods in `MoreCollections`. I don't understand why version resolution doesn't force a later version of Guava that has them naturally, but I haven't fully internalized gradle/Java's semantics here. I can try it later, but I'm OOO for the next 4d.

For 3, I ran a bulk editor in the monorepo that looked at `JacksonFactory` use, but it appears not to have found all the references in here, so I'll need to do a second pass manually.

Capstan: Had more time (and battery) than I thought, and I tried out removing the entire `resolutionStrategy` block, and that seems to have worked, so I'm going to roll that into the guava update commit.

Capstan:
Ugh. `git rebase -i` is not my friend. I redid the edits three different times and still managed to drop them on the floor. :(

distributions/demo-server/build.gradle
jzacsh: > For 3, I ran a bulk editor in the monorepo that looked at JacksonFactory use, but it appears not to have found all the references in here, so I'll need to do a second pass manually.

For context on this: we don't import the _entirety_ of this codebase into the monorepo (in case you have any other cases like ours in your lsc seeing only subsets of stuff).

seehamrun (approved):
this LGTM! Thanks for doing this cleanup

Capstan:
Is this waiting on an affirmative review by @jzacsh and/or @wmorland, or was that advisory only? In the latter case, how long until it's worth merging and dealing with potential aftermath of a late review (if any) later?

Capstan:
Still waiting on a review.

wmorland (approved):
Reviewed and looks good to me. From our side the biggest change may be the guava upgrade which we might require to propogate.

Issue #1056, opened 2022-03-31
Links to data models are dead


slash-dev:
On the following pages:

* [ ] fixed in PR #1070: https://github.com/google/data-transfer-project/tree/master/portability-types-transfer
* [x] fixed in PR #1099: ~~Website: https://datatransferproject.dev/documentation~~

Thanks!

PR #1099, opened 2022-06-27
fix public website links broken by model refactor

jzacsh:
fixes 1/2 of #1056 (context: models were moved in #916 but these links were hardcoded into the public site).

note: there's actually a lot of dead files in the site repo (eg `dtp-lockup.reverse.jpg`) and other strange things (like the site's source is deployed with the site), but I'm not touching them since they don't get in the way (the markdown I touched because it had the outdated link).


jzacsh:
actually it looks like we were _once_ maybe using [this method](https://jekyllrb.com/docs/continuous-integration/github-actions/#workspace-setup) which has this very loud warning:

> The Action we’re using here will create (or reset an existing) gh-pages branch on every successful deploy.
> So, if you have an existing gh-pages branch that is used to deploy your production build, ensure to make a backup of the contents into a different branch so that you can rollback easily if necessary.

This makes me even less confident anything more than trivially "can't crash this site"-changes. I'll leave the mystery-solving for a different branch on a rainy day: [`fix-reproducible-site-build`](https://github.com/jzacsh/data-transfer-project/tree/fix-reproducible-site-builds) ([diff](https://github.com/google/data-transfer-project/compare/gh-pages...jzacsh:data-transfer-project:fix-reproducible-site-builds?expand=1))

**tl;dr** I'm 70% sure this site isn't deployed with any ci/cd but rather folks were hand-tweaking `_site/` folder by running jekyll locally (though I can't find documentation that github will automatically only server `_site/`, so I'm not 100% sure).

holachuy (approved):
Thx for fixing

PR #1100, opened 2022-06-27
cleanup of site and its maintenance/documentation

jzacsh:
builds atop #1099 to try to cleanup our site's generation so updating is a bit easier to understand without making mistakes (like, eg, deleting the site).


jzacsh:
@lisad feel free to replicate the cleanup-sentiment of course if you want, but I'm closing this since it's old now. (I think the problems are still true: worth just documenting how to the site's built, and deleting ephemeral files)

PR #1079, opened 2022-05-23
Stricter idempotent import executor api

xokker:

PR #1102, opened 2022-06-29
Migrate Daybook importer to new executor API

xokker:
Following https://github.com/google/data-transfer-project/pull/1079 migrating Daybook photo importer to the new API. The new API supports size reporting.


extensions/data-transfer/portability-data-transfer-daybook/src/main/java/org/datatransferproject/transfer/daybook/photos/DaybookPhotosImporter.java
```diff
@@ -127,7 +128,7 @@ private int importPhoto(
       inputStream = new URL(photoModel.getFetchableUrl()).openStream();
     } else {
       monitor.severe(() -> "Can't get inputStream for a photo");
```
jzacsh, line 130: Big improvement in-passing here, thakns for doing this; (`IOException` > `-1` in my book)


just a nitpick below to factor out the message and document the magic `null` being passed.
```suggestion
      String errorMessage = "Can't get inputStream for a photo";
      monitor.severe(() -> errorMessage);
      return ItemImportResult.error(new IOException(errorMessage), null /*sizeInBytes*/);
```

jzacsh (approved):
LGTM (pending small comment I left you)

extensions/data-transfer/portability-data-transfer-daybook/src/main/java/org/datatransferproject/transfer/daybook/photos/DaybookPhotosImporter.java
xokker: I generally don't do argument comments because IDEA does exactly that automatically
Screenshot 2022-07-04 at 11 54 50 am


I don't mind adding comments as well, but I'd prefer it to be a universal rule all of us can always stick to. Shall we codify a rule? I wonder if it is doable through google-java-format config.

jzacsh: tl;dr I've added to our ongoing "things to discuss" doc.

> I generally don't do argument comments because IDEA does exactly that automatically

Ah, yeah makes sense. Doesn't work for someone without that IDE feature, but I think to your point below - it's more important that we're consistent than that we have to bother figuring things out like this on every review.

> I don't mind adding comments as well, but I'd prefer it to be a universal rule all of us can always stick to. Shall we codify a rule? I wonder if it is doable through google-java-format config.

100% agree; sorry I'm used to a particular style guide and forget I'm not working with that in this project :facepalm: 
But yeah, going forward, we should just let something enforce .. _some_ styleguide for us so we don't even have to discuss.

jzacsh: Oh, whoops - unmarking "resolved" as I think your thumbs-up was that you still wanted to factor out the errorMessage.

(feel free to mark resolved to merge - I don't think this change is a big deal!)

jzacsh (approved):

xokker:
Sorry for back and forth @jzacsh, but looks like I'll need another approval from you

wmorland (approved):
LGTM

PR #1107, opened 2022-07-02
fix #1106 cleanup IdempotentImportExecutorHelper

### LINKED ISSUE #1106: one-off idempotent-ID computation is obsolete with newer DownloadableItem work
need to continue the cleanup in #1079 by deleting IdempotentImportExecutorHelper and converting its usages to [`ImportableItem#getIdempotentId`](https://github.com/google/data-transfer-project/blob/1fdf324b7be73a7bcb1e9f6f9ec5166217a861db/portability-types-common/src/main/java/org/datatransferproject/types/common/ImportableItem.java#L11).

essentially we're switching now from having one-off implementations of ID building to an interface (I'm filing this because it's actually important to me as a open-closed-principle issue, since I'm extending via a new Media model and don't want to keep modifying the [`IdempotentImportExecutorHelper` class](https://github.com/google/data-transfer-project/blob/02e9918a0182795666b21414f296a6c256eed963/portability-spi-transfer/src/main/java/org/datatransferproject/spi/transfer/idempotentexecutor/IdempotentImportExecutorHelper.java)).

jzacsh:
fix #1106 cleanup rIdempotentImportExecutorHelper

replaces the older `IdempotentImportExecutorHelper` with newer
`ImportableItem#getIdempotentId` interface's usage.

motivation: I'm tackling this because it's actually a crufty
open-closed-principle issue I hit in #1054 where I'm extending via a new
`Media` model and don't want to keep modifying the
`IdempotentImportExecutorHelper`


xokker (approved):
Thank you for cleaning up the importers! 

My only concern is _deletion_ of the helper class—it's a backward-incompatible change. Are we comfortable with it? Personally I don't mind deleting it—Meta's code does not depend on it. However, we might want to be more stringent about it and start with marking the class `@Deprecated` and deleting in the next major version.

All of that brings us to versioning discussion. I promised to prepare an RFC on versioning, but haven't done it yet. Will probably do it this month.

xokker (approved):

kateyeo (approved):

PR #1108, opened 2022-07-02
add video logic to the new MicrosoftMedia* classes

jzacsh:
this finishes the copy-pasta that #1054 starts.

DO NOT MERGE before #1054 is merged!


portability-types-common/src/main/java/org/datatransferproject/types/common/models/photos/PhotoModel.java
```diff
@@ -94,14 +95,28 @@ public String getDescription() {
     return description;
   }
 
+  // TODO(zacsh) remove this in favor of getFolderId
```
kateyeo, line 98: Is getFolderId the right method to replace this? Based on the name I would think mediaType wouldn't correspond with folderId

portability-types-common/src/main/java/org/datatransferproject/types/common/Fileable.java
```diff
@@ -0,0 +1,14 @@
+package org.datatransferproject.types.common;
+
+/**
+ * Represent an item we expect to be recognizable as an ordinary file on someone's computer.
+ */
+public interface Fileable {
+  /**
+   * Returns a standard MIME-type string one would expect to find if they inspected th edownloaded
```
kateyeo, line 8: nit: spelling
```suggestion
   * Returns a standard MIME-type string one would expect to find if they inspected the downloaded
```

kateyeo (approved):

portability-types-common/src/main/java/org/datatransferproject/types/common/models/photos/PhotoModel.java
jzacsh: Oh good catch. This should have referred to `getMimeType` (and even that needs a follow up comment because as I've learned with my more recent commit, that is in conflict with another interface that provides `getEncoding` (which is the same thing)).

Will fix before merging.

jzacsh: Okay so because PhotoModel.java doesn't use MediaObject (and thus doesn't have a `getEncodingFormat`) I'm going to punt this issue and leave this as one of our TODOs to rethink some of our models.

In the meantime I've clarified the relationship of the newer, smaller `Fileable` to the older `MediaObject`

portability-types-common/src/main/java/org/datatransferproject/types/common/Fileable.java
jzacsh: fixed, thanks!

PR #1114, opened 2022-07-05
New Java and Gradle versions

xokker:
* Java 8 –> 11
* Gradle 4 –> 6
* Shadow-jar plugin 2 –> 6
* Auto-value framework 1.6 –> 1.9


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
-#Mon Sep 24 14:23:52 CEST 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.9.2-bin.zip
```
jzacsh, line 3: Looks good - but so we know going forward: is there a particular reason we're picking 6.9 here? That is: I see [Gradle's as far as 7.x now](https://gradle.org/releases) - should we just pick the latest?

xokker: Something to do with the new plugin format. There were a bunch of errors I didn't want to investigate :D Decided we can start with 6.X first and upgrade to the next one separately.

jzacsh (approved):

PR #1117, opened 2022-07-13
Add job storage support for google video imports

incjo:
Currently, `GoogleVideoImporter` doesn't support the `isInTempStore` flag and doesn't load videos passed via temporary job storage correctly. This functionality is already available for photo imports. This PR enables it for videos as well.

It is needed in order to support video creation timestamps when transferring videos from Facebook.


extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/videos/GoogleVideosImporter.java
```diff
@@ -313,6 +308,18 @@ private Pair<String, Long> uploadMediaItem(
     }
   }
 
+  private File createTempVideoFile(UUID jobId, VideoModel inputVideo) throws IOException {
+    try (InputStream inputStream = getVideoInputStream(jobId, inputVideo)) {
+      return dataStore.getTempFileFromInputStream(inputStream, inputVideo.getName(), ".mp4");
```
jzacsh, line 313: (not even from your PR, so don't block your work on this comment) does this hardcoding of `.mp4` seem right? I guess it's fine if the files passing through here are only ever mp4s, but.. I'd also hope our models would hold that kind of information instead of our adapters :confused:

jzacsh (approved):

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/videos/GoogleVideosImporter.java
incjo: Interesting point. 

I believe it to be superfluous. The file extension itself doesn't determine anything in this case and. E.g., the file we pick up from the temp store doesn't have an extension! In the end, it's just a mnemonic for the users as well as the operating system and doesn't serve a practical purpose in this context.

On a different note, I believe you might be pointing out a potential bug. I feel like calling `getTempFileFromInputStream` in such a manner might be introducing a race condition of sorts. It only depends on the video name which isn't sufficient to differentiate it from other instances of the same process, i.e., if a different job was migrating a video with the same name, they could write this file at the same time. I believe it shares the same directory. 

Perhaps a more robust approach to consider would be to call it in the following fashion:
`getTempFileFromInputStream(inputStream, jobId.toString(), inputVideo.getName());` that way it would be under the job's namespace. Would need to investigate further under its own ticket.

PR #980, opened 2021-04-28
Switching gradle release publishing to maven central

wmorland:
We are switching to publishing directly to Maven Central now that Bintray is sunsetting https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter

PR #1118, opened 2022-07-13
Bump version to 0.3.68

alexeyqu-fb:


gradle.properties
```diff
@@ -43,6 +43,6 @@ apacheHttpVersion=4.5.6
 # ossrhUsername=
 # ossrhPassword=
 
-# signing.keyId=F022E8B0
```
alexeyqu-fb, line 46: that was added in #980 -- @wmorland do you think we should clean this up?

seehamrun:
Should this be closed in favor of #1121 ?

PR #1109, opened 2022-07-04
[Google] skip missing albums during importPhotoBatch and add debug logging

alexeyqu-fb:
This PR has 2 changes:
* skipping the `The provided ID does not match any albums` exceptions thrown from [mediaItems/batchCreate](https://developers.google.com/photos/library/reference/rest/v1/mediaItems/batchCreate) by catching those;
* getting more debug information by querying for album afterwards (`getAlbum`) and printing the existing fields + 2 more: `mediaItemsCount` and `isWritable`.

The fields `mediaItemsCount` and `isWritable` are returned in the json response of [albums/get](https://developers.google.com/photos/library/reference/rest/v1/albums/get) method of Google Photos API.

Added 2 tests:
- no throw on the `The provided ID does not match any albums`  exception;
- throw on some other exception from `createPhotos`.

PR #1120, opened 2022-07-15
fix typo introduced in #1109

alexeyqu-fb:
The [API reference](https://developers.google.com/photos/library/reference/rest/v1/albums#Album) defines `isWriteable` property, not `isWritable`.


wmorland (approved):
Typos are like that sometimes 🫂

PR #1126, opened 2022-07-21
Migrate Koofr video impoter to new executor API

xokker:
Following https://github.com/google/data-transfer-project/pull/1079 migrating Koofr video importer to the new API.

I had to add a new model to support 'dynamic' idempotent id implementation.

I introduced `MediaObjectTest` to show the behaviour of `KoofrVideoModel` constructor is safe.


portability-types-common/src/test/java/org/datatransferproject/types/common/models/MediaObjectTest.java
```diff
@@ -0,0 +1,48 @@
...
+ * limitations under the License.
+ */
+
+package org.datatransferproject.types.common.models;
+
+import static org.junit.Assert.*;
+
+import org.junit.Test;
+
+public class MediaObjectTest {
```
jzacsh, line 23: thanks for adding more test coverage! these tests are great, but I'm just curious: is there one particular case you're seeing in practice (that inspired your test coverage)? Eg: is it empty strings coming back? or are you getting non-URI ("illegal") content set on the objects?

jzacsh (changes requested):

PR #1134, opened 2022-08-08
Add album support to google video importer

incjo:
Currently, all videos transferred to Google Photos end up at the top level. They're not assigned to any album. The albums specified via `VideoModel` are ignored.

This change adds support for specifying albums for video transfers. It's duplicating the functionality already present in the `GooglePhotosImporter`. Transfers where the album isn't specified work normally like before.

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/common/GooglePhotosImportUtils.java
+++ b/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/common/GooglePhotosImportUtils.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2022 The Data Transfer Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.datatransferproject.datatransfer.google.common;
+
+import com.google.common.base.Strings;
+
+public class GooglePhotosImportUtils {
+
+  public static String cleanAlbumTitle(String originalTitle) {
+    String title = Strings.nullToEmpty(originalTitle);
+
+    // Album titles are restricted to 500 characters
+    // https://developers.google.com/photos/library/guides/manage-albums#creating-new-album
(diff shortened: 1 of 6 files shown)


extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/common/GooglePhotosImportUtils.java
```diff
@@ -0,0 +1,43 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.datatransferproject.datatransfer.google.common;
+
+import com.google.common.base.Strings;
+
+public class GooglePhotosImportUtils {
```
seehamrun, line 20: This seems like some work that might already be present on the model itself: https://github.com/google/data-transfer-project/blob/master/portability-types-common/src/main/java/org/datatransferproject/types/common/models/videos/VideoModel.java#L126-L141

Can we add these utility classes on the model instead of adding them here?

seehamrun (approved):
this looks good to me overall, just a small nit about the name cleanup utility class that was added

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/common/GooglePhotosImportUtils.java
incjo: Thanks for bringing it up. I see that there's `TransmogrificationConfig` along with the `transmogrify` method in the model which is made to solve that exact problem. However, I lack the specific knowledge to implement the correct config for Google Photos. Perhaps it could be done in a dedicated ticket.

jzacsh, line 20:
```suggestion
// TODO(#yyyyyyy) consider porting this logic to a transmogrify config (and thus deleting this class).
public class GooglePhotosImportUtils {
```

jzacsh (commented):
just a nitpick to leave a TODO for a better solution wrt what I *think* we're duplicating against the transmogrify framework (worth spot-checking before you file the TODO-bug, though).

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/common/GooglePhotosImportUtils.java
jzacsh: ah, I see you [already discussed](https://github.com/google/data-transfer-project/pull/1134#discussion_r946667754) this exact thing -- yeah a TODO-tracker is sufficient I think.

jzacsh (dismissed):

wmorland (approved):
Approving to merge

PR #1135, opened 2022-08-08
Fix Java debugging of a running docker container

xokker:
I broke it when migrating to Java 11 in https://github.com/google/data-transfer-project/pull/1114.


wmorland (approved):
Thank you!

seehamrun (approved):

PR #1137, opened 2022-08-09
Add junit5 jupiter engine dependency & fixes

chaaransen:
* Added Jupiter engine dependency
* Added Unit test logging indicating test run state.
* Fixes the issue where some Junit5 tests might not be picked up & tested.
* Modified Junit5 @Test to Junit4 annotations on files that require other migration to be done before upgrading, which might fail if updated now. (eg) @Rule is not supported in Junit5 @Test classes


xokker (approved):
Thanks a lot for fixing my mistake!

PR #1139, opened 2022-08-11
Upgraded build.gradle methods

chaaransen:
compile & testCompile are deprecated and will fail with error when we upgrade to Gradle 7.0+ hence upgraded with implementation & testImplementation.


xokker (approved):
Thank you for keeping the repo up-to-date!

PR #1145, opened 2022-08-18
Migrate parameterized JUnit test to Junit5

chaaransen:
Migrated Parameterized runnner to parameterized test, since there's no class level support from Junit5+, with @MethodSource referencing the static parameter method.

### DIFF
--- a/portability-transfer/src/test/java/org/datatransferproject/transfer/copier/PortabilityInMemoryDataCopierTest.java
+++ b/portability-transfer/src/test/java/org/datatransferproject/transfer/copier/PortabilityInMemoryDataCopierTest.java
@@ -16,6 +16,12 @@
 
 package org.datatransferproject.transfer.copier;
 
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Optional;
+import java.util.Stack;
+import java.util.UUID;
 import org.datatransferproject.api.launcher.Monitor;
 import org.datatransferproject.copier.stack.PortabilityStackInMemoryDataCopier;
 import org.datatransferproject.spi.cloud.storage.JobStore;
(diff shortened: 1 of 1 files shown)


portability-transfer/src/test/java/org/datatransferproject/transfer/copier/PortabilityInMemoryDataCopierTest.java
```diff
@@ -103,20 +102,15 @@ protected ExportResult<?> copyIteration(
...
 
-  @Parameterized.Parameters
   public static Iterable<PortabilityAbstractInMemoryDataCopier> data() {
     return Arrays.asList(
-        (new PortabilityAbstractInMemoryDataCopier[] {
-          Mockito.spy(new TestPortabilityInMemoryDataCopier()),
-          Mockito.spy(new TestPortabilityStackInMemoryDataCopier())
+        (new PortabilityAbstractInMemoryDataCopier[]{
+            Mockito.spy(new TestPortabilityInMemoryDataCopier()),
+            Mockito.spy(new TestPortabilityStackInMemoryDataCopier())
```
jzacsh, line 109: optional nit: this isn't due to your PR, but since you're improving this test anyway: could you clarify the names of these two classes? eg: I don't know much about what's going on here, but it's not obvious to me why the additional parameter of a `JobStore` makes this spy a "stack" in-memory copier.

jzacsh (changes requested):
looks good - just an optional tweak if you have the time.

portability-transfer/src/test/java/org/datatransferproject/transfer/copier/PortabilityInMemoryDataCopierTest.java
chaaransen: I could see its not only the additional parameter of `JobStore` but the super classes of these test copiers add to the difference that make one of them the stack copier. `Portability**Stack**InMemoryDataCopier` vs `PortabilityInMemoryDataCopier`

`JobStore` is storing the `Stack<ExportInformation>` which is used in the `copy` method of `Portability**Stack**InMemoryDataCopier` to retrieve the stack for processing, while the "non-stack copier" doesn't require this field.

jzacsh (approved):

PR #1146, opened 2022-08-22
Fix skipping not found videos for Koofr

calmarj:
When video is imported to Koofr and it was not found in CDN, such video should be skipped.
Currently null is returned instead path to video, which breaks creation of successfull ItemImportResult, which requires non-null data.

Short-term solution: return dynamic skipped filepath based on video id.

Potential long term solution: introduce skipped result


wmorland (approved):
LGTM

PR #1147, opened 2022-08-22
Skip photos if album was deleted during transfer

calmarj:
Summary:
Upload File endpoint will return 404 when album was deleted in the middle of transfer, such photos should be skipped.

Introducted KoofrClientIOException in order to get code in importer and have centralized error message formatting

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-koofr/src/main/java/org/datatransferproject/transfer/koofr/common/KoofrClient.java
+++ b/extensions/data-transfer/portability-data-transfer-koofr/src/main/java/org/datatransferproject/transfer/koofr/common/KoofrClient.java
@@ -103,9 +103,7 @@ public class KoofrClient {
       if (code == 404) {
         return false;
       }
-      ResponseBody body = response.body();
-      String bodyAsString = body != null ? body.string(): null;
-      throw new KoofrClientIOException(code, response.message(), bodyAsString);
+      throw new KoofrClientIOException(response);
     }
   }
 
@@ -135,13 +133,7 @@ public class KoofrClient {
       int code = response.code();
       // 409 response code means that the folder already exists
       if ((code < 200 || code > 299) && code != 409) {
-        throw new IOException(
-            "Got error code: "
-                + code
-                + " message: "
-                + response.message()
-                + " body: "
-                + response.body().string());
+       throw new KoofrClientIOException(response);
       }
     }
   }
@@ -173,9 +165,7 @@ public class KoofrClient {
     try (Response response = getResponse(requestBuilder)) {
       int code = response.code();
       if ((code < 200 || code > 299) && code != 409) {
-        ResponseBody responseBody = response.body();
-        String responseBodyAsString = responseBody != null ? responseBody.string(): null;
-        throw new KoofrClientIOException(code, response.message(), responseBodyAsString);
+        throw new KoofrClientIOException(response);
       }
     }
   }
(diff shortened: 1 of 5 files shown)


extensions/data-transfer/portability-data-transfer-koofr/src/main/java/org/datatransferproject/transfer/koofr/common/KoofrClient.java
```diff
@@ -102,13 +103,9 @@ public boolean fileExists(String path) throws IOException, InvalidTokenException
...
         return false;
       }
-      throw new IOException(
-          "Got error code: "
-              + code
-              + " message: "
-              + response.message()
-              + " body: "
-              + response.body().string());
+      ResponseBody body = response.body();
```
jzacsh, line 106: since you're introducing Koofr exception anyway, and koovr client depends on okhttp3, consider just passing `ResponseBody` in as the third param, and let the KoofrClietnIOException do all these work (and then you can delete the other places this is copy/pasted)

extensions/data-transfer/portability-data-transfer-koofr/src/main/java/org/datatransferproject/transfer/koofr/exceptions/KoofrClientIOException.java
```diff
@@ -0,0 +1,20 @@
+package org.datatransferproject.transfer.koofr.exceptions;
+
+import java.io.IOException;
+
+public class KoofrClientIOException extends IOException {
+
+    private final int code;
+
+    public KoofrClientIOException(int code,
```
jzacsh, line 9:
```suggestion
    public KoofrClientIOException(int httpResponseCode,
```

extensions/data-transfer/portability-data-transfer-koofr/src/main/java/org/datatransferproject/transfer/koofr/common/KoofrClient.java
jzacsh, line 106: should we updat ethis method signature to clarify that it throws the more specific exception now?

extensions/data-transfer/portability-data-transfer-koofr/src/main/java/org/datatransferproject/transfer/koofr/common/KoofrClient.java
```diff
@@ -102,13 +103,9 @@ public boolean fileExists(String path) throws IOException, InvalidTokenException
...
+      String bodyAsString = body != null ? body.string(): null;
```
jzacsh, line 107: if the above (updating method signature) is done, we can also update the test for this file: https://github.com/google/data-transfer-project/blob/master/extensions/data-transfer/portability-data-transfer-koofr/src/test/java/org/datatransferproject/transfer/koofr/common/KoofrClientTest.java#L75

jzacsh (changes requested):
looks good! nothing major here - just some small cleanup asks

extensions/data-transfer/portability-data-transfer-koofr/src/main/java/org/datatransferproject/transfer/koofr/common/KoofrClient.java
calmarj: Switched to using okhttp Response for creating KoofrClientIOException

calmarj: getResponse can also throw IOException, so it is probably better to leave it as is.

jzacsh (dismissed):

wmorland (approved):
👍

PR #1148, opened 2022-08-22
Fixed null pointer exception in JobProcessor

denisapolitic:
Optional.of() throws a null pointer exception if readValue is returning null. Optional.ofNullable is returning an empty Optional instead


jzacsh:
unfortunate that there's not already test coverage of this file. if you think it's quick/easy to add coverage (_just_ for your codepath), lease do. otherwise, don't let that block you merging.

(note: I think the bot won't let you merge until you sign the CLA)

jzacsh (approved):

PR #1150, opened 2022-08-30
Junit4 @Rules migration

chaaransen:
-> Migrated @Rule to Junit5 
-> @Before to @BeforeEach


jzacsh:
I think this PR is a dependency of #1152 is that right? Specifically we should be reviewing and submitting the PRs in this order:
1. #1150
2. #1152

if so, then are there any others to be aware of (before, after, inbetween)?

chaaransen:
> I think this PR is a dependency of #1152 is that right? Specifically we should be reviewing and submitting the PRs in this order:
> 
> 1. [Junit4 @Rule migrations to Junit 5 #1150](https://github.com/google/data-transfer-project/pull/1150)
> 2. [Junit4 @Rule migrations to Junit 5 #1152](https://github.com/google/data-transfer-project/pull/1152)
> 
> if so, then are there any others to be aware of (before, after, inbetween)?

Hi @jzacsh, they don't have any dependency between each other, just the pr is broken into two smaller pr parts for easy reviewing,  Updated the title, sorry if it was unclear.

jzacsh (approved):

extensions/data-transfer/portability-data-transfer-backblaze/src/test/java/org/datatransferproject/datatransfer/backblaze/photos/BackblazePhotosImporterTest.java
```diff
@@ -42,119 +43,120 @@
...
+    when(executor.getCachedValue(albumId)).thenReturn(albumName);
+
+    HttpURLConnection connection = mock(HttpURLConnection.class);
+    when(connection.getInputStream()).thenReturn(IOUtils.toInputStream("photo content", "UTF-8"));
+    when(streamProvider.getConnection(photoUrl)).thenReturn(connection);
+
+    when(client.uploadFile(eq("Photo Transfer/albumName/dataId.jpg"), any())).thenReturn(response);
+    when(clientFactory.getOrCreateB2Client(jobId, authData)).thenReturn(client);
+
+    File file = folder.toFile();
```
xokker, line 129: Why are we changing `newFile` to `toFile`?

extensions/data-transfer/portability-data-transfer-backblaze/src/test/java/org/datatransferproject/datatransfer/backblaze/videos/BackblazeVideosImporterTest.java
```diff
@@ -39,104 +40,103 @@
...
+    VideosContainerResource data = mock(VideosContainerResource.class);
+    when(data.getVideos()).thenReturn(videos);
+
+    when(executor.getCachedValue(albumId)).thenReturn(albumName);
+
+    HttpURLConnection connection = mock(HttpURLConnection.class);
+    when(connection.getInputStream()).thenReturn(IOUtils.toInputStream("video content", "UTF-8"));
+    when(streamProvider.getConnection(videoUrl)).thenReturn(connection);
+
+    when(dataStore.getTempFileFromInputStream(any(), any(), any())).thenReturn(folder.toFile());
```
xokker, line 126: here as well

extensions/data-transfer/portability-data-transfer-backblaze/src/test/java/org/datatransferproject/datatransfer/backblaze/photos/BackblazePhotosImporterTest.java
chaaransen: The previous implementation was @Rule based in JUnit4 , the .newFile() previously as we see is invoked on TemporaryFolder object which is not supported anymore in JUnit5, instead its converted to a path which is annotated with @TempDir (Junit5) which supplies a temp dir location (cleanups are maintained by JUnit5), .toFile() method returns file object representing the file object representing the default path of the temp dir defined.

extensions/data-transfer/portability-data-transfer-backblaze/src/test/java/org/datatransferproject/datatransfer/backblaze/videos/BackblazeVideosImporterTest.java
chaaransen: same as above

extensions/data-transfer/portability-data-transfer-backblaze/src/test/java/org/datatransferproject/datatransfer/backblaze/photos/BackblazePhotosImporterTest.java
xokker: Oh I see! I totally missed the fact we also changed the field type from `TemporaryFolder` to `Path`. Then I got confused on converting a _directory_ path to a file (as opposed to creating a _file_ in a directory). Looks like in our case both approaches work because the only time we use it in the test [we take its size and pass the file to a mock](https://github.com/google/data-transfer-project/blob/9cc43140a6fea2190dd66cae4235700e8a622f33/extensions/data-transfer/portability-data-transfer-backblaze/src/main/java/org/datatransferproject/datatransfer/backblaze/photos/BackblazePhotosImporter.java#L110-L115).

xokker (approved):

PR #1136, opened 2022-08-09
Add automatic media job type compatibility

incjo:
makes progress against #1065 by building on top of https://github.com/google/data-transfer-project/pull/1062 by further expanding support for automatically handling media jobs.

It allows not writing dedicated media adapters in cases where the existing photo/video functionality is satisfactory. It also provides automatic interoperability with destinations that only provide a Media adapter.

Main contributions:
- replaced `MediaToPhotoConversionImporter` and `MediaToPhotoConversionExporter` with generic versions that can support conversion between any two adapters: `AnyToAnyImporter` and `AnyToAnyExporter`
- added special case adapters `MediaExporterDecorator` and `MediaImporterDecorator` which combine existing photo and video adapters to seamlessly cover the media job type.
- added a compatibility layer `TransferCompatibilityProvider` (perhaps rename to `AdapterCompatibilityProvider`?) which  attempts to build a substitute adapter upon not finding a real one. E.g. when a media job can't find a media importer, it will combine existing photo and video importers to perform the job.

Issue #1065, opened 2022-04-21
automate DTP's injection of auto-converters so extensions don't have to do this


jzacsh:
for context, see the TODO on lines of `extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/datatransferproject/transfer/microsoft/MicrosoftTransferExtension.java` in https://github.com/google/data-transfer-project/pull/1062/files inlined below:

```java
  // TODO(jzacsh,wmorland) don't keep adding here - just have the converters invoked automatically
  // when Media isn't supported on one or the other side of this equation; this is just a WIP
  // prototype to show the concept of converters at play.
  private static final ImmutableList<String> SUPPORTED_IMPORT_SERVICES =
      ImmutableList.of(CALENDAR, CONTACTS, PHOTOS, MEDIA);
```

PR #1062, opened 2022-04-18
WIP proposal: prototype of partially automated Media integration

jzacsh:
this PR is just a demonstration[^1] Will and I will use in our upcoming doc.

[^1]: and eventually possibly cleanup and commit among other changes

PR #1151, opened 2022-08-30
Add automatic media job type compatibility and introduce the DataVertical enum

incjo:
makes progress against #1065 by building on top of https://github.com/google/data-transfer-project/pull/1062 by further expanding support for automatically handling media jobs.

It allows not writing dedicated media adapters in cases where the existing photo/video functionality is satisfactory. It also provides automatic interoperability with destinations that only provide a Media adapter.

Main contributions:
- replaced `MediaToPhotoConversionImporter` and `MediaToPhotoConversionExporter` with generic versions that can support conversion between any two adapters: `AnyToAnyImporter` and `AnyToAnyExporter`
- added special case adapters `MediaExporterDecorator` and `MediaImporterDecorator` which combine existing photo and video adapters to seamlessly cover the media job type.
- added a compatibility layer `TransferCompatibilityProvider` which attempts to build a substitute adapter upon not finding a real one. E.g. when a media job can't find a media importer, it will combine existing photo and video importers to perform the job.
- added a `DataVertical` enum which replaces the hardcoded type strings like `"PHOTOS"`.

Review the commit separately as the `DataVertical` commit is very monotonic.

----

note: this PR has a lot in it due to https://github.com/google/data-transfer-project/pull/1136#issuecomment-1231640367

### DIFF
--- a/portability-types-common/src/main/java/org/datatransferproject/types/common/models/DataVertical.java
+++ b/portability-types-common/src/main/java/org/datatransferproject/types/common/models/DataVertical.java
@@ -1,0 +1,49 @@
+package org.datatransferproject.types.common.models;
+
+import com.fasterxml.jackson.annotation.JsonValue;
+import java.util.HashMap;
+import java.util.Map;
+
+
+/**
+ * Represents the data type of a transfer job
+ */
+public enum DataVertical {
+  // TODO consider if we can reference data models directly rather than enums; for now beware
+  //  that each item in this enum is represented by one and only one common data model.
+  BLOBS("BLOBS"),
+  CALENDAR("CALENDAR"),
+  CONTACTS("CONTACTS"),
+  MAIL("MAIL"),
+  MEDIA("MEDIA"),
+  NOTES("NOTES"),
+  OFFLINE_DATA("OFFLINE-DATA"),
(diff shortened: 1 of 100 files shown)


portability-types-common/src/main/java/org/datatransferproject/types/common/models/DataVertical.java
```diff
@@ -0,0 +1,46 @@
...
+
+import com.fasterxml.jackson.annotation.JsonValue;
+import java.util.HashMap;
+import java.util.Map;
+
+
+/**
+ * Represents the data type of a transfer job
+ */
+public enum DataVertical {
```
jzacsh, line 11:
```suggestion
// TODO consider if we can reference data models directly rather than enums; for now beware that each item in this enum is represented by one and only one common data model.
public enum DataVertical {
```

jzacsh (changes requested):
up to you, but this is easier to follow as two commits the way  you have them, so when you click the merge button in github, consider unticking the "squash" button so that it keeps the commits separated (your commit to automate stuff; your commit to add an enum).

jzacsh (approved):

wmorland (approved):
👍

PR #1152, opened 2022-08-31
Junit4 @Rule migrations to Junit 5

chaaransen:
Rules are deprecated in Junit5 hence migrated existing @rule to junit 5 equivalents.

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-backblaze/src/test/java/org/datatransferproject/datatransfer/backblaze/photos/BackblazePhotosImporterTest.java
+++ b/extensions/data-transfer/portability-data-transfer-backblaze/src/test/java/org/datatransferproject/datatransfer/backblaze/photos/BackblazePhotosImporterTest.java
@@ -26,6 +26,7 @@ import static org.mockito.Mockito.when;
 
 import java.io.File;
 import java.net.HttpURLConnection;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.UUID;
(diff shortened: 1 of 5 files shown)


extensions/data-transfer/portability-data-transfer-backblaze/src/test/java/org/datatransferproject/datatransfer/backblaze/photos/BackblazePhotosImporterTest.java
```diff
@@ -42,119 +43,120 @@
 import org.datatransferproject.types.common.models.photos.PhotoModel;
 import org.datatransferproject.types.common.models.photos.PhotosContainerResource;
 import org.datatransferproject.types.transfer.auth.TokenSecretAuthData;
-import org.junit.Before;
```
jzacsh, line 45: would you mind fixing the whitespace so it's not all reformatted? (or if we want to do that, then do it in a separate PR)

That is: I see it switches in this file from 2 char indentation to 4 char, but this makes your PR's real changes (junit migration) more difficult to follow

jzacsh (changes requested):

extensions/data-transfer/portability-data-transfer-backblaze/src/test/java/org/datatransferproject/datatransfer/backblaze/photos/BackblazePhotosImporterTest.java
jzacsh: > (or if we want to do that, then do it in a separate PR)

also on this^ front: Will and I have discussed automating style choices for the codebase, so it's probably not worth bothering here as it's on our future-hackathon-TODO list to tackle more holistically

chaaransen: > would you mind fixing the whitespace so it's not all reformatted? (or if we want to do that, then do it in a separate PR)
> 
> That is: I see it switches in this file from 2 char indentation to 4 char, but this makes your PR's real changes (junit migration) more difficult to follow

Sure will keep that in mind, Please let me know if we still find it difficult to follow after toggling this "Hide whitespace" checkbox on diff view from the settings icon, In that case i'll fix it to the previous formatting and reapply all the changes.

Screenshot 2022-08-31 at 4 12 04 pm

PR #1155, opened 2022-09-01
Adding future-proof error message matcher for upload hash validation

MewX:


extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosInterface.java
```diff
@@ -260,9 +260,13 @@ <T> T makePostRequest(String url, Optional<Map<String, String>> parameters,
    * Note that making this a separate method to avoid polluting throw lists.
    */
   private void maybeRethrowAsUploadError(HttpResponseException e) throws UploadErrorException {
```
jzacsh, line 262: optional nit: would be helpful to pull these strings out to consts and document them, or give them self-documenting names like `GOOG_MISMATCH_API_ERROR_MSG` and `GOOG_MISMATCH_API_ERROR_OLD_MSG`

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosInterface.java
```diff
@@ -260,9 +260,13 @@ <T> T makePostRequest(String url, Optional<Map<String, String>> parameters,
...
-    if (e.getStatusCode() == 400 && e.getContent()
-        .contains("Checksum from header does not match received payload content.")) {
-      throw new UploadErrorException(ERROR_HASH_MISMATCH, e);
+    if (e.getStatusCode() == 400) {
```
jzacsh, line 263: optional nit: your choice, but maybe this is clearer as a single expression? `e.getStatusCode() == 400 && (e.getContent().contains(GOOG_...) || e.getContent().contains(GOOG_...))`

I think this makes your comment about delegation a bit clearer (really _everything_ will be delegated to be handled eleswhere - even the non-400s).

jzacsh (changes requested):
could you add test coverage for this new edge case in to your existing tests? `GooglePhotosImporterTest.java`

jzacsh (changes requested):
could you add test coverage for this new edge case in to your existing tests? `GooglePhotosImporterTest.java`

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosInterface.java
MewX: Thanks! Updated.

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/photos/GooglePhotosInterface.java
MewX: Thanks! I feel like grouping those error messages in one line doesn't make much difference. Actually, how about I update the coment to explicitly mention other 400 errors and non-400 errors will be delegated?

MewX:
> could you add test coverage for this new edge case in to your existing tests? `GooglePhotosImporterTest.java`

Currently `GooglePhotosImporterTest` only receives the mocked `UploadErrorException` signal, which has been tested.

This new error string comes from network request, and should be in `GooglePhotosInterfaceTest`. However, it doesn't exist yet. Due to time limit, can I skip creating this test from scratch 😂 Thanks!

jzacsh:
> > could you add test coverage for this new edge case in to your existing tests? `GooglePhotosImporterTest.java`
> 
> Currently `GooglePhotosImporterTest` only receives the mocked `UploadErrorException` signal, which has been tested.
> 
> This new error string comes from network request, and should be in `GooglePhotosInterfaceTest`. However, it doesn't exist yet. Due to time limit, can I skip creating this test from scratch joy Thanks!

ah that's unfortunate - I've filed #1158 - thanks for pointing this out!

PR #1166, opened 2022-09-13
Assert migration to Junit5

chaaransen:
Migrated Junit4 Assert to Junit5 Assertions library


xokker (approved):
Thank you @chaaransen for migrating our code base to the modern version of junit!

PR #1167, opened 2022-09-22
Fix Google Drive Importer adapter

incjo:
This is a straightforward bug in the importer code.

When parent folder id is specified it should be used instead of defaulting to `ROOT_FOLDER_ID`


ayushagarwalfb (approved):

ayushagarwalfb (approved):

xokker:
hey @seehamrun, could you please take a look when you have time? It's a straightforward fix, but it does touch Google's importer.

PR #1169, opened 2022-10-04
check entire Throwable in RetryMapping::matchesThrowable

alexeyqu-fb:
This becomes important when the original exception (e.g. rate limit exceeded) is shadowed with a wrapping exception somewhere, so the `toString` doesn't contain the important information anymore.

Using `ExceptionUtils.getStackTrace` from `lang3.exception` is one of the ways to look into the stack trace -- however, I am open to suggestions.

cc @xokker


portability-types-transfer/src/main/java/org/datatransferproject/types/transfer/retry/RetryMapping.java
```diff
@@ -48,13 +49,18 @@ public RetryStrategy getStrategy() {
   }
 
   public boolean matchesThrowable(Throwable throwable) {
-    // TODO: examine entire throwable, not just toString
     String input = throwable.toString();
     for (String regex : regexes) {
       if (input.matches(regex)) {
         return true;
       }
     }
```
jzacsh, line 57: leaving the previous loop seems like a micro-optimization; should we just delete it for now and add something like this (or something else) in later, if we do performance testing and find this area problematic?

jzacsh (changes requested):
I'm not very familiar with this class, but maybe it'd increase confidence if we started adding unit test coverage? Maybe not to this file, but to its one and only user (as far as I can tell): `RetryStartegyLibrary` itself? eg: maybe then it'd be easy (for future us) to see this PR was just adding a new test case?

lisad:
Closing stale pull request - it's small, so if it's even still needed please re-build

PR #1192, opened 2022-12-02
Refactor PortabilityStackInMemoryDataCopier to improve clarity

incjo:
- Reduce code duplication by extracting duplicate code into a method
- Replace usages of Stack (it's been deprecated for ages and calls synchronize with every call) with Deque.


lisad:
sorry we missed the useful window on this PR... we're trying to catch up on weekly maintainers' meetings

PR #1197, opened 2022-12-12
Updated SocialActivityAttachment and SocialActivityModel classes to implement ImportableItem class

saratchebrolu:
Updated SocialActivityAttachment and SocialActivityModel classes to implement ImportableItem class

PR #1198, opened 2022-12-12
Updated Daybook posts importer to log bytes

saratchebrolu:
Updated Daybook posts importer to log bytes

NOTE: Dependent on https://github.com/google/data-transfer-project/pull/1197

Unlike in the case of other services, we don't seem to upload images in the importer. Instead, image URLs are being included in the form data, so, the size of the same is being logged.

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-daybook/src/main/java/org/datatransferproject/transfer/daybook/social/DaybookPostsImporter.java
+++ b/extensions/data-transfer/portability-data-transfer-daybook/src/main/java/org/datatransferproject/transfer/daybook/social/DaybookPostsImporter.java
@@ -23,6 +23,7 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.UUID;
+import java.util.concurrent.atomic.LongAdder;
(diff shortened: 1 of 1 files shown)


extensions/data-transfer/portability-data-transfer-daybook/src/main/java/org/datatransferproject/transfer/daybook/social/DaybookPostsImporter.java
```diff
@@ -155,6 +164,7 @@ private String insertActivity(SocialActivityModel activity, TokensAndUrlAuthData
 
     FormBody formBody = builder.build();
     requestBuilder.post(formBody);
+    size = formBody.contentLength();
```
alexeyqu-fb, line 167: Ok but are we sure we want to log the payload size here? 
E.g. what if the Daybook unpacks it / some fields in the form are stored more efficiently?

I'd say let's merge this and then check the numbers for real transfers..

alexeyqu-fb (approved):
LGTM, provided we want to log the `formBody` size here.

PR #1206, opened 2023-01-10
Separate getErrors from copy, so it isn't skipped when copy throws.

carrot-lion:
This should fix an issue where the errors from the idempotent executor weren't being added to the datastore when the job completed.

### DIFF
--- a/portability-transfer/src/main/java/org/datatransferproject/transfer/JobProcessor.java
+++ b/portability-transfer/src/main/java/org/datatransferproject/transfer/JobProcessor.java
@@ -150,10 +150,11 @@ class JobProcessor {
     } finally {
       final Collection<ErrorDetail> errors = copier.getErrors(jobId);
       final int numErrors = errors.size();
-      monitor.debug(
-          () -> format("Finished copy for jobId: %s with %d error(s).", jobId, numErrors));
+      // success is set to true above if copy returned without throwing
       success &= errors.isEmpty();
-      monitor.debug(() -> "Finished processing jobId: " + jobId, EventCode.WORKER_JOB_FINISHED);
+      monitor.debug(
+          () -> format("Finished processing jobId: %s with %d error(s).", jobId, numErrors),
+          EventCode.WORKER_JOB_FINISHED);
       addErrorsAndMarkJobFinished(jobId, success, errors);
       hooks.jobFinished(jobId, success);
       dtpInternalMetricRecorder.finishedJob(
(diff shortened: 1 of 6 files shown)


portability-transfer/src/main/java/org/datatransferproject/transfer/JobProcessor.java
```diff
@@ -153,6 +148,11 @@ void processJob() {
     } catch (IOException | CopyException | RuntimeException e) {
       monitor.severe(() -> "Error processing jobId: " + jobId, e, EventCode.WORKER_JOB_ERRORED);
     } finally {
+      final Collection<ErrorDetail> errors = copier.getErrors(jobId);
+      final int numErrors = errors.size();
+      monitor.debug(
+          () -> format("Finished copy for jobId: %s with %d error(s).", jobId, numErrors));
+      success &= errors.isEmpty();
```
jzacsh, line 155:
```suggestion
      boolean success = errors.isEmpty();
```

Seems we can simplify this by just defining `success` straight away right here, no? The way the code for `success` is laid out now makes me think there's _definitely_ some `success = ...` assignment in one of the catch blocks, but that turns out not to be true, right? So no need to have the variable's life cycle start so early, afterall?

portability-transfer/src/main/java/org/datatransferproject/transfer/JobProcessor.java
```diff
@@ -153,6 +148,11 @@ void processJob() {
...
       monitor.debug(() -> "Finished processing jobId: " + jobId, EventCode.WORKER_JOB_FINISHED);
```
jzacsh, line 156: should we get rid of this existing debug line? or we wanted to keep two of them on purpose?

portability-transfer/src/test/java/org/datatransferproject/transfer/JobProcessorTest.java
```diff
@@ -0,0 +1,91 @@
+/*
+ * Copyright 2020 The Data Transfer Project Authors.
```
jzacsh, line 2:
```suggestion
 * Copyright 2022 The Data Transfer Project Authors.
```

jzacsh (changes requested):
nice, lgtm - just tiny comments

portability-transfer/src/main/java/org/datatransferproject/transfer/JobProcessor.java
carrot-lion: `success` is initially defined as false and set to true at the bottom of the try block when copy returns without throwing.

portability-transfer/src/main/java/org/datatransferproject/transfer/JobProcessor.java
carrot-lion: Combined debug messages.

jzacsh (approved):

jzacsh (dismissed):

jzacsh (approved):

PR #1216, opened 2023-02-24
Improvements to backblaze data transfer

alexlipa91:
- Add support for MEDIA transfer
- Add support for photos and videos in the same folder
- Use media name instead of id
- Store media creation time as S3 object metadata (it is not possible to set the creation date in S3 since it is automatically set by the server)


lisad:
Closing stale pull request. It would need to have conflicts fixed if it needs to be re-opened. Pinging @ameya9 because this may have originally been a contribution from a colleague of yours.

PR #1217, opened 2023-02-24
Store entire Throwable in ErrorDetail class

alexeyqu-fb:
There is a place at [CallableImporter](https://github.com/google/data-transfer-project/blob/889f1fa92fc54046b42a9448b53d9de0df84912c/portability-transfer/src/main/java/org/datatransferproject/transfer/CallableImporter.java#L77) where we leak the exception trace into raw logs by wrapping the `ErrorDetail` contents in an `IOException`.

```
throw new IOException(
  "Problem with importer, forcing a retry, "
    + "first error: "
    + (errors.iterator().hasNext() ? errors.iterator().next().exception() : "none"));
```

This is bad for 3 reasons:
- The actual root cause of the problem is masked with a generic `IOException`;
- The full trace of the original exception is leaked into the exception message, bloating the logs;
- If the original exception was itself a wrapper, we'd lose the underlying cause completely.

As a first step towards fixing that, we want to modify `ErrorDetail` to store the `Throwable` instance instead of its trace in the `String` format.

### DIFF
--- a/extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleCloudIdempotentImportExecutor.java
+++ b/extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleCloudIdempotentImportExecutor.java
@@ -88,7 +88,7 @@ public class GoogleCloudIdempotentImportExecutor implements IdempotentImportExec
           ErrorDetail.builder()
               .setId(idempotentId)
               .setTitle(itemName)
-              .setException(Throwables.getStackTraceAsString(e))
+              .setThrowable(e)
               .build();
       addError(idempotentId, errorDetail);
       monitor.severe(() -> jobIdPrefix + "Problem with importing item: " + errorDetail);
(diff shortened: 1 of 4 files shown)


extensions/cloud/portability-cloud-google/src/main/java/org/datatransferproject/cloud/google/GoogleCloudIdempotentImportExecutor.java
```diff
@@ -232,6 +232,8 @@ private Entity createErrorEntity(String idempotentId, ErrorDetail error) throws
   @VisibleForTesting
   Entity createErrorEntity(String idempotentId, UUID jobId, ErrorDetail error)
       throws IOException {
+    System.out.println("AQAQAQAQ");
```
alexeyqu-fb, line 235: TODO delete this

alexeyqu-fb:
Hi @seehamrun and thanks for the reply!

I am happy to make the change with marking the field as non-indexable.
This does not block anything specifically, but does impact the error tracking across the codebase (see my original comment for examples and reason).

Do you think it needs to be tested on Google side specifically?

alexeyqu-fb:
Hm, maybe the best approach here is to wait for the Jackson's bugfix to serialize `Throwable`s properly.
https://github.com/FasterXML/jackson-databind/issues/1986

...which might be coming quite soon, see [this discussion](https://github.com/openrewrite/rewrite/pull/3229#discussion_r1190834198).

PR #1221, opened 2023-03-15
Add specification of callback for google/imgur

frankduncan:
When setting up the demo, and enabling oauth on the systems, they required the redirect being sent from DTP to match the one in their system, which was <hostname>/callback/<system>


lisad:
Closing stale PR - looks like the only issue was the CLA

PR #1222, opened 2023-03-24
Throw invalid token exception for video import

srujithm:
Summary: Capture and throw invalid token exception for video import. This is a temporary check as exception is not properly captured and wrapped into UploadMediaItemResponse error field for videos.

Test Plan: ./gradlew test


alexeyqu-fb (approved):
LGTM, for some reason the `uploadMediaItem` handles the invalid_grant exception in a different way than `batchCreateMediaItems` (both methods are part of the Google Photos Java Library we use).

PR #1225, opened 2023-04-04
Add getRetryingIdempotentImportExecutor

kateyeo:


portability-spi-transfer/src/main/java/org/datatransferproject/spi/transfer/idempotentexecutor/InMemoryIdempotentImportExecutorExtension.java
```diff
@@ -14,6 +14,11 @@ public IdempotentImportExecutor getIdempotentImportExecutor(ExtensionContext ext
     return new InMemoryIdempotentImportExecutor(extensionContext.getMonitor());
   }
 
+  @Override
```
seehamrun, line 17: it looks liek the build is failing because this something is missing this implementation, i think we need to mark this as @default

Moo331: I think so too.

kateyeo: I realized I forgot to add the getRetryingIdempotentImportExecutor to IdempotentImportExecutorExtension and it fixed the build issues, should I still mark this @default?

seehamrun: Ah I see -- so i think you need a default implementation in IdempotentImportExecutorExtension.java -- that way the specific implementations (GoogleCloudIdempotentImportExecutor and InMemoryIdempotentImportExecutor and others that might not be in github) don't need to add an override for that implementation.

portability-spi-transfer/src/main/java/org/datatransferproject/spi/transfer/idempotentexecutor/IdempotentImportExecutorExtension.java
```diff
@@ -6,4 +6,5 @@
 
 public interface IdempotentImportExecutorExtension extends BootExtension {
   IdempotentImportExecutor getIdempotentImportExecutor(ExtensionContext extensionContext);
+  IdempotentImportExecutor getRetryingIdempotentImportExecutor(ExtensionContext extensionContext);
```
seehamrun, line 9: you need the default implementation here (it can either return null or call getIdempotentImportExecutor()..) then you can remove the overrides in the specific implementions

kateyeo: Ah ok, that makes sense, I'll change that

seehamrun (approved):

Issue #1105, opened 2022-06-30
unit tests not hermetic: actual network dns lookup appears to happen (and maybe more) for imgur


jzacsh:
unlike #1104 I think this set of failures isn't actually my fault but is a real bug. the problem is there's a lack of test double being put in place so I'm seeing `java.net.UnknownHostException: i.imgur.com` errors.

### step 1: prime gradle cache and prove things are passing

`gradle test` is passing with your network turned on

### step 2: try offline

`gradle test` with your network off (eg: you're traveling) ~or force it via `gradle --offline test`~ (edit: nope, that applies to gradle's fetching build resources, not cutting off its subprocesses) and you'll get:

```
$ gradle test
> Task :extensions:data-transfer:portability-data-transfer-imgur:test FAILED

org.datatransferproject.transfer.imgur.photos.ImgurPhotoExporterTest > testAlbumPhotosExport FAILED
    java.net.UnknownHostException at ImgurPhotoExporterTest.java:148

org.datatransferproject.transfer.imgur.photos.ImgurPhotoExporterTest > testPagination FAILED
    java.net.UnknownHostException at ImgurPhotoExporterTest.java:213

org.datatransferproject.transfer.imgur.photos.ImgurPhotoExporterTest > testAlbumAndNonAlbumPhotoExport FAILED
    java.net.UnknownHostException at ImgurPhotoExporterTest.java:169

org.datatransferproject.transfer.imgur.photos.ImgurPhotoExporterTest > testNonAlbumPhotoExport FAILED
    java.net.UnknownHostException at ImgurPhotoExporterTest.java:196

5 tests completed, 4 failed
```

and the locally-generated gradle test webpage shows failures for `Class org.datatransferproject.transfer.imgur.photos.ImgurPhotoExporterTest` with `i.imgur.com` unknown; specifically:

> ImgurPhotoExporterTest. testAlbumAndNonAlbumPhotoExport
> ImgurPhotoExporterTest. testAlbumPhotosExport
> ImgurPhotoExporterTest. testNonAlbumPhotoExport
> ImgurPhotoExporterTest. testPagination

and here's sample stacktrace from that first failing test in the list:

```
testAlbumAndNonAlbumPhotoExport

java.net.UnknownHostException: i.imgur.com
	at java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:184)
	at java.net.SocksSocketImpl.connect(SocksSocketImpl.java:392)
[...]

PR #1240, opened 2023-05-02
Add mocks for InputStream so that Imgur tests do not fail offline

kateyeo:
Fix for https://github.com/google/data-transfer-project/issues/1105

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosExporter.java
+++ b/extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosExporter.java
@@ -103,8 +103,6 @@ public class ImgurPhotosExporter
     ALL_PHOTOS_URL_TEMPLATE = baseUrl + "/account/me/images/%s?perPage=" + RESULTS_PER_PAGE;
   }
 
-
-
   /**
    * Exports albums and photos. Gets albums first, then photos which are contained in albums and
    * non-album photos
--- a/extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosExporter.java
+++ b/extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosExporter.java
@@ -17,6 +17,7 @@
 package org.datatransferproject.datatransfer.imgur.photos;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
@@ -306,7 +307,8 @@ public class ImgurPhotosExporter
     }
   }
 
-  private InputStream getImageAsStream(String imageUrl) throws IOException {
+  @VisibleForTesting
+  public InputStream getImageAsStream(String imageUrl) throws IOException {
     URL url = new URL(imageUrl);
     HttpURLConnection conn = (HttpURLConnection) url.openConnection();
     conn.connect();
--- a/extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosExporter.java
+++ b/extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosExporter.java
@@ -95,13 +95,11 @@ public class ImgurPhotosExporter
       ObjectMapper objectMapper,
       TemporaryPerJobDataStore jobStore,
       String baseUrl,
-      // URL url,
       UrlFactory urlFactory) {
     this.client = client;
     this.objectMapper = objectMapper;
     this.monitor = monitor;
     this.jobStore = jobStore;
-    // this.url = url;
     this.urlFactory = urlFactory;
     ALBUM_PHOTOS_URL_TEMPLATE = baseUrl + "/album/%s/images";
     ALBUMS_URL_TEMPLATE = baseUrl + "/account/me/albums/%s?perPage=" + RESULTS_PER_PAGE;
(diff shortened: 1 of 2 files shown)


extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosExporter.java
```diff
@@ -67,6 +69,9 @@
   private final Monitor monitor;
   private final TemporaryPerJobDataStore jobStore;
 
+  // private URL url;
+  private UrlFactory urlFactory;
+
   public ImgurPhotosExporter(
```
jzacsh, line 75: here you can just use constructor overloading (just call `this(params, here, all, passed, through,as, is, (url) -> new URL(url))`)

extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosExporter.java
```diff
@@ -77,6 +82,27 @@ public ImgurPhotosExporter(
     this.objectMapper = objectMapper;
     this.monitor = monitor;
     this.jobStore = jobStore;
+    this.urlFactory = (urlString)->{return new URL(urlString);};
```
jzacsh, line 85: I _think_ you can simplify this lambda a tiny bit

(and then also the diff I'm showing is assuming you go with the other comment i'm leaving you above)

```suggestion
    (url)-> new URL(url),
```

extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosExporter.java
```diff
@@ -307,9 +333,20 @@ private List<Map<String, Object>> requestData(TokensAndUrlAuthData authData, Str
...
     HttpURLConnection conn = (HttpURLConnection) url.openConnection();
     conn.connect();
     return conn.getInputStream();
   }
+
+  // private static URL from(String imageUrl) throws IOException{
+  //   return new URL(imageUrl);
+  // }
+
+  public interface UrlFactory{
```
jzacsh, line 349: I don't _think_ you need this, btw. I think you can just declare the type of your `urlFactory` param to be `Function<String, URL> urlFactory` in your constructor.

extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosExporter.java
kateyeo: changed that, thanks!

extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosExporter.java
kateyeo: changed, thanks!

extensions/data-transfer/portability-data-transfer-imgur/src/main/java/org/datatransferproject/datatransfer/imgur/photos/ImgurPhotosExporter.java
kateyeo: changed from interface to declaring the param type, thanks

jzacsh (changes requested):

jzacsh:
LGTM!

extensions/data-transfer/portability-data-transfer-imgur/src/test/java/org/datatransferproject/transfer/imgur/photos/ImgurPhotoExporterTest.java
```diff
@@ -100,7 +104,16 @@ public void setUp() throws IOException {
     server = new MockWebServer();
     server.start();
     exporter =
-        new ImgurPhotosExporter(monitor, client, mapper, jobStore, server.url("").toString());
+        new ImgurPhotosExporter(monitor, client, mapper, jobStore, server.url("").toString(),
+            (urlString) -> {
+              URL u = mock(URL.class);
```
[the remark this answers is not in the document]
kateyeo: I can't call when(url.openConnection) on a real URL so I think I need to mock both the URL and HttpURLConnection to completely block the network calls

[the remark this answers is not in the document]
jzacsh: chatted offline and figured out we needed a tiny bit of spying instead of a full mock.

extensions/data-transfer/portability-data-transfer-imgur/src/test/java/org/datatransferproject/transfer/imgur/photos/ImgurPhotoExporterTest.java
```diff
@@ -100,7 +108,23 @@ public void setUp() throws IOException {
     server = new MockWebServer();
     server.start();
     exporter =
-        new ImgurPhotosExporter(monitor, client, mapper, jobStore, server.url("").toString());
+        new ImgurPhotosExporter(monitor, client, mapper, jobStore, server.url("").toString(),
+            (urlString) -> {
+              URL u = null;
+              try {
+                u = spy(new URL(urlString));
+              } catch (MalformedURLException e) {
```
jzacsh, line 116: (nit; up to you if you feel like tackling this right now)

i see why the try/catch in the real code, but probably not necessary in the test?

jzacsh (approved):

extensions/data-transfer/portability-data-transfer-imgur/src/test/java/org/datatransferproject/transfer/imgur/photos/ImgurPhotoExporterTest.java
kateyeo: The build failed without the try/catch, even with the call being in the test

PR #1247, opened 2023-06-01
Change idempotent id for Social Activity Model

calmarj:
- Change idempotent id for Social Activity Model to id (to be consistent with other models)
- Add final to all models in social folder
- Fix link in documentation


alexeyqu-fb (approved):
LGTM

PR #1250, opened 2023-06-16
Create ActivityModel.java

justflatley:
Adding initial user activity model


chriswryly:
I wanted to chime in with my support for this activity model, as something that will prove useful in a range of potential contexts to describe non-content-creating (perhaps "passive"?) activity as it is customarily collected and stored by a service, and routinely made available for users to download through data export interfaces.

One substantive suggestion I'd like to make is the addition of a field of metadata reflecting the URI of the content at issue at the time of the activity. For example, the current photos data model has "fetchableUrl" as a specific property (though that may serve other purposes). Perhaps this is implicit in the "action" string? I think there might be value in breaking it out, if so, or adding it if it's not already included.

lisad:
Closing stale PR - would be lovely to have but nobody's been working on this in their deployments in the last 3 years

PR #1269, opened 2023-09-20
refactor apple media exceptions

yy9669:


extensions/data-transfer/portability-data-transfer-apple/src/main/java/org/datatransferproject/datatransfer/apple/photos/AppleMediaInterface.java
```diff
@@ -267,29 +272,31 @@ private String sendPostRequest(@NotNull String url, @NotNull final byte[] reques
...
+      convertAndThrowException(e, con);
     } finally {
       con.disconnect();
     }
     return responseString;
   }
 
-  private void convertAndThrowException(@NotNull final AppleHttpException e)
-      throws CopyExceptionWithFailureReason {
+  private void convertAndThrowException(@NotNull final IOException e, @NotNull final HttpURLConnection con)
```
seehamrun, line 282: Q: Should any of these cases not throw errors and just be skipped instead? I think becuase they are CopyExceptions they will cause the job to automatically fail, but if you convert them into an IOException or a different exception they will get retried using the executor (and within Google we have the skippable errors in place so we can tag the exception to just skip instead of retry)

yy9669: Thank, that makes sense, I have updated some errors to be IOException to make it retryable and skippable.

portability-spi-transfer/src/main/java/org/datatransferproject/spi/transfer/types/FailureReasons.java
```diff
@@ -8,8 +8,11 @@ public enum FailureReasons {
   SESSION_INVALIDATED("SESSION_INVALIDATED"),
   UNCONFIRMED_USER("UNCONFIRMED_USER"),
   UPLOAD_ERROR("UPLOAD_ERROR"),
-  USER_CHECKPOINTED("USER_CHECKPOINTED");
-
+  USER_CHECKPOINTED("USER_CHECKPOINTED"),
+  SERVICE_UNAVAILABLE("SERVICE_UNAVAILABLE"),
```
seehamrun, line 12: Are all these new ones still used? From a quick glance seems like only the DESTINATION_NOT_FOUND one is?

seehamrun (dismissed):
just a quick question -- will merge once thats cleared up

portability-spi-transfer/src/main/java/org/datatransferproject/spi/transfer/types/FailureReasons.java
yy9669: Updated ! One further question, do we need to add these IOExceptions into some config files ? 
All IOExceptions thrown from apple extension are expected to be retry and skip at maximum retry.

seehamrun (approved):

PR #1273, opened 2023-09-28
Add explicit type to GoogleTrack

gauss0818:


extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/music/GoogleMusicExporter.java
```diff
@@ -280,8 +280,8 @@ private MusicPlaylistItem convertPlaylistItem(
                 release.getIcpn(),
                 release.getTitle(),
                 createMusicGroups(release.getArtists())),
-            createMusicGroups(track.getArtists()), false),
-        // TODO(critical WIP-feature step): Add explicit flag to GooglePlaylistItem
+            createMusicGroups(track.getArtists()),
+            "EXPLICIT_TYPE_EXPLICIT".equals(track.getExplicitType())),
```
jzacsh, line 284: Oh I didn't quite realize these would be arbitrary strings. Should we  normalize this in the shared model somehow? (eg: each adapter maps its own types into the shared type?)

jzacsh (changes requested):

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/music/GoogleMusicExporter.java
gauss0818: The explicit type is an enum from our side. I'm not sure if other platforms has the same definition. Apple Music is okay with boolean isExplicit in music data model.

gauss0818: I think we can discuss the topic with Apple Music in our next meetup. Open to update the music data model.

jzacsh (approved):

PR #1292, opened 2023-11-09
fix: wrap invalid token, destination full exceptions from DriveImporter

alexeyqu-fb:
This diff enabled `DriveImporter` to wrap the API responses into `CopyExceptionWithFailureReason` instances, handling 2 cases:
1) similar to other importers, `invalid_grant` API response is wrapped into `InvalidTokenException`;
2) similar to other importers, `storage quota has been exceeded` error message is wrapped into `DestinationMemoryFullException`.

Test:
`./gradlew test`


wmorland (approved):
LGTM.

PR #1296, opened 2023-11-15
fix video count in import result of AppleMediaImporter

yy9669:


extensions/data-transfer/portability-data-transfer-apple/src/main/java/org/datatransferproject/datatransfer/apple/photos/AppleMediaImporter.java
```diff
@@ -97,7 +97,7 @@ public ImportResult importItem(
                 importResultMap.getOrDefault(ApplePhotosConstants.COUNT_KEY, 0L).intValue())
```
jzacsh, line 97:
```suggestion
                importPhotosMap.getOrDefault(ApplePhotosConstants.COUNT_KEY, 0L).intValue())
```

your call, but seems the name was a bit too generic, leading to this easy mistake, so maybe the making both maps specific to the data type will help.

jzacsh (changes requested):

extensions/data-transfer/portability-data-transfer-apple/src/main/java/org/datatransferproject/datatransfer/apple/photos/AppleMediaImporter.java
jzacsh: note i hit "request changes" but just wanted to make sure you see this comment. feel free to mark it resolved and not do this.

jzacsh (approved):

PR #1309, opened 2023-12-01
fix: include timestamp in DriveImporter root folder

alexeyqu-fb:
Change the root folder name from `MigratedContent` to `transfer-<timestamp>`.

|Before|After|
|---|---|
|MigratedContent|transfer-2023:11:29-23:01:28|

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/drive/DriveImporter.java
+++ b/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/drive/DriveImporter.java
@@ -12,6 +12,8 @@ import com.google.api.services.drive.model.File;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 import java.io.IOException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
 import java.util.UUID;
 import org.datatransferproject.api.launcher.Monitor;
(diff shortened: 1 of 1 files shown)


extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/drive/DriveImporter.java
```diff
@@ -153,4 +156,10 @@ private synchronized Drive getDriveInterface(TokensAndUrlAuthData authData) {
 
     return driveInterface;
   }
+
+  private String getRootFolderName() {
+    return String.format(
+            ROOT_FOLDER_FORMAT_STRING,
+            new SimpleDateFormat("yyyy-MM-dd-HH-mm-ss").format(new Date()));
```
wmorland, line 163: In theory we could use the created timestamp from the job to give us a more stable date. Is there a reason we didn't go down that route?

wmorland (dismissed):
LGTM, although it might be nice to have a datetime which is more meaningful to the user, see my comment

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/drive/DriveImporter.java
alexeyqu-fb: oh, good point!

wmorland (approved):

PR #1310, opened 2023-12-03
Mention the impacts the project will have on tech-illiterate people like the elderly

upintheairsheep:
Some people do not know how to move their data from one service to another manually altogether, with an example being the elderly. Normally, a professional or family member would have to be called, the former having costs attached to it and the latter causing them to become unavailable for the time being.


lisad (changes requested):
Can you please fix the part about "elderly and other tech-illiterate" folks - there's no need to make that assumption or characterization!  (I'm >50 and my dad who's almost 80 is an geological engineer who has done plenty with computers.)

Maybe "to support folks who would find it challenging to manually manage these transfers for any reason" - there are plenty of other ways to word this too

PR #1328, opened 2024-01-09
Data copier fix

aksingh737:
NullPointerException fix for PortabilityAbstractInMemoryDataCopier which was caused by trying to access a null PortabilityJob object.

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-apple/src/main/java/org/datatransferproject/datatransfer/apple/AppleTransferExtension.java
+++ b/extensions/data-transfer/portability-data-transfer-apple/src/main/java/org/datatransferproject/datatransfer/apple/AppleTransferExtension.java
@@ -33,6 +33,8 @@ import org.datatransferproject.datatransfer.apple.photos.ApplePhotosImporter;
 import org.datatransferproject.datatransfer.apple.photos.AppleVideosImporter;
 import org.datatransferproject.spi.cloud.storage.AppCredentialStore;
 import org.datatransferproject.spi.transfer.extension.TransferExtension;
+import org.datatransferproject.spi.transfer.idempotentexecutor.IdempotentImportExecutor;
+import org.datatransferproject.spi.transfer.idempotentexecutor.IdempotentImportExecutorExtension;
 import org.datatransferproject.spi.transfer.provider.Exporter;
 import org.datatransferproject.spi.transfer.provider.Importer;
 import org.datatransferproject.types.common.models.DataVertical;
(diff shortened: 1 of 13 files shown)


portability-transfer/src/main/java/org/datatransferproject/transfer/copier/PortabilityAbstractInMemoryDataCopier.java
```diff
@@ -126,7 +126,7 @@ protected ExportResult<?> copyIteration(
     if (exportedData != null) {
       PortabilityJob job = jobStore.findJob(jobId);
       TransferMode transferMode =
-          job.transferMode() == null ? TransferMode.DATA_TRANSFER : job.transferMode();
+          job == null ? TransferMode.DATA_TRANSFER : job.transferMode();
```
jzacsh, line 129: looks like we already use Preconditions.checkNotNull library in DTP; maybe just fail on the first line of this entire function (copyIteration with a valid job id is a larger/different issue), and then leave this code as is?


see [`Preconditions.checkNotNull(id, "transfer job ID required for GetTransferJobAction");`](https://github.com/google/data-transfer-project/blob/a81eaed0ba1ecbbdd14ff4a4e844a27a09fdd537/portability-api/src/main/java/org/datatransferproject/api/action/transfer/GetTransferJobAction.java#L37)

jzacsh (changes requested):

portability-transfer/src/main/java/org/datatransferproject/transfer/copier/PortabilityAbstractInMemoryDataCopier.java
aksingh737: Changed the code as suggested

portability-transfer/src/main/java/org/datatransferproject/transfer/copier/PortabilityAbstractInMemoryDataCopier.java
```diff
@@ -125,6 +126,8 @@ protected ExportResult<?> copyIteration(
     DataModel exportedData = exportResult.getExportedData();
     if (exportedData != null) {
       PortabilityJob job = jobStore.findJob(jobId);
+      Preconditions.checkNotNull(job, "Transfer cannot proceed without fetching the job. "
```
jzacsh, line 129: lgtm; optional nit about including id and compacting the message a bit:

```suggestion
      checkNotNull(job, "Transfer job somehow from jobStore, mid-way through copy iteration, for the Job ID \"%s\"", jobId.toString());
```

jzacsh (changes requested):

portability-transfer/src/main/java/org/datatransferproject/transfer/copier/PortabilityAbstractInMemoryDataCopier.java
```diff
@@ -16,6 +16,7 @@
 package org.datatransferproject.transfer.copier;
 
 import com.google.cloud.datastore.DatastoreException;
+import com.google.common.base.Preconditions;
```
jzacsh, line 19:
```suggestion
import static com.google.common.base.Preconditions.checkNotNull;
```

jzacsh (changes requested):

portability-transfer/src/main/java/org/datatransferproject/transfer/copier/PortabilityAbstractInMemoryDataCopier.java
aksingh737: Done

portability-transfer/src/main/java/org/datatransferproject/transfer/copier/PortabilityAbstractInMemoryDataCopier.java
aksingh737: Done

sparuvu (approved):

sparuvu (approved):

PR #1329, opened 2024-01-19
add Retry and skip logic in AppleMediaImporter

yy9669:
The major difficulty I met is that the IdempotentExecutor used by CallableImporter cannot be a retryable executor, so I have to use a different retryable executor in AppleMediaImporter itself.

- Use retryable executor in AppleMediaImporter
- Add Apple Photos error prefix to all errors thrown by AppleMediaInterface
- Please help add the following Apple retry strategy to the yaml config:
    - Skippable:  ".\*APPLE PHOTOS IMPORT: Fail to upload content.\*" , ".\*APPLE PHOTOS IMPORT: Fail to create album.\*", ".\*APPLE PHOTOS IMPORT: Fail to create media.\*", ".\*APPLE PHOTOS IMPORT: Fail to get upload url.\*"
    - Non-Retryable:  ".\*APPLE PHOTOS IMPORT: iCloud Photos Library not found.\*", ".\*APPLE PHOTOS IMPORT: iCloud Storage is full.\*"

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-apple/src/main/java/org/datatransferproject/datatransfer/apple/photos/AppleMediaImporter.java
+++ b/extensions/data-transfer/portability-data-transfer-apple/src/main/java/org/datatransferproject/datatransfer/apple/photos/AppleMediaImporter.java
@@ -18,12 +18,18 @@ package org.datatransferproject.datatransfer.apple.photos;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
+
+import java.io.IOException;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.UUID;
 import org.datatransferproject.api.launcher.Monitor;
 import org.datatransferproject.datatransfer.apple.AppleInterfaceFactory;
 import org.datatransferproject.datatransfer.apple.constants.ApplePhotosConstants;
 import org.datatransferproject.spi.transfer.idempotentexecutor.IdempotentImportExecutor;
+import org.datatransferproject.spi.transfer.idempotentexecutor.RetryingInMemoryIdempotentImportExecutor;
 import org.datatransferproject.spi.transfer.provider.ImportResult;
 import org.datatransferproject.spi.transfer.provider.Importer;
 import org.datatransferproject.transfer.JobMetadata;
@@ -31,7 +37,12 @@ import org.datatransferproject.types.common.models.DataVertical;
 import org.datatransferproject.types.common.models.media.MediaContainerResource;
 import org.datatransferproject.types.transfer.auth.AppCredentials;
 import org.datatransferproject.types.transfer.auth.TokensAndUrlAuthData;
+import org.datatransferproject.types.transfer.errors.ErrorDetail;
 import org.jetbrains.annotations.NotNull;
+
+import javax.annotation.Nullable;
+
+import static java.lang.String.format;
 
 /**
  * An Apple importer to import the Photos and Videos into Apple iCloud-photos.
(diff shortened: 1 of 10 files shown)


kateyeo:
Hi Xuyang, just saw this, I'll review it today

extensions/data-transfer/portability-data-transfer-apple/src/main/java/org/datatransferproject/datatransfer/apple/photos/AppleMediaImporter.java
```diff
@@ -67,38 +84,68 @@ public ImportResult importItem(
...
+                              MediaContainerResource.PHOTOS_COUNT_DATA_NAME,
+                              importPhotosMap.getOrDefault(ApplePhotosConstants.COUNT_KEY, 0L).intValue())
+                      .put(
+                              MediaContainerResource.VIDEOS_COUNT_DATA_NAME,
+                              importVideosResult.getOrDefault(ApplePhotosConstants.COUNT_KEY, 0L).intValue())
+                      .build());
+
+      Collection<ErrorDetail> errors = idempotentExecutor.getRecentErrors();
+      if (!errors.isEmpty() && executor instanceof RetryingInMemoryIdempotentImportExecutor) { // throw the error for retryExecutor to retry, only include the actual error message, but not the stack traces
+        throw new IOException(errors.iterator().hasNext() ? errors.iterator().next().exception().lines().findFirst().get() : ApplePhotosConstants.APPLE_PHOTOS_IMPORT_ERROR_PREFIX + " Unknown Error");
```
osamahan999, line 145: Why do we not want to throw the entire stack trace? Is this logic necessary? Can we not just let the errors be thrown by the executor?

extensions/data-transfer/portability-data-transfer-apple/src/main/java/org/datatransferproject/datatransfer/apple/photos/AppleMediaInterface.java
```diff
@@ -443,15 +444,15 @@ public int importAlbums(
...
                 AuditKeys.albumId, mediaAlbum.getId(),
                 AuditKeys.errorCode, newPhotoAlbumResponse.getStatus().getCode());
 
-              throw new IOException(
+              throw new IOException(getApplePhotosImportThrowingMessage(
                 String.format(
-                  "Failed to create album, error code: %d",
-                  newPhotoAlbumResponse.getStatus().getCode()));
+                  "Fail to create album, error code: %d",
+                  newPhotoAlbumResponse.getStatus().getCode())));
```
osamahan999, line 455: Logging & throwing is unnecessary, we should usually only do one of the two. 

If we log here, the client which received the error may double-log, or they may throw the error leaving the logs in the stack trace

osamahan999:
I'm going to leave the actual review to @kateyeo because I don't understand some parts of this PR.

extensions/data-transfer/portability-data-transfer-apple/src/main/java/org/datatransferproject/datatransfer/apple/photos/AppleMediaImporter.java
yy9669: If we let the error be thrown by idempotentExecutor midway of the import, then it will fail the whole batch even if only one file failed. 
Here we throw after the importResult was updated, and the importResult will include all the files except for the failed one.

kateyeo (commented):
Hi, looking over this and just wanting to make sure I'm understanding how your implementing retrying in this PR. By using the retrying executor on the entire call within  importItem() the entire call would be retried if you throw a retry-able exception. As opposed to if you passed in a retrying executor to the individual importAlbums/importAllMedia calls which would allow skipping/retrying to be handled a the item level.

Let me know if you have it set up how you intended or if I'm misunderstanding anything

yy9669:
> Hi, looking over this and just wanting to make sure I'm understanding how your implementing retrying in this PR. By using the retrying executor on the entire call within importItem() the entire call would be retried if you throw a retry-able exception. As opposed to if you passed in a retrying executor to the individual importAlbums/importAllMedia calls which would allow skipping/retrying to be handled a the item level.
> 
> Let me know if you have it set up how you intended or if I'm misunderstanding anything

Yes the is intended, I want to retry the whole batch. 
All Apple calls are handled in batches, and it's step by step - get upload url, upload to Apple Content, and create metadata.

Sometimes failure in one step is due to corruption in previous step, eg. fail to upload to Apple Content is because we got wrong upload url in previous step, so it's useful to retry from the first step instead of retrying current step multiple times.

extensions/data-transfer/portability-data-transfer-apple/src/main/java/org/datatransferproject/datatransfer/apple/photos/AppleMediaImporter.java
yy9669: Oh just to make regex easier.

kateyeo (approved):
Ok sounds good, thanks for the response

PR #1332, opened 2024-02-01
Preserve filenames when importing items

aksingh737:
Preserve filenames when importing items by populating the filename field in the SimpleMediaItem as described in Google Photos API doc - https://developers.google.com/photos/library/reference/rest/v1/mediaItems/batchCreate#simplemediaitem


jzacsh (approved):
eek. good find. we should follow-up with unit test coverage somewhere :|

PR #1334, opened 2024-02-13
feat: add userAlias field to PortabilityJob class

alexeyqu-fb:
Needed to customize folder names on the destination sides.

`./gradlew build test`


wmorland (approved):
LGTM. When possible we should add an example of how this is used either in comment or code.

alexeyqu-fb:
This will be done in follow-up PRs, yes.

PR #1338, opened 2024-02-29
Fix swallowed apple server errors

jzacsh:
failures to upload to Apple servers (via [`AppleMediaInterface#importAllMedia`](https://github.com/google/data-transfer-project/blob/b4c3c89821b8356f466c112e22193e8bfd32100d/extensions/data-transfer/portability-data-transfer-apple/src/main/java/org/datatransferproject/datatransfer/apple/photos/AppleMediaInterface.java#L462)) now include the root cause in their stack trace.

nitty details of this change:
- noop~ish: improve error handling: capturing errors as specifically as is reasonable:
  - ensure legit internal `apple.photos.streaming.StreamingContentClient` connection-setup (constructor) errors are distinguished from illegal states (like bad HTTP Methods)
  - ensure download/upload sequence (encapsulated by the `AppleMediaInterface#uploadContent` method) bubbles error-reasons up:
     - changes its API to return something even on failure; added a new oneof called `DownUpResult` for this purpose (see the change to the [previously-empty `continue` catch-block](https://github.com/google/data-transfer-project/blob/b4c3c89821b8356f466c112e22193e8bfd32100d/extensions/data-transfer/portability-data-transfer-apple/src/main/java/org/datatransferproject/datatransfer/apple/photos/AppleMediaInterface.java#L210))
     - when skipping errors, bubble the cause (the `Throwable`) back up via `IdempotentImportExecutor#executeAndSwallowIOExceptions` (there was already a `executeAndSwallowIOExceptions` but the [exception it previously threw didn't pass a second param](https://github.com/google/data-transfer-project/blob/b4c3c89821b8356f466c112e22193e8bfd32100d/extensions/data-transfer/portability-data-transfer-apple/src/main/java/org/datatransferproject/datatransfer/apple/photos/AppleMediaInterface.java#L562) (the "cause") to its `IOException`, and now it does). 
- nit: making the internal `apple.photos.streaming.StreamingContentClient` an `AutoCloseable` so that we can get a `HttpURLConnection#disconnect` via [try-with-resource](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html)
- noop: test refactoring
  - updated test-double logic now that `uploadContent` returns something for _every_ file passed in
  - changed some test-double setup code so it reuses the `uploadContent` spy infra that was already in place

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-apple/src/main/java/org/datatransferproject/datatransfer/apple/photos/DownUpResult.java
+++ b/extensions/data-transfer/portability-data-transfer-apple/src/main/java/org/datatransferproject/datatransfer/apple/photos/DownUpResult.java
@@ -23,7 +23,7 @@ public abstract class DownUpResult {
   }
 
   /**
-   * Data-ID returned from Apple servers, indicaing the upload start is successful.
+   * Data-ID returned from Apple servers, indicating the upload start is successful.
    *
    * <p>Sometimes called the "single file upload response"
    */
--- a/extensions/data-transfer/portability-data-transfer-apple/src/main/java/org/datatransferproject/datatransfer/apple/photos/DownUpResult.java
+++ b/extensions/data-transfer/portability-data-transfer-apple/src/main/java/org/datatransferproject/datatransfer/apple/photos/DownUpResult.java
@@ -1,0 +1,34 @@
+package org.datatransferproject.datatransfer.apple.photos;
+
+import com.google.auto.value.AutoOneOf;
+
+@AutoOneOf(DownUpResult.Status.class)
+public abstract class DownUpResult {
+  public enum Status {
+    SUCCESS_DATA_ID,
+    ERROR
+  }
+
+  public abstract Status getStatus();
+
+  public boolean isOk() {
+    return this.getStatus().equals(Status.SUCCESS_DATA_ID);
+  }
+
+  /**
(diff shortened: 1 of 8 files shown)


jzacsh:
moving out of "draft" state so folks know to review it. assigning to folks on my team to take a look for trivial stuff before I bug the Apple team. Though feel free to chime in early, Sundeep/Xuyang!

extensions/data-transfer/portability-data-transfer-apple/src/main/java/org/datatransferproject/datatransfer/apple/photos/DownUpResult.java
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  public abstract Status getStatus();
+
+  public boolean isOk() {
+    return this.getStatus().equals(Status.SUCCESS_DATA_ID);
+  }
+
+  /**
+   * Data-ID returned from Apple servers, indicaing the upload start is successful.
```
georginaLopez, line 26: nit: minor typo
```suggestion
   * Data-ID returned from Apple servers, indicating the upload start is successful.
```

jzacsh: fixed, thanks!

yy9669 (approved):
LGTM

PR #1341, opened 2024-03-09
bugfix: NullPointerExc swallowing real errors

jzacsh:
ImmutableMap is being used to log error cases during a transfer. Unfortunately said error cases are _themselves_ obfuscated by a less interesting `NullPointerException` due to ImmutableMap receiving nulls.

ImmutableMap doesn't allow null values (see guava's issues 1782) but it's being passed an `albumId` sometimes set to null. This might be happening in multiple places but I've at least seen it in the primary download/upload loop (on line 588, before this change).


holachuy (approved):
LGTM.  Not sure if this is test worthy or maybe in the future we provide a wrapper for the error map that does this internally.

PR #1342, opened 2024-03-13
Tokensurlauth partial primitive obsession fix

jzacsh:
open the door to let us start moving away from "primitive obsession"
(strings, all strings), which _in turn_ lets us start handling errors
more concisely and int he right places.

this is a prefactor for an upcoming apple-music PR (#1339) that relies
on this class for oauth token refresh.


hgandhi90 (approved):
Thanks @jzacsh !

PR #1348, opened 2024-03-20
add skip logic in PortabilityAbstractInMemoryDataCopier import module…

yy9669:
- add skip logic in PortabilityAbstractInMemoryDataCopier import module
- refactor AppleMediaImporter to remove in extension retry/skip, and remove corresponding tests
- remove unnecessary logs which was used to check the missing creationDate issue, since we already resolved it.


extensions/data-transfer/portability-data-transfer-apple/src/main/java/org/datatransferproject/datatransfer/apple/photos/AppleMediaImporter.java
```diff
@@ -88,90 +71,38 @@ public ImportResult importItem(
       return ImportResult.OK;
     }
 
-    // log
-    for (PhotoModel photoModel: data.getPhotos()) {
```
akshayau, line 92: it might be useful to keep these logs, maybe at a debug level

extensions/data-transfer/portability-data-transfer-apple/src/main/java/org/datatransferproject/datatransfer/apple/photos/AppleMediaImporter.java
```diff
@@ -88,90 +71,38 @@ public ImportResult importItem(
...
+            data.getPhotos(),
+          DataVertical.MEDIA.getDataType());
+    final Map<String, Long> importVideosResult =
+        mediaInterface.importAllMedia(
+            jobId,
+            idempotentExecutor,
+            data.getVideos(),
+          DataVertical.MEDIA.getDataType());
+
+    final Map<String, Integer> counts =
```
akshayau, line 96: can we log the counts after import?

kateyeo (dismissed):
Looks good! Thanks for getting the fix in so quickly

kateyeo (approved):

PR #1375, opened 2024-07-12
Handshake signal

sparuvu:
The code changes in this PR are to add the Signalling flows to the existing framework. Whenever a transfer job hits a milestone (Started, Completed, Errored, Cancelled, etc) it is helpful to the providers to know the status of the transfer job. 

The signaling framework should be passive to the existing job processing flows and come with a default implementation where it just logs the event. Providers can add their implementation where an API can be hooked up send the signal across the board.


portability-transfer/src/main/java/org/datatransferproject/transfer/JobProcessor.java
jzacsh, line 1: perhaps we should add javadoc (in this PR) to *both*:
- [extant `EventCode`](https://github.com/dtinit/data-transfer-project/blob/c7b9684059733db21dceab8f2363e0111340f1c4/portability-api-launcher/src/main/java/org/datatransferproject/launcher/monitor/events/EventCode.java) class
- this _new_ `SignalType`

and have both javadocs refer to eachother to clarify how they do/don't relate to eachother

portability-transfer/src/main/java/org/datatransferproject/transfer/JobProcessor.java
```diff
@@ -159,6 +177,8 @@ void processJob() {
           EventCode.WORKER_JOB_FINISHED);
       addErrorsAndMarkJobFinished(jobId, success, loggedErrors);
       hooks.jobFinished(jobId, success);
+      SignalType finalStatus = success ? SignalType.JOB_COMPLETED : SignalType.JOB_ERRORED;
```
jzacsh, line 180: note: looking at `success` here is probably getting into the weeds of a more nuanced state then we need to for your signals-change. perhaps drop this to be a static `JOB_DONE` (and/or incorporate the other comment I've made above directly on the signals enum)

----

(this is more a discussion for the SignalType enum, but making it here for clarity)

replace `JOB_COMPLETED` `JOB_PERFECT_TRANSFER` and/or replace `JOB_ERRORED` with `JOB_COMPLETED_PARTIAL`

also: we can tell if something was _skipped_ if `success && !loggedErrors.isEmpty()`


tl;dr replace `COMPLETED` with `DONE` and have a suffix to capture nuance (eg: `_PARTIAL` or `_PERFECT`).

portability-spi-transfer/src/main/java/org/datatransferproject/spi/transfer/types/signals/SignalType.java
```diff
@@ -0,0 +1,7 @@
+package org.datatransferproject.spi.transfer.types.signals;
+
+public enum SignalType {
```
jzacsh, line 3: (see next comments for context) here's a way we could maintain one chunk of unified logic insdie DTP that delivers both views of a job's lifecycle.

```
public enum JobLifecycleChange {
   ...._PERFECT,
  ... _PARTIAL,
  ..._SKIPPED,

}
```


```java
/** Whether the job is running still. All other nuance should be captured elsewhere. */
public enum JobLifecycle {
  JOB_STARTED,
  JOB_DONE
}

public static JobLifecycle from(JobLifecycleChange change) {
}
```

jzacsh (commented):
(ephemeral notes that are only gonna make sense in tandem to our video chat)

calumcalder:
The `SignalHandler` abstraction you've created seems very similar to the existing `JobHooks` mechanism - they both hook in to the lifecycle of a job at the same places, differing mostly in the mechanisms for registering the hook; `JobHook`s are registered by extension modules, and `SignalHandler`s are registered by the `TransferExtension`s being used.

Rather than duplicating the hooks concept (which I assume you couldn't use as you currently have no way to create a hook as the importer) I expect it would be tidier to instead extend the existing `JobHooks` mechanism to allow registering hooks from the `Importer` and `Exporter` (I suggest at the `(Im|Ex)porter` rather than `TransferExtension` so you have access to the `AuthData` required to make calls to your API, and no longer have to pass auth data through the signal/hook interface).

sparuvu:
> which I assume you couldn't use as you currently have no way to create a hook as the importer

Exactly. 

> I expect it would be tidier to instead extend the JobHooks mechanism to allow registering hooks from the Importer and Exporter

If I am understanding the suggestion correctly, the idea is to create Provider specific JobHooks which are expected to load only for those specific providers. At this point JobHooks & SignalHandlers are same (different names). 

But there is another problem, SignalHandlers can't be an implementation of [JobHook](https://github.com/dtinit/data-transfer-project/blob/d228037bb8e661cd1ebe31c517c6721ed90652d6/portability-spi-transfer/src/main/java/org/datatransferproject/spi/transfer/hooks/JobHooks.java) as all the JobHooks are loaded via SPIs. We need to change the way they are intended to load. Which would be a non-passive change for all forks DTP platforms.

The key distinction here is that...

1. JobHooks are common for all the Providers. Hence they are loaded at runtime for all the providers. 
2. SignalHandlers are specific to providers. So, they are loaded only when the provider specific extensions are loaded.

extensions/data-transfer/portability-data-transfer-apple/src/main/java/org/datatransferproject/datatransfer/apple/music/AppleMusicInterface.java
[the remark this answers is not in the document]
sparuvu: this is supposed to be refactored to ensure that AppleBaseInterface is being used by both media & playlists. This is just a temp workaround. I am planning on a refactor once we get this in.

extensions/data-transfer/portability-data-transfer-apple/src/main/java/org/datatransferproject/datatransfer/apple/photos/AppleMediaInterface.java
```diff
@@ -319,7 +321,7 @@ public byte[] makePhotosServicePostRequest(
     } catch (CopyExceptionWithFailureReason e) {
       if (e instanceof UnconfirmedUserException
           || e instanceof PermissionDeniedException) {
-        refreshTokens();
+        this.authData = refreshTokens(authData, appCredentials, monitor);
```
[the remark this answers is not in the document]
sparuvu: this is going back to the default method in `AppleBaseInterface`

portability-transfer/src/main/java/org/datatransferproject/transfer/JobProcessor.java
```diff
@@ -170,6 +190,23 @@ void processJob() {
...
+      .withJobStatus(signalType.name())
+      .withExportingService(JobMetadata.getExportService())
+      .withImportingService(JobMetadata.getImportService())
+      .build();
+
+    try {
+      exportSignalHandlerProvider.get().sendSignal(signalRequest, exportAuthData, monitor);
+      importSignalHandlerProvider.get().sendSignal(signalRequest, importAuthData, monitor);
+    } catch (CopyExceptionWithFailureReason | IOException | RetryException e) {
+      // Swallow the exception. The Job Processing shouldn't be effected by Signal failure.
```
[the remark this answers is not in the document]
sparuvu: there should be logs related to any failure from. Adding this here would be redundant. But I can add here as well for completeness.

extensions/data-transfer/portability-data-transfer-apple/src/main/java/org/datatransferproject/datatransfer/apple/AppleBaseInterface.java
```diff
@@ -16,8 +16,106 @@
...
  * The Base Interface for all the {@link org.datatransferproject.types.common.models.DataVertical}s
  * that Apple Supports.
  */
-public interface AppleBaseInterface { }
+public interface AppleBaseInterface {
+    default String sendPostRequest(@Nonnull String url, @Nonnull final byte[] requestData) throws IOException, CopyExceptionWithFailureReason {
+      throw new NotImplementedException("sendPostRequest is not implemented !! ");
+    }
+
+    default TokensAndUrlAuthData refreshTokens(final TokensAndUrlAuthData authData, final AppCredentials appCredentials, Monitor monitor)
```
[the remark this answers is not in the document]
lisad: I agree that there could be a lot of benefit from reusing OAuth client code; is there a longer term plan for that rather than just tactical ideas which can result in a lot of whackamole work?  It could obviously be part of a separate PR anyway

lisad (dismissed):

extensions/data-transfer/portability-data-transfer-apple/src/main/java/org/datatransferproject/datatransfer/apple/AppleBaseInterface.java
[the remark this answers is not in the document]
sparuvu: don't want to mix up signalling code with refactoring. hence did not pushed into this PR.

portability-spi-transfer/src/main/java/org/datatransferproject/spi/transfer/provider/SignalHandler.java
```diff
@@ -0,0 +1,36 @@
...
+import org.datatransferproject.spi.transfer.types.signals.SignalType;
+import org.datatransferproject.types.transfer.auth.AuthData;
+import org.datatransferproject.types.transfer.auth.TokensAndUrlAuthData;
+import org.datatransferproject.types.transfer.retry.RetryException;
+
+/**
+ * Signals the status of the Job to the corresponding Service.
+ */
+public interface SignalHandler<A extends AuthData> {
+  default void sendSignal(SignalRequest signalRequest, AuthData authData, Monitor monitor)
```
calumcalder, line 32: Maybe a bit lost in my earlier comment, but wanted to reiterate that `authData` feels out of place here - that you need `AuthData` to send the signal is an implementation detail of the `AppleSignalHandler` which is leaking through to the abstraction here.

extensions/data-transfer/portability-data-transfer-apple/src/main/java/org/datatransferproject/datatransfer/apple/signals/AppleSignalInterface.java
```diff
@@ -0,0 +1,137 @@
...
+
+    HttpURLConnection con = null;
+    String responseString = null;
+    try {
+      URL signalUrl = new URL(url);
+
+      con = (HttpURLConnection) signalUrl.openConnection();
+      con.setDoOutput(true);
+      con.setRequestMethod("POST");
+      con.setRequestProperty(Headers.AUTHORIZATION.getValue(), authData.getAccessToken());
```
calumcalder, line 90: Following on from my comment in `SignalHandler` - do you need to be using `authData` at all here? `authData` are credentials granted by the user to act on their behalf, and while I'm not aware of what your infra is doing on the other end of this call it seems more like the worker itself informing your infrastructure of a job starting/stopping, in which case `appCredentials` should be sufficient.

calumcalder (commented):
> We need to change the way they are intended to load. Which would be a non-passive change for all forks DTP platforms.

That would depend on how you make the change - if you were to add a `JobHooksBase` interface in`portability-types-transfer` and make both `AppleSignalHandler` (perhaps via a `TransferExtensionJobHooks` abstraction) and `JobHooks` an implementation of `JobHooksBase`, then the `portability-transfer` worker could load both the existing SPI-based `JobHooks` and new `TransferExtensionJobHooks` in a backwards-compatible way.

extensions/data-transfer/portability-data-transfer-apple/src/main/java/org/datatransferproject/datatransfer/apple/signals/AppleSignalInterface.java
sparuvu: think of `appCrentials` as a contract between the OAuth and the Client of OAuth (which is DTP). Not even the public APIs have access to that secret. The Signalling APIs are no different from public APIs. App credentials are not the right auth materials to authorize this call. Moreover, we also need to authenticate this user. Without that, DTP platform can update the status of a job of any user.

portability-spi-transfer/src/main/java/org/datatransferproject/spi/transfer/provider/SignalHandler.java
sparuvu: If a provider makes a decision to implement their SignalHandler then that would have to be a public API protected by some Auth Material. This auth material should guarantee that worker is updating the job that belongs to its owner. The default implementation just ignores it.

calumcalder: I agree that if you're calling an API with your `SignalHandler` it should be authorized, what I'm suggesting is that passing the auth data in the `sendSignal` call leaks that implementation detail - that the default implementation ignores it suggests it's better provided elsewhere, for example in the constructor of specific `SignalHandler`s that require it

extensions/data-transfer/portability-data-transfer-apple/src/main/java/org/datatransferproject/datatransfer/apple/signals/AppleSignalInterface.java
calumcalder: > The Signalling APIs are no different from public APIs

For the reasons I mentioned in my earlier comment I don't believe that's the case, but if that's the implementation approach you're taking I think that's all the more reason to not include `authData` in the public API of `SignalHandler` - by not leaking that implementation detail in to the abstraction, you could implement it as part of your public API while another user could implement it through a telemetry API that uses another auth scheme.

portability-spi-transfer/src/main/java/org/datatransferproject/spi/transfer/provider/SignalHandler.java
calumcalder: Per our conversation during the maintainers meeting, resolving as it's consistent with existing APIs

extensions/data-transfer/portability-data-transfer-apple/src/main/java/org/datatransferproject/datatransfer/apple/signals/AppleSignalInterface.java
calumcalder: As above

portability-transfer/src/main/java/org/datatransferproject/transfer/JobProcessor.java
```diff
@@ -124,6 +142,7 @@ void processJob() {
           JobMetadata.getExportService(),
           JobMetadata.getImportService());
       JobMetadata.getStopWatch().start();
+      sendSignals(jobId, exportAuthData, importAuthData, JobLifeCycle.JOB_STARTED(), monitor);
       copier.copy(exportAuthData, importAuthData, jobId, exportInfo);
```
jzacsh, line 146: One critical change: can you please add flag guarding here. This is that settings/value propagating mechanism I shared. Just wrap this new line and the next two new lines in a conditional ([eg](https://github.com/dtinit/data-transfer-project/blob/b92d3317efdbe127f101fab9cfb693e0d94b9bb7/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/GoogleTransferExtension.java#L120) `if (context.getSetting("useJobSignal", false)) {`)

portability-transfer/src/main/java/org/datatransferproject/transfer/JobProcessor.java
```diff
@@ -159,6 +178,8 @@ void processJob() {
           EventCode.WORKER_JOB_FINISHED);
       addErrorsAndMarkJobFinished(jobId, success, loggedErrors);
       hooks.jobFinished(jobId, success);
+      JobLifeCycle finalStatus = deriveFinalJobStatus(success);
```
jzacsh, line 181: Wrap these two lines as well (see above comment)

jzacsh (changes requested):
Looks great just missed one critical offline-chat point about flag guarding. Once we add that (should really be a few lines) we can deploy ASAP.

jzacsh (approved):

PR #1380, opened 2024-08-16
Remove check for `authPublicKey` when claiming jobs

calumcalder:
This check was originally added in 9d5302c, where it was used to check that two workers didn't attempt to claim the same job. Since then the surrounding code has seen a lot of refactoring, including migrating to calling `claimJob` on the job store which, per the comment in `JobPollingService` and on the `JobStore` interface docs, should do the job of ensuring this doesn't happen.

Removing this line unblocks the option of allowing workers to reclaim jobs from another crashed worker that couldn't inform the job store it was exiting (e.g. OOM killed), but also continues to allow blocking `claimJob` on claimed jobs by delegating that decision to the `JobStore`.

### DIFF
--- a/portability-transfer/src/main/java/org/datatransferproject/transfer/JobPollingService.java
+++ b/portability-transfer/src/main/java/org/datatransferproject/transfer/JobPollingService.java
@@ -153,11 +153,6 @@ class JobPollingService extends AbstractScheduledService {
     // Lookup the job so we can append to its existing properties.
     PortabilityJob existingJob = store.findJob(jobId);
     monitor.debug(() -> format("JobPollingService: tryToClaimJob: jobId: %s", existingJob));
-    // Verify no transfer worker key
-    if (existingJob.jobAuthorization().authPublicKey() != null) {
-      monitor.debug(() -> "A public key cannot be persisted again");
-      return false;
-    }
 
     // TODO: Consider moving this check earlier in the flow
     String scheme = existingJob.jobAuthorization().encryptionScheme();


portability-transfer/src/main/java/org/datatransferproject/transfer/JobPollingService.java
```diff
@@ -153,11 +153,6 @@ private boolean tryToClaimJob(UUID jobId, WorkerKeyPair keyPair) {
     // Lookup the job so we can append to its existing properties.
     PortabilityJob existingJob = store.findJob(jobId);
     monitor.debug(() -> format("JobPollingService: tryToClaimJob: jobId: %s", existingJob));
-    // Verify no transfer worker key
```
seehamrun, line 156: Since the state check is being done in the default impl of `claimjob` this should be safe because checking the state is still creds_avaialble is the same as checking if this job has the worker key information since a state change happens when that information is written

if this turns out to be an issue, anyone requiring this check can add it in as an additional check in claimjob in their impl of jobstore

 https://github.com/dtinit/data-transfer-project/blob/d228037bb8e661cd1ebe31c517c6721ed90652d6/portability-spi-cloud/src/main/java/org/datatransferproject/spi/cloud/storage/JobStoreWithValidator.java#L19-L23

seehamrun (approved):

calumcalder:
Thanks for the review! Just waiting on legal to review the new CLA after the switch to DTI ownership

calumcalder:
CLA signed 👍

PR #1386, opened 2024-10-15
Handle 401 errors on Google Video uploads correctly

calumcalder:
#1262 was created in an attempt to propagate `InvalidTokenException`s from `GoogleVideosInterface` on 401 errors, but didn't handle the case where an `ApiException` is raised from `photosLibraryClient.uploadMediaItem` rather than returned in the `UploadMediaItemResponse` error field.

This PR adds test cases for both ways the error may be returned, and fixes the above case.


extensions/data-transfer/portability-data-transfer-google/src/test/java/org/datatransferproject/datatransfer/google/videos/GoogleVideosImporterTest.java
```diff
@@ -85,6 +91,15 @@ public class GoogleVideosImporterTest {
   private PhotosLibraryClient client;
   private UUID jobId;
 
+  class TestStatusCode implements StatusCode {
```
calumcalder, line 94: I couldn't find an obvious implementation of the `StatusCode` interface in the codebase - happy to swap this out with something more sensible if it exists

seehamrun (approved):

jzacsh (approved):

Issue #1158, opened 2022-09-08
techdebt: backfill missing unit tests for GooglePhotosInterface.java


jzacsh:
@MewX ran into this at https://github.com/google/data-transfer-project/pull/1155#issuecomment-1240244384

mahika2:
Hey, is this issue still open? If yes, can I work on this?

anand007s:
please assign it to me...

FilippoULIVO:
Hey, I noticed that most methods on the class depend on private method `makeGetRequest` in order to mock it I would need to make it visible but I don't want to break encapsulation. I noticed there is also a `makePostRequest` which is public and is used for mocked responses in `GooglePhotosImporterTest`. If it's okay to change the access modifier on the method I can complete unittesting the methods dependent on `makeGetRequest`.

MewX:
Would package-private sufficient for the tests? If so, usually you could mark these methods package-private and annotate them with `@VisibleForTesting`.

FilippoULIVO:
yes ok it works. currentlyt working on this

FilippoULIVO:
Hey ,sorry for the noob question, I can't find this info, I'm trying to make a first commit with most of the test cases but I'm not able to push to origin, I'm trying ` git push -u origin test-importer-interface`   and receiving 403 error with message `remote: Permission to google/data-transfer-project.git denied to FilippoULIVO` I just submitted my CLA agrrement, but i'm not sure it's related to that. could you help? 

wanted to make a first commits also to check if I could pass all the different style checks

FilippoULIVO:
I completed the test coverage for the class. 
Btw  @MewX as I was sayng`makePostRequest` is public, I didn't write test for it because I belive it should be made package-private just as the `makeGetRequest` ,having it public defeats the purpose of the interface, from IntelliJ I can see it's only used in tests as mock.
I would need some guidance on the Pull request process for submitting my contribute

jzacsh:
> I'm trying to make a first commit with most of the test cases but I'm not able to push to origin, I'm trying  git push -u origin test-importer-interface and receiving 403 error with message remote: Permission to google/data-transfer-project.git denied to FilippoULIVO

I believe that's because you're accidentally trying to develop directly on this repo, but you instead want to maintain your branch _on your own repo_ (both in draft state and when you're ready to open a PR for it). You can get your own repo by clicking the "fork" button (or visiting https://github.com/google/data-transfer-project/fork).

<details>
[...]

PR #1387, opened 2024-11-01
Solution for #1158

7321569Lhy:
An unit test for GooglePhotosInterface.java


lisad:
This is a very narrow test, whereas the method under test is already used in some other tests that do more -- and there's also PR 1257 which we do need to follow up on.

PR #1389, opened 2024-11-14
feat: Generic Importers

calumcalder:
To support importing data via DTP, developers are required to implement a Java extension module exposing a `TransferExtension` and `Importer`, with the `Importer` interfacing with a new or existing API also supported by the developer. Creating a Java extension creates a barrier for some developers who don't have the context or bandwidth to implement and maintain such an extension.
As a way to make it easier to integrate with DTP this PR implements a `GenericTransferExtension` and `GenericImporter`, which can be reused by developers without committing or maintaining Java code.

Generic Importers can be configured to call a HTTP endpoint created and maintained by the developer, allowing them to use the programming language and frameworks of their choice, as long as their API conforms to the OAuth Authorization Code Flow, and the HTTP JSON API laid out in the included README.

While this PR contains a functional implementation for the BLOBS, CALENDAR, MEDIA, and SOCIAL-POSTS verticals, I anticipate the implementation and API will be iterated on as developers begin testing and integration.

## TODO
- [x] Auth
- [x] Configuration
  - [x] Rework `TransferExtension` loading
- [x] Data-types
  - [x] Blobs
  - [x] Calendar
  - [x] Media
  - [x] Social Posts
- [x] Documentation
  - [x] Service endpoints
  - [x] Service schema
  - [x] OAuth flow
  - [x] Configuration
- [x] Tests
  - [x] Serialization
  - [x] OAuth
  - [x] Importer


extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/GenericTransferExtension.java
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  @Override
+  public void initialize(ExtensionContext context) {
+    JobStore jobStore = context.getService(JobStore.class);
+
+    importerMap.put(
+        BLOBS,
+        new GenericFileImporter<BlobbyStorageContainerResource>(
+            (head, om) -> {
```
calumcalder, line 88: Callbacks in-line are getting a bit unwieldy, so planning on refactoring this out to another file. That'll also help with making unit tests.

extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/GenericTransferExtension.java
```diff
@@ -0,0 +1,240 @@
...
+
+              BlobbyContainerPath containerAndPath;
+              while ((containerAndPath = horizon.poll()) != null) {
+                BlobbyStorageContainerResource container = containerAndPath.getContainer();
+                String parentPath = containerAndPath.getPath();
+                String path = format("%s/%s", parentPath, container.getName());
+                // Import the current folder
+                // TODO: Consider making a POJO for mapping BlobbyFolderData and BlobbyFileData
+                ObjectNode folder = JsonNodeFactory.instance.objectNode();
+                folder.set("@type", JsonNodeFactory.instance.textNode("BlobbyFolderData"));
```
calumcalder, line 112: There's some inconsistency in `@type` fields generated manually and those autogenerated for e.g. `VideoModel` - this inconsistency should be resolved

extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/auth/OAuthTokenManager.java
```diff
@@ -0,0 +1,171 @@
...
+  public Optional<String> getExpiresIn() {
+    return expiresIn;
+  }
+}
+
+/**
+ * Utility to manage {@link TokensAndUrlAuthData} containing OAuth refresh and access tokens.<br>
+ * See {@see #withAuthData} for how to use the auth data managed by this class.
+ */
+public class OAuthTokenManager {
```
lisad, line 67: I'm surprised this isn't already part of the DTP framework. Doesn't it have to be reimplemented like this for each extension? Not something that needs to be addressed in this PR for sure, just asking!

extensions/data-transfer/portability-data-transfer-generic/src/test/java/org/datatransferproject/datatransfer/generic/GenericImportSerializerTestBase.java
```diff
@@ -0,0 +1,33 @@
...
+
+class GenericImportSerializerTestBase {
+  static ObjectMapper objectMapper = new ObjectMapper();
+
+  @BeforeClass
+  public static void onlyOnce() {
+    GenericImporter.configureObjectMapper(objectMapper);
+  }
+
+  <T> void assertJsonEquals(String expectedPayload, GenericPayload<T> actualWrapperPayload)
```
lisad, line 20: This seems really useful but the longer instances where this is used were still somewhat ugly and hard to write with all the quotes, quote escapes, and line concatenations.  Is it easy for assertJsonEquals to handle that too and make these tests even easier to write/read?  And is this another utility that is potentially used elsewhere in DTP?

portability-types-common/src/main/java/org/datatransferproject/types/common/models/calendar/RecurrenceRule.java
```diff
@@ -1,5 +1,6 @@
 package org.datatransferproject.types.common.models.calendar;
 
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
```
lisad, line 3: no idea if this matters, but this file change is only a single import line added, not used

lisad (dismissed):
love it.  The documentation is already very readable to an outsider.

extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/GenericImporter.java
```diff
@@ -0,0 +1,172 @@
...
+        error = om.readValue(body, ErrorResponse.class);
+      } catch (JsonParseException | JsonMappingException e) {
+        throw new IOException(
+            format(
+                "Unexpected response (%d) '%s'",
+                response.code(), new String(body, StandardCharsets.UTF_8)),
+            e);
+      }
+
+      if (response.code() == 401 && error.getError().equals("invalid_token")) {
```
sparuvu, line 147: isn't 401 already mean invalid token one way or the other ? I feel like the additional  check for error message is redundant.

extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/GenericImporter.java
```diff
@@ -0,0 +1,172 @@
...
+        throw new InvalidTokenException(error.toString(), null);
+      } else {
+        throw new IOException(format("Error (%d) %s", response.code(), error.toString()));
+      }
+    }
+    if (response.code() < 200 || response.code() >= 300) {
+      throw new IOException(format("Unexpected response code (%d)", response.code()));
+    }
+    return true;
```
sparuvu, line 156: this implies that we are okay with 200 response code while in the documentation it seems like we are suggesting the API should return 201?

extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/GenericImporter.java
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  boolean importSingleItem(UUID jobId, TokensAndUrlAuthData authData, ImportableData<R> dataItem)
+      throws IOException, InvalidTokenException {
+    Request request =
+        new Request.Builder()
+            .url(endpoint)
+            .addHeader("Authorization", format("Bearer %s", authData.getToken()))
```
sparuvu, line 164: you don't have to address this now. But probably a good practice to extract these common strings to a highlevel constants file that is accessible to all the extensions.

extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/GenericTransferConstants.java
```diff
@@ -0,0 +1,6 @@
+package org.datatransferproject.datatransfer.generic;
+
+final class GenericTransferConstants {
+  public static final String SCHEMA_SOURCE_BASE =
+      "https://github.com/dtinit/data-transfer-project/blob/master";
```
sparuvu, line 5: I am a little confused by this. 

It seems like we are using this only in the tests to compare what is getting serialized. If that is the case, we should compare what is currently at the runtime but not what is currently in the master branch. There is 100% chance that someone might be running a little older version of generic importers compared to master and that would break the tests. Instead, we should just read the file from classpath and do what we are doing.

extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/GenericTransferExtension.java
```diff
@@ -0,0 +1,214 @@
...
+
+  public DataVertical getVertical() {
+    return vertical;
+  }
+}
+
+class GenericTransferServiceConfig {
+  private final String serviceId;
+  private final URL endpoint;
+  private final List<GenericTransferServiceVerticalConfig> verticals;
```
sparuvu, line 54: This can be a set. There is no need to maintain the order here.

extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/GenericTransferExtension.java
```diff
@@ -0,0 +1,214 @@
...
+    return verticals;
+  }
+
+  public boolean supportsVertical(DataVertical vertical) {
+    for (GenericTransferServiceVerticalConfig verticalConfig : verticals) {
+      if (verticalConfig.getVertical().equals(vertical)) {
+        return true;
+      }
+    }
+    return false;
```
sparuvu, line 84: nitpick: Probably it's worth building a set of Verticals from `verticals` attribute and do a `Set.contains`

extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/GenericTransferExtension.java
```diff
@@ -0,0 +1,214 @@
...
+              format("%s_KEY", serviceNameUpper), format("%s_SECRET", serviceNameUpper));
+    } catch (IOException e) {
+      throw new RuntimeException(
+          format(
+              "Failed to get application credentials for %s (%s)",
+              serviceNameUpper, serviceConfig.getServiceId()),
+          e);
+    }
+
+    if (serviceConfig.supportsVertical(BLOBS)) {
```
sparuvu, line 136: For better readability, it's worth extracting these conditions to their private methods. 

Even better,  if we can auto-update the `importerMap` these by iterating over the `DataVertical` entries. We may need to maintain a map of  DataVertical -> corresponding Serializer.

extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/GenericImporter.java
lisad: We should be OK with a 200 response code.  That's a more normal response to a POST request anyway, and it indicates success without promising that a HTTP resource was created.  A 201 response code comes with expectations that the responder will include Location headers, which may not be appropriate depending on what the responder has done with the content sent.  From the perspective of the sender/exporter being DTP, either response code is considered a success.

extensions/data-transfer/portability-data-transfer-generic/README.md
```diff
@@ -0,0 +1,694 @@
...
+Connection: Keep-Alive
+Accept-Encoding: gzip
+User-Agent: okhttp/3.9.1
+
+refresh_token=refreshToken&client_id=appKey&client_secret=appSecret&grant_type=refresh_token
+```
+
+The token refresh endpoint should return a response containing the access token. If a new refresh_token is provided it will be used for future token refreshes.
+
+```http
```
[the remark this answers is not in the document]
calumcalder: I've added a note in 0d4bee31 that you should probably be using a library or third-party auth server, and figured the refresh token API is documented well enough elsewhere that I've removed these details

extensions/data-transfer/portability-data-transfer-generic/README.md
```diff
@@ -0,0 +1,694 @@
+# Generic Importers
+
+## Implementing a Server
+
+To consume data from a generic importer, create a HTTP service in the language of your choice.
+The service should accept POST requests on a given base URL, and kebab-case sub-path specific to each data vertical (`https://example.com/import/blobs`, `https://example.com/import/social-posts`).
+
+### HTTP Interface
+
+Data will be POSTed to your endpoint with a `Content-Type` of either `application/json` for basic data-types, or `multipart/related` for file-based data-types. See below for how to interpret each of these.
```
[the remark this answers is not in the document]
calumcalder: Added what ended up spiralling in to a lot of details (mostly what we discussed in the maintainers meeting) in 0d4bee31 - hopefully that covers the spirit of what you suggested here

extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/GenericTransferExtension.java
```diff
@@ -0,0 +1,214 @@
...
+          CALENDAR,
+          new GenericImporter<CalendarContainerResource, CalendarSerializer.ExportData>(
+              CalendarSerializer::serialize,
+              appCredentials,
+              urlAppend(serviceConfig.getEndpoint(), "calendar"),
+              context.getMonitor()));
+    }
+  }
+
+  private URL urlAppend(URL base, String suffix) {
```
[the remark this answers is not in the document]
calumcalder: Sure is! 🙂 
I think because most other extensions use SDKs they don't have to deal with URLs manually, but definitely could be moved to a utility class if we spot it being reimplemented elsewhere

extensions/data-transfer/portability-data-transfer-generic/src/test/java/org/datatransferproject/datatransfer/generic/GenericImportSerializerTestBase.java
calumcalder: Agreed, this is another one where the python equivalent is much simpler!
Mutli-line string literals weren't added until Java 15 (13 with a compiler flag), and we're still on 11 so annoyingly can't utilise those and I think this is the best we can do for embedding large JSON strings in java files.
Looking at other tests, the imgur importer tests store [expected JSON data in resource files](https://github.com/dtinit/data-transfer-project/blob/6705a21f3ae691cc5ff6abd9c781717448074da4/extensions/data-transfer/portability-data-transfer-imgur/src/test/resources/all_images.json#L4) which would be another option, but it does mean the data doesn't live in the test file itself

I could be convinced either way on this, and I'd be open to other suggestions if anyone has them

portability-types-common/src/main/java/org/datatransferproject/types/common/models/calendar/RecurrenceRule.java
calumcalder: Thanks! Removed

extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/GenericImporter.java
calumcalder: It would probably be safe to refresh in most cases, but there are other reasons to send 401, e.g. missing token, invalid token format. In those cases the following token refresh request is likely to fail, which could mask the underlying issue slightly when debugging. Very minor, and I think you're right that it's likely the only reason for a 401 in most implementations, but this is one area I wanted the behaviour to be specifically defined

See https://github.com/dtinit/data-transfer-project/pull/1413 where this was fixed recently for the Microsoft extension which was just checking 401 - maybe @jzacsh can weigh in

extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/GenericImporter.java
calumcalder: Casting a big net for success responses was intentional, to try to follow the robustness principle. Agreed that the docs don't need to be so prescriptive, so I've updated them 👍

extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/GenericTransferExtension.java
calumcalder: I'm not sure I understand what you mean by extracting to private methods - could you add an example?

For auto-updating the importer map, we'd need a map of DataVertical -> GenericImporter, because sometimes we need GenericFileImporter and sometimes just GenericImporter (although they could be merged).
I gave it a go with BLOBS and CALENDAR in https://github.com/dtinit/data-transfer-project/commit/19db8804f9910bbe855b90525cbe3a621b8211a0 and I don't believe it's a significant improvement in readability, plus it's going to get messier when trying to tackle the PHOTOS/VIDEOS vs MEDIA logic, which is straightforward with the longer approach

extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/GenericTransferConstants.java
calumcalder: It gets used when building the schema source string to go in `GenericPayload`, in each `Serializer` class. That ends up in the payload to the importer endpoint.

It could go out of date, but I don't anticipate the code moving very often and would hope instances catch up relatively quickly. I did try to get it generated at runtime or compile time, but I couldn't find a way to get it to point to the right place (the path ended up in a build directory that doesn't exist in the repo) - if you know a way to do it that would be helpful!

One idea I had here was to switch to generating the POJOs from a JSON schema at build time, at which point the schema source could point to that file and make a bit more sense

calumcalder:
Thanks for the reviews!

lisad (approved):

sparuvu (approved):

extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/GenericTransferExtension.java
sparuvu: The length of this method increased and will continue to increase as we add more verticals. So, extracting to private methods that mutates the importer map would reduce the length of this method and makes it more readable. 

We can fix this later as well. This shouldn’t be a blocker for merging this PR.

Issue #1397, opened 2024-11-23
upgrade gradle: v6 is at end of life 


jzacsh:
https://endoflife.date/gradle probably worth spending a bit of time to test and make sure we can upgrade. and if we're gonna do that, we might as well jump to the latest (8.x)

jzacsh:
(overloading this issue) should probably also upgrade codeql (also past EOL) per steps in https://github.blog/changelog/2023-01-18-code-scanning-codeql-action-v1-is-now-deprecated/#what-do-i-need-to-change-in-my-workflow

jzacsh:
apparently (5 year old PR) this can be somewhat automated: https://github.com/dtinit/data-transfer-project/pull/936

FYI for whoever tackles this issue going forward

jzacsh:
once this is done, we can probably cleanup (or see if newer gradle versions can help point out unused deps) obsolete deps like log4j, per #1031

PR #1400, opened 2024-12-05
fix: EOL'd github actions: upgrade all that are possible to upgrade

jzacsh:
```sh
$ find .github/ -type f -exec \
  sed -Ei 's,actions?\/(.*)\@v[[:digit:]],actions/\1\@v3,g' {} \;
```

unsure if any action names changed (so unsure this will work), but guessing ci/cd will tell us.

part part of #1397


seehamrun:
Also related: https://github.com/dtinit/data-transfer-project/issues/1022

seehamrun (approved):

PR #1401, opened 2024-12-05
fix broken request/response management in Microsoft adapter

jzacsh:
fix buggy API requests we make by deleting/de-duping majority of
request/response handling from business logic.

**tl;dr of the below list**: now all the main functions in this class simply
construct a request and pass it to the same exact helper codepaths: "try with
creds-autorefresh" and "auto-extract json fields for me". This means all the
error edges cases are handled in a uniform way and we don't have to debug why
some of the codepaths are erroring in different ways. The only differences now
are: what is the "context" message you want included when/if an error condition
is bubbled up.

a nitty-gritty breakdown of functional changes here:

-   functional: when Microsoft server response indicates it's needed, *always*
    refresh creds now and auto-retry the failed response.
    -   previously only happened sometimes, and happened differently
    -   this is what the extant `tryWithCreds()` helper was trying to enable
        (now moved to the bottom of file)
    -   this is what `tryWithCreds*()` helpers are all doing now (just slightly
        different use-cases, but identical codepaths now)
-   functional: streaming chunks was ignoring permission-denied server errors
    (see `uploadChunk` diff)
-   nit/functional: always close our OkHTTP3 `Closeable`
    -   previously only happened sometimes
    -   this is what `sendMicrosoftReques()` does
-   nit/functional: always check error-conditions when unwrapping JSON responses
    -   previously this happened differently in every request path, sometimes
        checking body's own presence, sometimes now
    -   this is what `tryWithCredsOrFail(..., jsonResponseKey, ...)` gets us
    -   this is enabled by the new `MicrosoftApiResponse#getJsonValue()` API
-   nit/functional: always show full details of the request when we're throwing
    an exception
    -   previously error/cause strings were being built again and again by-hand
    -   this is what `tryWithCredsOrFail(...., causeMessage)` gets us
    -   this is enabled by the new `MicrosoftApiResponse#toIoExcepion()` API (as
        well as `MicrosoftApiResponse#throwDtpException(cause)`)
    -   this is also just a very basic AutoValue-win we get from
        `MicrosoftApiResponse#toString()`
-   noop(readability): it's just much easier to reason about where the real
    "business logic" is now: see how small the implementations of the original
    functions are:
    -   `createOneDriveFolder`
    -   `importDownloadableItem`
    -   `uploadStreamInChunks`
    -   `uploadChunk`
    -   `createUploadSession`
    -   ... everything below these are just re-used helpers that unwrap
        requests/responses

wrt Photos/Videos forks of this codebase: at this point my previous strategy of
hand-syncing the changes isn't worth while. I'm going to send a separate PR to
finally just resolve the [TODO] in that file and refactor all three adapters so
they're all using `MicrosoftMedia*` codebase's generics-approach.

[TODO]: https://github.com/dtinit/data-transfer-project/blob/ef14eb62a8b946dce67cd6247add58fdb4c8ca13/extensions/data-transfer/portability-data-transfer-microsoft/src/main/java/org/datatransferproject/transfer/microsoft/photos/MicrosoftPhotosImporter.java#L56


portability-api/.attach_pid4035535
seehamrun, line 1: does this need to be deleted?

seehamrun (approved):

portability-api/.attach_pid4035535
jzacsh: good catch. done

seehamrun (approved):

PR #1419, opened 2024-12-25
Line

oumz11:


calumcalder:
Hi @oumz11 - thanks for opening this PR. It looks like you're trying to set up automated publishing for this repo, which we happen to already be working on! (see #1420, #1421, #1423)

Your proposal seems to be based on Github's docs https://github.com/actions/setup-java/blob/main/docs/advanced-usage.md#publishing-using-gradle which doesn't quite map to our set-up; in particular, we publish to maven central and use conventional-commits to generate the changelog and version. You can see some more details in the linked PRs.
With that in mind, I'm going to close out this PR since the work for setting up automated publishing is mostly done in those other PRs. If there's a requirement to also publish to github's package repository, feel free to open an issue and we can look at setting that up.

Issue #1422, opened 2025-01-06
Create Maven credential that can be used to automate release processes with GitHub Actions


lisad:
Some companies use the DTP artifact registered in the Maven repository (https://mvnrepository.com/) in their deploy processes, but the release to the Maven repository has not been automated.

One step required for that is to get a Maven credential that can be used in GitHub Actions .  I'll get @jzacsh or @holachuy and @calumcalder's help with setting that up and applying it.

lisad:
I've created the account "LisaDusseault" with my DTI email address. Next steps @jzacsh or @holachuy  ?

wmorland:
@lisad I'm submitting a request to add you to the namespace via email with Central Support (as described in the instructions at https://central.sonatype.org/register/legacy/#managing-your-project )

wmorland:
Hi @lisad, that should now be granted

lisad:
Completed - added OSSRH_USERNAME and OSSRH_PASSWORD secrets to this repo

PR #1423, opened 2025-01-09
ci: release to maven on merge

calumcalder:
Creates a new github action pipeline that runs on merge to `master`. The pipeline will:
1. Create a new commit with an updated CHANGELOG.md
2. Tag that commit and push it to `master`
3. Make a clean checkout of that commit
4. Build, sign, and release all java artefacts to maven

Depends on #1422 to set up the secrets in github

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,52 @@
+name: Release
+
+permissions:
+  # The commitizen action pushes a new commit to the main
+  # branch to generate the changelog + tag, so needs write
+  # permission
+  contents: write
+
+on:
+  push:
+    branches:
+      - 'master'
+
+jobs:
+  bump_version:
+    if: "!startsWith(github.event.head_commit.message, 'bump:')"
+    runs-on: ubuntu-latest
+    name: "Bump version"
+    outputs:
+      version: ${{ steps.cz.outputs.version }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0  # Fetch tags, which are required to calculate the new version
+          token: "${{ secrets.GITHUB_TOKEN }}"
+      - id: cz
+        name: "Generate Changelog and Tag"
+        uses: commitizen-tools/commitizen-action@master
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+  release:
+    needs: bump_version
+    runs-on: ubuntu-latest
+    name: "Release"
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ needs.bump_version.outputs.version }}
+      - name: "Set up JDK"
+        uses: actions/setup-java@v4
+        with:
+          java-version: '11'
+          distribution: 'adopt'
+          cache: gradle
+      - name: "Sign and Release"
+        env:
+          GRADLE_SIGNING_KEY: "${{ secrets.GRADLE_SIGNING_KEY }}"
+          GRADLE_SIGNING_PASSWORD: "${{ secrets.GRADLE_SIGNING_PASSWORD }}"
+          OSSRH_USERNAME: "${{ secrets.OSSRH_USERNAME }}"
+          OSSRH_PASSWORD: "${{ secrets.OSSRH_PASSWORD }}"
+        # Exclude client-rest as it's not part of the java release
+        run: ./gradlew clean build sign uploadArchives --exclude-task :client-rest:uploadArchives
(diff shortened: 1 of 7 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,52 @@
...
+
+on:
+  push:
+    branches:
+      - 'master'
+
+jobs:
+  bump_version:
+    if: "!startsWith(github.event.head_commit.message, 'bump:')"
+    runs-on: ubuntu-latest
```
jzacsh, line 17: is it worth updating some docs within this same PR? or just wait til after things are settled/working?


- [ ] somewhere to mention expected commit-message norms when mergers squash their commits and craft that final message? (maybe same doc as below, maybe new doc?) 
- [ ] to mention the automated process https://github.com/dtinit/data-transfer-project/blob/master/Documentation/Publishing.md

jzacsh (commented):
looks good (obviously pending the secrets as you said). left you a question on the YAML startswith.

.github/workflows/release.yml
calumcalder: Ah yeah I had meant to update the docs - I'll do it in this PR.
For the commit message rules I updated [CONTRIBUTING.md](https://github.com/dtinit/data-transfer-project/blob/master/CONTRIBUTING.md#pr-titles) in #1420, but happy to add it elsewhere if there's somewhere more obvious

jzacsh: Huh, interestingly I already hit a weird edge-case here: github's default behavior (when you click "squash and commit") is to take the *old*/no-longer-visible title of the PR, and use that as the commit message.

So in my case [3eae3298b1f4263689661ea7ee8ac5a464b90673](https://github.com/dtinit/data-transfer-project/commit/3eae3298b1f4263689661ea7ee8ac5a464b90673) got in with a bad message because
1. PR #1425 was missing a colon (`:`) --> CI/CD warned me during PR review --> I fixed it (edited my PR title)
1. days later I squashed/merged (forgetting about that^ detail)
1. turns out the auto-populated form-field in step 2 didn't have the colon :(

jzacsh: super-optional nit based on my PR (step 1 [error message I saw](https://github.com/dtinit/data-transfer-project/actions/runs/12834848865/job/35793019492)): maybe modify the linter so it provides a link directly to your new doc you're landing in this PR? ie: link to https://github.com/dtinit/data-transfer-project/blob/master/Documentation/Publishing.md#semantic-versioning-and-conventional-commits ([preview](https://github.com/dtinit/data-transfer-project/blob/94641e1307dc756e23e9c042f201c6c06f67434a/Documentation/Publishing.md#semantic-versioning-and-conventional-commits)) directly from the runner (this is a bit ugly, but your call):

```yaml
jobs:
  commitlint:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - name: "Install commitizen"
        run: pip install --user -U commitizen
      - name: "Lint the PR title"
        run: cz check --message "${{ github.event.pull_request.title }}" || { echo "see https://github.com/dtinit/data-transfer-project/blob/master/Documentation/Publishing.md#semantic-versioning-and-conventional-commits" >&2; exit 1; }
```


(maybe CONTRIBUTING.md "PR Titles" section could also just link directly to this new Publish.md section in hopes they stay in sync)

jzacsh (approved):
I see this is still in draft, but LGTM already. (despite our one basically-resolved comment thread on docs)

.github/workflows/release.yml
calumcalder: > Huh, interestingly I already hit a weird edge-case here: github's default behavior (when you click "squash and commit") is to take the old/no-longer-visible title of the PR, and use that as the commit message.

Looks like this is the default behaviour for pull requests with a single commit - I've updated the settings so it always uses the PR title (Settings -> Pull Requests section -> Default commit message)

calumcalder: > maybe modify the linter so it provides a link directly to your new doc you're landing in this PR? 

Yep, good idea - thanks!

calumcalder:
Rebased

.github/workflows/release.yml
jzacsh: > Looks like this is the default behaviour for pull requests with a single commit - I've updated the settings so it always uses the PR title (Settings -> Pull Requests section -> Default commit message)

I guess just an unavoidable FYI for us then: when a PR author gets this lint, the time to fix it is _not really_ right away (I mean, yes right-away to make the lint disappear), but at the moment they squash their commit. :\

jzacsh (approved):

PR #1429, opened 2025-01-30
Update ContinuationDataTest.java

Rian-Ismael:
Improves the readability of the test. The assertions are clearer and include descriptive messages, making it easier to identify which assertion fails. This enhances code maintainability, understanding and debugging efficiency.


lisad:
We discussed and don't think that the error assertions really add anything that isn't already visible in the code.

PR #1430, opened 2025-01-31
feat: Destination Full support for Generic Importers 

### LINKED ISSUE #1424: Generic Importers - Destination Full Support
## Background

Importers can signal failure for specific reasons with various subclasses of `CopyExceptionWithFailureReason`, some of which are treated as restartable errors. For example, `InvalidTokenException` and `DestinationMemoryFullException` signal a token expiry (which requires reauth and a restart) and a full destination (which requires the user to create space and restart) respectively.

Generic Importers (added in #1389) supports throwing `InvalidTokenException` on a 401 response with an `invalid_token` error from the API, but does not support `DestinationMemoryFullException`.

## Proposal

To support destinations using Generic Importers creating a `DestinationMemoryFullException`, the API could be extended to define another error code and error name combination - e.g. 400 and `destination_full`. The Generic Importers code would then need to be extended to support handling this case.

ameya9:
Fixes #1424.

Generic importers will throw `DestinationMemoryFullException`(which extends `CopyExceptionWithFailureReason`) when any of the import post API implemented by client throws error with code: `413` & error: `destination_full`

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-generic/README.md
+++ b/extensions/data-transfer/portability-data-transfer-generic/README.md
@@ -753,3 +753,4 @@ Content-Type: application/json
 }
 ```
 The worker will pause the job and won't retry in such scenario.
+Some exporters also support the ability to retry transfers after user frees up space in the destination.
--- a/extensions/data-transfer/portability-data-transfer-generic/README.md
+++ b/extensions/data-transfer/portability-data-transfer-generic/README.md
@@ -709,7 +709,7 @@ If there is an error importing an item on any endpoint, the relevant HTTP respon
 }
 ```
 
-The combination of the HTTP response code and `error` field can be used to encode for specific failure modes; see [Token Refresh](#token-refresh) below.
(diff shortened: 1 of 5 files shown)


extensions/data-transfer/portability-data-transfer-generic/README.md
```diff
@@ -734,3 +734,22 @@ Content-Type: application/json
...
+
+```
+HTTP/1.1 413 Payload Too Large
+Content-Type: application/json
+{
+  "error": "destination_full",
+  "error_description": "The destination storage is full"
+}
+```
+The worker will pause the job and won't retry in such scenario.
```
calumcalder, line 755: Could also add that some deployments allow a user to retry the transfer after making space in the destination

calumcalder (dismissed):

extensions/data-transfer/portability-data-transfer-generic/README.md
ameya9: Added

PR #1434, opened 2025-02-14
ci: release without committing to main branch

calumcalder:
Committing to the master branch isn't allowed from github actions because of our branch protection rules. There's no easy way to allow github actions to bypass this.

As a workaround, skip all the parts of the release that needed to commit to the repo. This means moving the version config to an environment variable, which shouldn't be an issue, and using git tags as the source of truth.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,17 +17,22 @@ jobs:
     runs-on: ubuntu-latest
     name: "Bump version"
     outputs:
-      version: ${{ steps.cz.outputs.version }}
+      version: ${{ steps.tag.outputs.version }}
     steps:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0  # Fetch tags, which are required to calculate the new version
           token: "${{ secrets.GITHUB_TOKEN }}"
-      - id: cz
-        name: "Generate Changelog and Tag"
-        uses: commitizen-tools/commitizen-action@master
-        with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
+      - name: "Install commitizen"
+        run: pip install --user -U commitizen
+      - id: tag
+        name: "Tag release"
+        run: |
+          VERSION=$(cz bump --get-next)
+          TAG="v${VERSION}"
+          git tag "${TAG}"
+          git push origin "${TAG}" --tags
+          echo "version=${VERSION}" >> "$GITHUB_OUTPUT"
   release:
     needs: bump_version
     runs-on: ubuntu-latest
@@ -35,7 +40,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          ref: ${{ needs.bump_version.outputs.version }}
+          ref: "v${{ needs.bump_version.outputs.version }}"
       - name: "Set up JDK"
         uses: actions/setup-java@v4
         with:
@@ -48,5 +53,6 @@ jobs:
           GRADLE_SIGNING_PASSWORD: "${{ secrets.GRADLE_SIGNING_PASSWORD }}"
           OSSRH_USERNAME: "${{ secrets.OSSRH_USERNAME }}"
           OSSRH_PASSWORD: "${{ secrets.OSSRH_PASSWORD }}"
+          RELEASE_VERSION: "${{ needs.bump_version.outputs.version }}"
         # Exclude client-rest as it's not part of the java release
         run: ./gradlew clean build sign uploadArchives --exclude-task :client-rest:uploadArchives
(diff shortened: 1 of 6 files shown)


.github/workflows/release.yml
```diff
@@ -17,25 +17,30 @@ jobs:
...
         with:
           fetch-depth: 0  # Fetch tags, which are required to calculate the new version
           token: "${{ secrets.GITHUB_TOKEN }}"
-      - id: cz
-        name: "Generate Changelog and Tag"
-        uses: commitizen-tools/commitizen-action@master
-        with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
+      - name: "Install commitizen"
+        run: pip install --user -U commitizen
```
jzacsh, line 27: optional: wearing my supply-chain-debugging hat

could we add a line of --version output or something, so we have historical data on what version of `commitizen` is working for us (in case we're debugging a break int he future)

```suggestion
        run: |
          pip install --user -U commitizen &&
            cz version --verbose
```

([looks like it](https://github.com/commitizen-tools/commitizen/blob/43ee8a06b2ff003f45e8975cf9487a802cdf6d89/commitizen/commands/version.py#L31) self-reports its name, so don't need to add our own echo or anything to explain in our own  logs)

Documentation/Publishing.md
```diff
@@ -14,7 +14,7 @@ The action contains two jobs; one to bump the version number of the DTP packages
 
 DTP uses [Semantic Versioning](https://semver.org/) for published packages. We also enforce [Conventional Commits](https://conventionalcommits.org/) on the `master` branch through the `.github/workflows/commitlint.yml` Github action, which lets us automatically calculate version numbers.
 
-Commitizen is a tool that supports automatically incrementing SemVer version numbers based on git commit history. Our usage of Commitizen is configured in `.cz.toml`, and uses the `commitizen-tools/commitizen-action` Github action to automatically bump the package version number and to tag the new version in git.
+Commitizen is a tool that supports automatically incrementing SemVer version numbers based on git commit history. Our usage of Commitizen is configured in `.cz.toml`, and is used in a Github action to automatically tag the new version in git.
```
jzacsh, line 17:
```suggestion
Commitizen is a tool that supports automatically incrementing SemVer version numbers based on git commit history. Our usage of Commitizen is configured in `.cz.toml`, and is used in a Github action to automatically tag the new version in git and publish to maven (per "automated publishing" section of this doc).
```

jzacsh (changes requested):
thanks for figuring this all out! Feel free to ack/ignore all these since they're all optional!

calumcalder:
Thanks for the suggestions! Applied them all

.github/workflows/release.yml
```diff
@@ -24,15 +24,17 @@ jobs:
...
-          VERSION=$(cz bump --get-next)
-          TAG="v${VERSION}"
-          git tag "${TAG}"
-          git push origin "${TAG}" --tags
-          echo "version=${VERSION}" >> "$GITHUB_OUTPUT"
+          NEW_VERSION=$(cz bump --get-next) || exit 1
+          TAG="v${NEW_VERSION}"
+          git tag "${TAG}" &&
+            git push origin "${TAG}" --tags &&
+            echo "version=${NEW_VERSION}" | tee "$GITHUB_OUTPUT"
```
jzacsh, line 37: I see you switched to tee, but specifically away from an append to an overwrite. (The new version is equivalent to if you had switched from the `>>` you had to `>`, clobbering any prior state of the output file).

I don't _think_ this matters (idk what the GitHub docs say about that magic file) but if you want to append still then just use `-a` flag (`tee -a`)

jzacsh (dismissed):

.github/workflows/release.yml
calumcalder: Ah good catch - all the github docs use append, I assume in-case you're populating multiple outputs which we're not doing here, but it's probably the safer choice

jzacsh (approved):

jzacsh:
Hey @calumcalder  if I understand [this latest tag](https://github.com/dtinit/data-transfer-project/releases/tag/v1.1.0) correctly, then it:

1. was autogenerated from our new code here
2. bumped us v1.0.4 --> v1.1.0

Is that kind of bump what we want? Or would we expect just a patch bump for a commit like this? (ie: do we want it to have done `v1.0.4 --> v.1.0.5`?). If we just want patch, then looks like we just need to tweak our `bump` CLI call to [pass `--increment-mode PATCH` flag](https://commitizen-tools.github.io/commitizen/commands/bump/#-increment-mode):

```diff
-          NEW_VERSION=$(cz bump --get-next) || exit 1
+          NEW_VERSION=$(cz bump --get-next --increment-mode PATCH) || exit 1
```

PR #1449, opened 2025-04-02
fix(generic): don't fail on unknown properties while deserialising RefreshTokenResponse

ameya9:
don't fail on unknown properties while deserialising RefreshTokenResponse


calumcalder (approved):

extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/auth/OAuthTokenManager.java
```diff
@@ -157,6 +159,7 @@ public <T, Ex extends Exception> T withAuthData(FunctionRequiringAuthData<T, Ex>
       return f.execute(authData);
     } catch (InvalidTokenException e) {
       if (authData.getRefreshToken() == null || authData.getRefreshToken().isEmpty()) {
+        monitor.severe(() -> "Refresh token not present with auth data");
```
calumcalder, line 162: nit: we're throwing here already, so this will just make the error appear twice

ameya9: We throw Invalid token exception from multiple places, I thought it would be good to know if we are throwing it because refresh token is missing. 
This did happen with Fabric, as their OAuth response wasn't sending refresh token in one of the iteration

PR #1450, opened 2025-04-07
fix(generic): Fix Blobs serializer  

ameya9:
Changes in diff

1. Using job store to save ID to path mapping 
2. Blobs serializer only parses the tree single depth, it uses job store to link parent & subfolders across different iterations

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-generic/build.gradle
+++ b/extensions/data-transfer/portability-data-transfer-generic/build.gradle
@@ -21,11 +21,14 @@ plugins {
 
 dependencies {
     compile project(':portability-spi-transfer')
+    compile project(':portability-transfer')
     compile project(':portability-spi-cloud')
     compile project(':portability-types-common')
     compile "com.squareup.okhttp3:okhttp:${okHttpVersion}"
     compile "com.fasterxml.jackson.datatype:jackson-datatype-jsr310:${jacksonVersion}"
     compile "com.fasterxml.jackson.datatype:jackson-datatype-jdk8:${jacksonVersion}"
+    testCompile project(':portability-transfer').sourceSets.test.output
+    testCompile project(':extensions:cloud:portability-cloud-local')
     testCompile "com.squareup.okhttp3:mockwebserver:${okHttpVersion}"
     testCompile "commons-fileupload:commons-fileupload:1.5"
 }
--- a/extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/BlobIdToName.java
+++ b/extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/BlobIdToName.java
@@ -1,0 +1,22 @@
+package org.datatransferproject.datatransfer.generic;
+
+import java.util.HashMap;
+import java.util.Map;
+import org.datatransferproject.types.common.models.DataModel;
+
+public class BlobIdToName extends DataModel {
+
+  private Map<String, String> idToName;
+
+  public BlobIdToName() {
+    this.idToName = new HashMap<>();
+  }
+
+  public String get(String id) {
+    return idToName.get(id);
+  }
+
+  public String add(String id, String name) {
+    return idToName.put(id, name);
+  }
+}
--- a/extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/BlobbySerializer.java
+++ b/extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/BlobbySerializer.java
@@ -7,12 +7,18 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonSubTypes;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import com.fasterxml.jackson.annotation.JsonTypeName;
+import java.io.IOException;
 import java.time.ZonedDateTime;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 import java.util.Queue;
+import java.util.UUID;
+import org.datatransferproject.spi.cloud.storage.JobStore;
+import org.datatransferproject.spi.cloud.storage.TemporaryPerJobDataStore;
+import org.datatransferproject.spi.transfer.hooks.JobHooks;
+import org.datatransferproject.transfer.JobMetadata;
 import org.datatransferproject.types.common.DownloadableItem;
 import org.datatransferproject.types.common.models.blob.BlobbyStorageContainerResource;
 import org.datatransferproject.types.common.models.blob.DigitalDocumentWrapper;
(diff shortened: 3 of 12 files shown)


extensions/data-transfer/portability-data-transfer-generic/src/test/java/org/datatransferproject/datatransfer/generic/CalendarSerializerTest.java
```diff
@@ -18,8 +19,8 @@ public class CalendarSerializerTest extends GenericImportSerializerTestBase {
   public void testCalendarSerializer() throws Exception {
     CalendarContainerResource container =
         new CalendarContainerResource(
-            Arrays.asList(new CalendarModel("calendar123", "Calendar 123", "Calendar description")),
-            Arrays.asList(
```
ameya9, line 22: reformatting

extensions/data-transfer/portability-data-transfer-generic/src/test/java/org/datatransferproject/datatransfer/generic/BlobbySerializerTest.java
```diff
@@ -22,12 +31,12 @@ public void testBlobbySerializerFolders() throws Exception {
             "root",
             "rootdir",
             new ArrayList<>(),
-            Arrays.asList(
+            List.of(
                 new BlobbyStorageContainerResource(
```
ameya9, line 35: Mostly reformatting

extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/BlobbySerializer.java
```diff
@@ -106,72 +109,114 @@ public String getPath() {
...
 
-    public String getPath() {
-      return path;
+  private void initialiseBlobIdToNameIfNot(UUID jobId) {
+    if (blobIdToName == null) {
+      try {
+        blobIdToName = jobStore.findData(jobId, BLOB_ID_TO_NAME_KEY, BlobIdToName.class);
+      } catch (IOException e) {
+        monitor.severe(() -> "Not able to find blobIdToNameKey map from JobStore", e);
+        throw new RuntimeException(e);
```
calumcalder, line 152: You could let the IOException propagate and use the Copier's retry logic to try again - might need to update the `serializer` interface

extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/BlobIdToName.java
```diff
@@ -0,0 +1,22 @@
+package org.datatransferproject.datatransfer.generic;
+
+import java.util.HashMap;
+import java.util.Map;
+import org.datatransferproject.types.common.models.DataModel;
+
+public class BlobIdToName extends DataModel {
+
+  private Map<String, String> idToName;
```
calumcalder, line 9: I guess this is a map getting stored inside a KV store so you don't have to write it as often as if you stored the id -> path mapping in the KV store directly?
Have you thought about the expected size of the map?

extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/BlobbySerializer.java
```diff
@@ -106,72 +109,114 @@ public String getPath() {
...
-  public interface ExportData {}
 
-  static class BlobbyContainerPath {
-    private BlobbyStorageContainerResource container;
-    private String path;
+  static final String SCHEMA_SOURCE =
+      GenericTransferConstants.SCHEMA_SOURCE_BASE
+          + "/extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/BlobbySerializer.java";
+  private static final String BLOB_ID_TO_NAME_KEY = "blobIdToNameKey";
+  private final JobStore jobStore;
```
calumcalder, line 117: I think the only functions being used on `jobStore` are from `TemporaryPerJobDataStore`, so you could depend on that interface instead

extensions/data-transfer/portability-data-transfer-generic/build.gradle
```diff
@@ -21,11 +21,14 @@ plugins {
 
 dependencies {
     compile project(':portability-spi-transfer')
+    compile project(':portability-transfer')
```
calumcalder, line 24: Do you need all of `portability-transfer` or is `portability-spi-cloud` sufficient?

extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/BlobbySerializer.java
```diff
@@ -106,72 +109,114 @@ public String getPath() {
...
 
+    saveStateToStore();
     return results;
   }
+
+  @JsonSubTypes({
+    @JsonSubTypes.Type(FolderExportData.class),
+    @JsonSubTypes.Type(FileExportData.class),
+  })
+  public interface ExportData {}
```
calumcalder, line 221: #nit a bit of reformatting noise here and in other files - in future can we keep that in a separate commit or PR so it's easier to review?

calumcalder (commented):
Generally LGTM, some comments and questions in-line

extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/BlobbySerializer.java
ameya9: done

extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/BlobbySerializer.java
ameya9: done

extensions/data-transfer/portability-data-transfer-generic/build.gradle
ameya9: yeah, I need portability-transfer for JobMetadata & portability-spi-cloud for Jobstore

extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/BlobbySerializer.java
ameya9: Sorry, It's my IntelIj, doing refactoring as I touch the files

extensions/data-transfer/portability-data-transfer-generic/src/main/java/org/datatransferproject/datatransfer/generic/BlobIdToName.java
ameya9: I did not. I thought since it's just folders, it should be okay. 

But If I do rough calculation. 
If we have 1000 folders, each entry < 100 bytes (assuming max depth less than 5) -> 10^5 bytes -> 0.1MB 

10^4 folders -> 1MB
10^5 folders -> 10MB

It should definitely work for structured data BLOBS data, but might not work when exporter is google drive or dropbox. 
Happy to change it to per key Model

calumcalder: That sounds ok to me 👍 can always change in the future if it causes issues

calumcalder (approved):

PR #1469, opened 2025-08-27
feat: google music release exporter

Shashank1116:
Enables the support for exporting music releases from Google.

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/music/GoogleMusicExporter.java
+++ b/extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/music/GoogleMusicExporter.java
@@ -32,6 +32,7 @@ import java.util.UUID;
 import javax.annotation.Nullable;
 import org.datatransferproject.api.launcher.Monitor;
 import org.datatransferproject.datatransfer.google.common.GoogleCredentialFactory;
+import org.datatransferproject.datatransfer.google.musicModels.ExportReleaseResponse;
 import org.datatransferproject.datatransfer.google.musicModels.GoogleArtist;
 import org.datatransferproject.datatransfer.google.musicModels.GooglePlaylist;
 import org.datatransferproject.datatransfer.google.musicModels.GooglePlaylistItem;
@@ -98,8 +99,7 @@ public class GoogleMusicExporter implements Exporter<TokensAndUrlAuthData, Music
   public ExportResult<MusicContainerResource> export(
       UUID jobId, TokensAndUrlAuthData authData, Optional<ExportInformation> exportInformation)
       throws IOException, InvalidTokenException, PermissionDeniedException, ParseException {
-    // TODO: Remove the logic when testing is finished.
-    // Local demo-server test usage. Start transfer job without ExportInformation.
+    // Used in production
     if (!exportInformation.isPresent()) {
       StringPaginationToken paginationToken = new StringPaginationToken(PLAYLIST_TOKEN_PREFIX);
       return exportPlaylists(authData, Optional.of(paginationToken), jobId);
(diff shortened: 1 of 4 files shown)


extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/music/GoogleMusicExporter.java
```diff
@@ -237,6 +236,57 @@ ExportResult<MusicContainerResource> exportPlaylistItems(
...
+    ResultType resultType = ResultType.END;
+    if (!Strings.isNullOrEmpty(token)) {
+      nextPageData = new StringPaginationToken(RELEASE_TOKEN_PREFIX + token);
+      resultType = ResultType.CONTINUE;
+    }
+
+    ContinuationData continuationData = new ContinuationData(nextPageData);
+    MusicContainerResource containerResource = null;
+    GoogleRelease[] googleReleases = exportReleaseResponse.getReleases();
+    List<MusicRelease> exportReleases = new ArrayList<>();
```
karthiknsmz, line 257: The function is already named `exportReleases`. Shall we use `musicReleases` here instead?

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/music/GoogleMusicHttpApi.java
```diff
@@ -128,6 +131,20 @@ PlaylistItemExportResponse exportPlaylistItems(String playlistId, Optional<Strin
...
+  ExportReleaseResponse exportReleases(Optional<String> pageToken)
+      throws InvalidTokenException, PermissionDeniedException, IOException {
+
+    Map<String, String> params = new LinkedHashMap<>();
+    params.put(PAGE_SIZE_KEY, String.valueOf(RELEASE_ITEM_PAGE_SIZE));
+    if (pageToken.isPresent()) {
+      params.put(TOKEN_KEY, pageToken.get());
+    }
+    return makeGetRequest(
+        RELEASE_BASE_URL + "releases", Optional.of(params),
```
karthiknsmz, line 143: Shouldn't `RELEASE_BASE_URL` have `releases` at the end of the url instead of adding them here separately?

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/music/GoogleMusicExporter.java
Shashank1116: I agree! we can definitely avoid the repeating names. What about `exportableReleases` ? It can add more context.

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/music/GoogleMusicHttpApi.java
Shashank1116: It was a bit confusing to me actually.

The current BASE_URL  for YTM is `https://youtubemediaconnect.googleapis.com/v1/users/me/musicLibrary` while it should be just `https://youtubemediaconnect.googleapis.com`. But to keep the scope of this feature limited, I created a separate one. But I agree the value is misleading.

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/music/GoogleMusicExporter.java
karthiknsmz: Sounds good. Thanks!

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/music/GoogleMusicHttpApi.java
karthiknsmz: Hmm.. Yeah, that makes sense. For the scope of this PR, just fixing the release base url should be enough. We can fix the base url in a separate one.

karthiknsmz:
I believe the linter is throwing error. Please fix it before merging the PR.

PR #1493, opened 2026-06-16
feat: add containerized build/test environment via Dockerfile

alexeyqu:
Adds a `Dockerfile` pinned to `gradle:8.10.2-jdk11` and a `docker-compose.yml` so contributors can run the full test suite with no local JDK installed:

```bash
docker compose run --rm test
```

The compose file encodes the source bind-mount and a named volume for the Gradle dependency cache, so the first run resolves dependencies and subsequent runs reuse them.

## Notable decisions / tradeoffs

- **Bind-mount design:** source is bind-mounted at run time rather than `COPY`ed into the image, so the image doesn't need rebuilding on every code change. The `docker-compose.yml` hides this from contributors -- they just run `docker compose run`.
- **`.gitignore` cleanup:** removed the bare `Dockerfile` rule. The rule's comment said it was for the demo-server's generated `Dockerfile`, but that file lands in `distributions/demo-server/build/demo/Dockerfile` -- already covered by the `build/` ignore rule. The bare rule was redundant.
- **Image name `datatransferproject/dev`:** mirrors the existing `datatransferproject/demo` image produced by the `:dockerize` task, establishing a consistent namespace.
- **JDK pinned to 11:** the Gradle wrapper (6.9.2) can't parse Java 17 bytecode when compiling build scripts -- `gradle:*-jdk17` fails outright. Not a preference; a hard constraint from the wrapper version.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,8 @@
+.git
+.gradle
+build
+out
+generated
+target
+node_modules
+client-rest/dist
--- a/.gitignore
+++ b/.gitignore
@@ -11,9 +11,6 @@ env.secrets
 distributions/*/resources/
 
 #**/src/main/resources/settings/*.yaml
-# Dockerfile is generated each time we run build.sh
-Dockerfile
-
 # compiled output
 build/
 out/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,18 @@
+# Runs the Java/Gradle build and test suite with no local JDK install.
+#
+# Source is bind-mounted at `docker run` time -- this image is not rebuilt
+# per code change. Gradle 6.9.2 (pulled by the wrapper regardless of this
+# image's bundled Gradle version) can't parse Java 17 bytecode when
+# compiling build.gradle/settings.gradle, so the JDK here is pinned to 11.
+#
+# The gradle:* image's default GRADLE_USER_HOME is /home/gradle/.gradle --
+# mount a named volume there to cache resolved dependencies across runs:
+#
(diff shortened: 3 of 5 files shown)


alexeyqu:
Looks like the Java CI failure is preexisting -- I reran the latest `master` branch and [it failed despite passing last month](https://github.com/dtinit/data-transfer-project/actions/runs/26077383377/job/81793753427) -- investigating...

alexeyqu:
Confirmed, looks like a flaky test -- see https://github.com/dtinit/data-transfer-project/actions/runs/26077383377 run history

lisad (approved):
additional documentation is appreciated!

alexeyqu:
cc @simonxander -- I am curious if this direction would simplify your work / Developer Experience / deployment setup. Let me know if you'd be up for a discussion here (probably via a call).

simonxander:
> cc @simonxander -- I am curious if this direction would simplify your work / Developer Experience / deployment setup. Let me know if you'd be up for a discussion here (probably via a call).

Hi @alexeyqu, thanks for the ping!

I have tested the `docker compose` setup, and it works perfectly. This is very helpful for me, especially when switching between different projects that need different Java environments. It also makes setting up the environment much faster for new people joining the project.

For further discussion, I prefer text-based communication. Please feel free to share your thoughts right here in the comments!

alexeyqu:
Thank you @simonxander -- I am continuing to improve the DevEx setup here, would greatly appreciate any ideas. Currently the next proposed step is some docker image tightening in #1505.

PR #1494, opened 2026-06-17
fix(synology): await producer thread shutdown before returning from uploadVideoChunks

alexeyqu:
`producerFuture.cancel(true)` requests interruption but returns immediately, leaving the producer thread alive after `createVideo` returns. On resource-constrained CI runners, this races with Mockito's post-test stub validation, causing `CreateMedia` VideoModel tests to flake as SKIPPED.

Added `producerFuture.get(5, TimeUnit.SECONDS)` after the cancel so the method only returns once the producer has actually stopped. `CancellationException` and timeout are swallowed -- both mean the producer is done or on its way out.

Fixes the flaky test (see [Gradle CI attempts on current master branch](https://github.com/dtinit/data-transfer-project/actions/runs/26077383377/attempts/4))

### DIFF
--- a/extensions/data-transfer/portability-data-transfer-synology/src/main/java/org/datatransferproject/datatransfer/synology/service/SynologyDTPService.java
+++ b/extensions/data-transfer/portability-data-transfer-synology/src/main/java/org/datatransferproject/datatransfer/synology/service/SynologyDTPService.java
@@ -307,6 +307,11 @@ public class SynologyDTPService {
     } finally {
       consumerAborted.set(true);
       producerFuture.cancel(true);
+      try {
+        producerFuture.get(5, TimeUnit.SECONDS);
+      } catch (Exception ignored) {
+        // Expected: CancellationException if cancelled, or timeout -- producer is stopping
+      }
     }
   }
 


simonxander:
The added `get(5, TimeUnit.SECONDS)` doesn't actually wait for the producer - it needs to come before `cancel()`.

According to the [`CompletableFuture.get(long, TimeUnit)`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/CompletableFuture.html#get(long,java.util.concurrent.TimeUnit)) docs , `get()` throws `CancellationException` if the future was cancelled. Since `producerFuture.cancel(true)` runs first and completes the future into the cancelled state immediately, the subsequent `get` throws `CancellationException` right away and gets swallowed — so it returns instantly without ever waiting for the producer thread to finish. The race condition still occurs even with these changes.

#### Suggested fix

```java
} finally {
  consumerAborted.set(true);
  try {
    producerFuture.get(5, TimeUnit.SECONDS);
  } catch (Exception ignored) {
    // Expected: CancellationException if cancelled, or timeout -- producer is stopping
  }
  producerFuture.cancel(true);
}
```

alexeyqu:
@simonxander thank you! However, this PR didn't fix the flaky test.

After some discovery around #1495 I found the issue is test JVM OOM-ing during the test suite for Synology.

#1498 is the proposed temporary fix to unblock the CI -- and kickstart the conversation around a more permanent fix.

This PR is no longer needed, closing.

PR #1497, opened 2026-06-22
fix(google-calendar): throw InvalidTokenException on invalid_grant

### LINKED ISSUE #1499: GoogleCalendarImporter does not throw InvalidTokenException on invalid_grant
When importing data into Google Calendar, an expired or revoked OAuth token causes Google's token endpoint to return 400 Bad Request with "error": "invalid_grant". GoogleCalendarImporter does not catch this TokenResponseException, so the failure surfaces as a generic/uncategorized error rather than as an InvalidTokenException. As a result the transfer framework cannot recognise the token as invalid and react appropriately (e.g. trigger re-authentication).
  
This differs from DriveImporter, which already handles this case correctly.

ameya9:
Same as google drive importer so the transfer job recognises the token as invalid instead of treating it as a generic failure

fixes this: https://github.com/dtinit/data-transfer-project/issues/1499


extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/calendar/GoogleCalendarImporter.java
```diff
@@ -113,26 +116,42 @@ public ImportResult importItem(UUID jobId,
...
-    com.google.api.services.calendar.model.Calendar calendarResult =
-        getOrCreateCalendarInterface(authData).calendars().insert(toInsert).execute();
-    return calendarResult.getId();
+    try {
+      com.google.api.services.calendar.model.Calendar calendarResult =
+          getOrCreateCalendarInterface(authData).calendars().insert(toInsert).execute();
+      return calendarResult.getId();
+    } catch (TokenResponseException e) {
+      TokenErrorResponse details = e.getDetails();
+      if (details != null && details.getError().equals("invalid_grant")) {
```
alexeyqu: this might cause NPE if `details.getError()` is null

```suggestion
      if (details != null && "invalid_grant".equals(details.getError())) {
```

extensions/data-transfer/portability-data-transfer-google/src/main/java/org/datatransferproject/datatransfer/google/calendar/GoogleCalendarImporter.java
```diff
@@ -113,26 +116,42 @@ public ImportResult importItem(UUID jobId,
...
-        .getId();
+    try {
+      return getOrCreateCalendarInterface(authData)
+          .events()
+          .insert(newCalendarId, event)
+          .execute()
+          .getId();
+    } catch (TokenResponseException e) {
+      TokenErrorResponse details = e.getDetails();
+      if (details != null && details.getError().equals("invalid_grant")) {
```
alexeyqu: ditto

```suggestion
      if (details != null && "invalid_grant".equals(details.getError())) {
```

alexeyqu (approved):
ok -- please fix the potential NPE before merging

PR #1501, opened 2026-07-06
ci: add Docker-based test workflow alongside Gradle CI

alexeyqu:
Runs the test suite via the Dockerfile/docker-compose added in #1493, on the same push/PR triggers as the existing Gradle workflow, so the two can be compared side by side before the containerized path becomes the default.


lisad (approved):
This looks good and useful to me!

Issue #553, opened 2018-08-17
Generalize AuthServiceExtension


rtannenbaum:
Consolidate auth extensions to generalized ones, one for OAuth1 and one for OAuth2.

The plan is to combine all the service-level auth flows where possible to allow for more uniform and easy-to-understand auth flows based on underlying protocol, and to reduce reliance on third-party SDKs for future secure deployment environments.

out of date:
<del>As discussed in https://portability.slack.com/archives/C8PH5R1C2/p1533592780000248</del>
<del>We should generalize AuthServiceExtension used in the API servers to be basically a config wrapper, which returns auth-related config to the FE</del>

<del>FE's should know how to handle various auth flows given the auth configuration returned from the API</del>

rtannenbaum:
update: this bug is now tracking the effort to consolidate auth extensions to generalized ones, one for OAuth1 and one for OAuth2.

The comment in the PR description about making API servers be auth config managers is now out of date. The latest discussion/thinking is for auth logic to be available in API servers for those who wish to use the API server as a web server to handle auth, but to also provide a separate, more minimal, mode as well for the API server to not handle auth flow (just provide an interface to JobStore). In that scenario, distributions would move auth logic to their frontend. This dual-mode API server approach was [discussed](https://portability.slack.com/files/U8NTF9DNU/FDF7KUCP6/Maintainer_meeting_notes_15_Oct_2018) in our maintainer meeting on 10/15/18.

PR #1516, opened 2026-09-01
test(e2e): add black-box offline-demo transfer harness

alexeyqu:
Adds a `docker compose` harness that runs a **real** `offline-demo` → `offline-demo` transfer — job creation, auth, worker claim, export, import — against the packaged `demo-server` jar, and asserts that the exported payload actually arrives.

```bash
./e2e/run.sh
```

No provider credentials, no local JDK, no local Python. ~35s with a warm Gradle cache.

Nothing in the repo runs a transfer end to end today. The closest is `JobProcessorTest`, which stands up `JobProcessor` against sixteen mocks — useful, but it never moves data between two services.

### Stack

| PR | Scope |
|---|---|
| #1509 (`feat/offline-demo-exporter`) | offline-demo becomes a complete export/import pair — **base of this PR** |
| **this** | the harness that runs it |

Based on #1509 rather than `master` because the transfer is only possible once offline-demo has an exporter. Retarget to `master` after #1509 merges.

### Shape

Three services in the root `docker-compose.yml`; two share the repo's existing `Dockerfile`. **No new Dockerfile, no `build.gradle`, no `settings.gradle` entry, no Java.**

| Service | What it is |
|---|---|
| `gradle` | the Gradle build container (renamed from `test`, see below) |
| `dtp` | the same image, entrypoint overridden to `java -jar` the shadowJar |
| `e2e` | stock `python:3.12-slim` + pytest, no Dockerfile |

The jar is built through the `gradle` service, whose `ENTRYPOINT` is already `./gradlew`, so arguments pass straight through. That keeps the pinned Gradle 6.9.2 / JDK 11 toolchain single-sourced from the root `Dockerfile` and reuses the warm `gradle-cache` volume — which a `docker build` stage could not mount. `shadowJar` has no path to `copyWebApp`, so no Node or Angular toolchain is involved.

`dtp` deliberately defines **no image of its own**. The repo publishes no server image today, so the harness pins a *contract* — an API on 8080 plus a readable log stream — that a future real image can satisfy by replacing three keys on one service. Meanwhile it does run the real packaged artifact, including the `mergeServiceFiles()` `META-INF/services` merge that `ServiceLoader` extension discovery depends on.

### Why the driver is Python, not Java

Fair question on a Java project, so stating it plainly rather than leaving it to review.

The decision that actually mattered was **whether the driver should be a Gradle module** — and a Java driver has to be one. In this repo that is not inert: root `build.gradle` selects subprojects by `new File(it.projectDir, 'src').exists()`, and `configure(sourceProjects())` then applies the `java` plugin, `group`/`version` from `RELEASE_VERSION`, six test dependencies, and **`sourcesJar` + `javadocJar`** to whatever it finds. A test-only module would silently acquire publication-shaped artifacts in a repo that publishes to Maven Central. It would also join `./gradlew build` (or need an opt-in gate), and resolve its dependencies through the pinned Gradle 6.9.2 / JDK 11 toolchain.

An earlier in-process JUnit version of this harness was built and hit exactly that; its `build.gradle` needed a comment explaining that JUnit and Truth were arriving from a block it never asked to join.

Out of the build, a container is the natural unit, and Python fits it: `pytest` + `requests`, pinned, isolated, invisible to the shipped dependency graph. It is also legible to people evaluating DTP who will not read Java — the same audience the credential-free path is for. And once real provider adapters are covered, their mock backends have to be containers anyway, so the driver's language stays invisible to DTP either way.

**The cost, stated honestly:** the Java driver built requests from the real `portability-types-client` DTOs, so it *could not drift* from the API — rename a field and it stops compiling. The Python driver hand-writes those field names, so the same rename surfaces as a runtime 400 instead. CI still catches it, but later and less precisely. That is a real regression and the most likely thing to bite first.

Not a line-count argument: both versions are ~400 lines. If the contract-drift risk is judged too high, the fallback is **JUnit + Testcontainers** — it keeps the typed DTOs and drops the in-process server, at the price of bringing the Gradle module back.

### Included rename: `test` → `gradle`

The build container was called `test` while the service that actually runs tests was called something else. `Documentation/Developer.md` had the tell:

```bash
docker compose run --rm test test --tests SomeTest   # service name and Gradle task, same word
```

`gradle` is exact — the service's `ENTRYPOINT` *is* `./gradlew`.

**The name `test` is left unused rather than reassigned to the pytest service.** It is documented and hardcoded in `docker-build.yml`, so a `test` that survived but meant something else would silently run the e2e suite for anyone with the old command in their fingers — and pass. It now fails with `no such service`. `Developer.md` and `docker-build.yml` are updated to match.

Also corrects a stale claim in `Developer.md` that the root `Dockerfile` pins `gradle:8.10.2-jdk11`; it pins `eclipse-temurin:11-jdk-jammy`.

### Why completion is read from the server log

Because there is no alternative. `TransferJob.state` is a hardcoded `State.CREATED` with **no getter**, so Jackson never emits it and `GET /api/transfer/{id}` reports no progress at all — `GetTransferJobAction` already fetches the real `PortabilityJob` and discards everything but the ids (`TODO(#553)`).

Giving `TransferJob` a real nullable `state` would let the harness poll HTTP instead of grepping. That is a change to shipped code — and a binary-compatibility one, since `portability-types-client` is published — so it is deliberately **not** bundled here.

Worth noting for anyone evaluating the endpoint: `GET /api/transfer/{id}` currently has **no callers at all**. `client-rest` has no `getTransferJob` method, so the route is effectively dead surface today.

### One finding worth reviewer attention

`Finished processing … with 0 error(s).` is **necessary but not sufficient**. That count is `copier.getErrors().size()`, so a job that never copied anything reports zero too. Building the jar with `-PencryptionScheme=jwe` while the driver posts `cleartext` reproduces it exactly:

```
SEVERE  No auth decrypter found for scheme cleartext while processing job: <uuid>
DEBUG   Finished processing jobId: <uuid> with 0 error(s).
```

The success marker fires on a transfer that moved nothing. So failure is detected from `SEVERE` lines naming the job — a clean run emits none — and the delivered payload is asserted **separately** from job completion. This generalises past this PR: any future assertion built on "the job finished cleanly" inherits the flaw.

### Verification

Three negative checks were run by hand; none is committed:

1. **Scheme mismatch** → fails in ~5s on the `SEVERE` line with the server's own message. This is the check that exposed the trap above.
2. **Corrupted expected payload** → the payload assertion fails while completion still passes, proving the two are independent.
3. **Unsupported `dataType`** → both tests fail; the harness does not pass on "nothing crashed".

`docker compose run --rm gradle` (the pre-existing build) confirmed still green.

### What a green run does not prove

- **No real provider adapter is covered.** `offline-demo` has no HTTP surface, no pagination and no sub-resources, so this exercises the machinery rather than a data path. Covering an adapter that actually paginates — Imgur is the cheapest candidate — is the intended next step.
- **Only `cleartext` is exercised.** JWE cannot be a runtime toggle: `SecurityExtensionLoader` asserts exactly one security extension on the classpath, so the scheme is fixed at build time and covering JWE means a second jar.
- **Failures are diagnosable only from a log file**, not from structured received payloads. Pointing a real adapter at a mock HTTP backend is what would make received data directly assertable.

### CI

Runs non-blocking (`continue-on-error: true`) to start with, uploading the server log as an artifact whether the run passes or fails. A flaky e2e that blocks merges is worse than no e2e; drop the flag once it has been green for a while.


lisad (approved):
From my perspective this looks like a very good thing

